PR #100, opened 2019-10-09
[WIP] Add seq2seq_interpret.ipynb to Tutorials

wz337:
Current stage: still looking for a translation pre-trained model that is easy to work with to demonstrate Captum's functionality. 

The notebook contains two parts: Part 1 is a vanilla seq2seq model, which does not perform that well for demonstrative purpose. This part will be discarded once we found a reasonably good pre-trained model.

Part 2 is the tutorial we want to show, including how we are going to interpret results from the models and show visualization. At the end of part 2, there is an example of attribution matrix. Ideally, we want the diagonal to be blue (having higher attribution score).

### DIFF
--- a/tutorials/seq2seq_interpret.ipynb
+++ b/tutorials/seq2seq_interpret.ipynb
@@ -1,0 +1,857 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


edward-io:
Closing due to inactivity.

PR #113, opened 2019-10-11
Fixing Neuron Indexing

vivekmig:
Switching to suggested new approach for indexing neurons with arbitrary length tuples as opposed to an iterated select.


NarineK:
Look great! Thank you!

PR #131, opened 2019-10-22
fix division by zero in captum/insights/features.py

miguelmartin75:
- if attribution vector == 0, L2 and L-infinity norm will be zero
- removed Optional annotation from `TextFeature`'s `visualization_transform` since it is required

### DIFF
--- a/captum/attr/_utils/common.py
+++ b/captum/attr/_utils/common.py
@@ -16,7 +16,7 @@ class ExpansionTypes(Enum):
 def safe_div(denom, quotient, default_value=None):
     r"""
         A simple utility function to perform `denom / quotient`
-        if `quotient` == 0 => result will be `default_value`
+        if the statement is undefined => result will be `default_value`
     """
     return denom / quotient if quotient != 0.0 else default_value
 
--- a/captum/attr/_utils/common.py
+++ b/captum/attr/_utils/common.py
@@ -13,6 +13,14 @@ class ExpansionTypes(Enum):
     repeat_interleave = 2
 
 
+def safe_div(denom, quotient, default_value=None):
+    r"""
+        A simple utility function to perform `denom / quotient`
+        if `quotient` == 0 => result will be `default_value`
+    """
+    return denom / quotient if quotient != 0.0 else default_value
+
+
 def validate_target(num_samples, target):
     if isinstance(target, list) or (
         isinstance(target, torch.Tensor) and torch.numel(target) > 1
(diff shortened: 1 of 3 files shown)


captum/insights/features.py
```diff
@@ -116,7 +116,11 @@ def visualize(self, attribution, data, contribution_frac) -> FeatureOutput:
         attribution = attribution.sum(dim=1)
 
         # L-Infinity norm
-        normalized_attribution = attribution / abs(attribution).max()
+        normalized_attribution = attribution
+        attr_max = abs(attribution).max()
+        if attr_max > 0:
```
miguelmartin75: Probably want to refactor this into a function, since I do it multiple times (I think it is also done [here](https://github.com/pytorch/captum/blob/master/captum/insights/api.py#L154)) - just not sure where I should place it

captum/insights/features.py
```diff
@@ -116,7 +116,11 @@ def visualize(self, attribution, data, contribution_frac) -> FeatureOutput:
...
+            normalized_attribution /= attr_max
```
edward-io: `if attr_max == 0`, then the `normalized_attribution` should be 0, right?

captum/insights/features.py
```diff
@@ -146,7 +150,11 @@ def visualize(self, attribution, data, contribution_frac) -> FeatureOutput:
         data = data.squeeze(0)
 
         # L-2 norm
-        normalized_attribution = attribution / attribution.norm()
+        normalized_attribution = attribution
+        l2_norm = attribution.norm()
+        if l2_norm > 0:
+            normalized_attribution /= l2_norm
```
edward-io: ditto here

captum/insights/api.py
```diff
@@ -150,9 +150,10 @@ def _calculate_net_contrib(self, attrs_per_input_feature: List[Tensor]):
         )
 
         # normalise the contribution, s.t. sum(abs(x_i)) = 1
+        from captum.attr._utils.common import safe_div
```
edward-io: Let's import this at the top

captum/attr/_utils/common.py
```diff
@@ -13,6 +13,14 @@ class ExpansionTypes(Enum):
     repeat_interleave = 2
 
 
+def safe_div(denom, quotient, default_value=None):
```
edward-io: should the default value be 0.0 here?

miguelmartin75: The `default_value` is what is returned if `denom / quotient` is undefined, will document this

PR #134, opened 2019-10-23
Insights API Documentation

vivekmig:
Initial version of Captum Insights API documentation for api.py (AttributionVisualizer and Batch)

### DIFF
--- a/captum/insights/api.py
+++ b/captum/insights/api.py
@@ -46,6 +46,8 @@ class Batch:
                         input tensor corresponds to the number of examples
                         (batch size) and is aligned for all input tensors.
             labels (tensor): Tensor containing correct labels for input examples.
+                        This must be a 1D tensor with length matching the first
+                        dimension of each input tensor.
             additional_args (tuple, optional): If the forward function
                         requires additional arguments other than the inputs for
                         which attributions should not be computed, this argument
--- a/captum/insights/api.py
+++ b/captum/insights/api.py
@@ -47,6 +47,37 @@ class AttributionVisualizer(object):
         score_func: Optional[Callable] = None,
         use_label_for_attr: bool = True,
     ):
+        r"""
+        Args:
+
+            models (torch.nn.module): PyTorch module for attribution visualization.
+                          We plan to support visulizing and comparing multiple models
+                          in the future, but currently this supports only a single
+                          model.
+            classes (list of strings): List of strings corresponding to the names of
+                          classes for classification.
+            features (list of BaseFeatures): List of BaseFeatures, which correspond
+                          to input arguments to the model. Each feature object defines
+                          relevant transformations for converting to model input,
+                          visualizing and constructing baselines. The length of the
+                          features list should exactly match the number of (tensor)
+                          arguments expected by the given model.
+                          For instance, an image classifier should only provide
+                          a single BaseFeature, while a multimodal classifier may
+                          provide a list of features, each corresponding to a different
+                          tensor input and potentially different modalities.
+            dataset (iterable of Batch): Defines dataset to visualize attributions for.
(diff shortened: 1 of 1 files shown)


captum/insights/api.py
```diff
@@ -32,6 +32,34 @@ def __init__(
...
+            inputs (tensor or tuple of tensors): Batch of inputs for a model.
+                        These may be either a Tensor or tuple of tensors. Each tensor
+                        must correspond to a feature for AttributionVisualizer, and
+                        the corresponding input transform function of the feature
+                        is applied to each input tensor prior to passing it to the
+                        model. It is assumed that the first dimension of each
+                        input tensor corresponds to the number of examples
+                        (batch size) and is aligned for all input tensors.
+            labels (tensor): Tensor containing correct labels for input examples.
+                        This must be a 1D tensor with length matching the first
```
vivekmig: Similar here, this could also just be a list of ints? We'd need to update the type hints as well here.

vivekmig:
I also need to add a docstring for the render method here, but will do that once this PR https://github.com/pytorch/captum/pull/135 merges, since it adds a port argument.

captum/insights/api.py
```diff
@@ -32,6 +32,34 @@ def __init__(
         labels: Optional[Tensor],
         additional_args=None,
     ):
+        r"""
+        Constructs batch of inputs to be attributed and visualized.
+
+        Args:
+
+            inputs (tensor or tuple of tensors): Batch of inputs for a model.
```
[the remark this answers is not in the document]
edward-io: We pass the inputs to `_batched_generator`'s inputs, which expect a tensor from `_format_tensor_into_tuples`.

captum/insights/api.py
edward-io: here, label can be a list.

captum/insights/api.py
```diff
@@ -47,6 +75,46 @@ def __init__(
...
+        Args:
+
+            models (torch.nn.module): PyTorch module (model) for attribution
+                          visualization.
+                          We plan to support visulizing and comparing multiple models
+                          in the future, but currently this supports only a single
+                          model.
+            classes (list of strings): List of strings corresponding to the names of
+                          classes for classification.
+            features (list of BaseFeatures): List of BaseFeatures, which correspond
```
edward-io: should this be list of `BaseFeature` (not plural?) Not sure how our documentation rendering system works. Same with list of strings above.

captum/insights/api.py
```diff
@@ -47,6 +75,46 @@ def __init__(
         score_func: Optional[Callable] = None,
         use_label_for_attr: bool = True,
     ):
+        r"""
+        Args:
+
+            models (torch.nn.module): PyTorch module (model) for attribution
+                          visualization.
+                          We plan to support visulizing and comparing multiple models
```
miguelmartin75: visulizing -> visualizing

captum/insights/api.py
```diff
@@ -47,6 +75,46 @@ def __init__(
...
+                          to input arguments to the model. Each feature object defines
+                          relevant transformations for converting to model input,
+                          constructing baselines, and visualizing. The length of the
+                          features list should exactly match the number of (tensor)
+                          arguments expected by the given model.
+                          For instance, an image classifier should only provide
+                          a single BaseFeature, while a multimodal classifier may
+                          provide a list of features, each corresponding to a different
+                          tensor input and potentially different modalities.
+            dataset (iterable of Batch): Defines dataset to visualize attributions for.
```
miguelmartin75: Defines the

captum/insights/api.py
vivekmig: Sounds good, let's make this update separately, updating the type hint and docstring.

PR #137, opened 2019-10-23
Fixing python3 shebang

vivekmig:
Adding python3 shebang line to all files for consistency.

### DIFF
--- a/tests/attr/test_common.py
+++ b/tests/attr/test_common.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 from __future__ import print_function
 
 import torch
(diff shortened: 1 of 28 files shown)


tests/attr/test_common.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 from __future__ import print_function
```
edward-io: should we remove the `__future__` statements then? The future is now :)  ditto to all the files in the tests directory.

vivekmig: Sure, good point, will remove them!

Issue #35, opened 2019-09-21
Returning only the gradients/"multipliers"


AvantiShri:
Hi all,

Just wanted to put this particular use-case on your radar. Sometimes we find that it is useful to get access to just the gradients ("multipliers"), before they are multiplied by the difference-from-reference to get the final attribution. Specifically, we use the multipliers to estimate how the network _might have_ responded had it seen slightly different inputs. We refer to these estimates as "hypothetical contribution scores". If you are curious how these hypothetical contributions look, here's a notebook (on a fork of the DeepSHAP repository) where I compute hypothetical contributions in the context of genomic data: https://github.com/AvantiShri/shap/blob/0b0350ba3a42af275f6e99ca2e3c5877d7d94f8a/notebooks/deep_explainer/PyTorch%20Deep%20Explainer%20DeepSEA%20example.ipynb

You've all done an awesome job with this repository, and I will definitely point it to the pytorch users in my lab once the release is formally announced. I totally understand if the ability to return just the multipliers is not something that you are likely to incorporate in the main release; I'm sure we can easily fork the repository and add that feature in for our lab's purposes.

Thanks again!
Av

orionr:
Not going to close this, since we can consider it as a feature request that we aren't working on yet. Thanks, @AvantiShri!

AvantiShri:
Thanks @orionr! One thing to note about how I implemented support for this feature in my fork of the shap repository (which is linked to in the notebook above) is that I allow the user to optionally provide a custom function that takes as input the `(multipliers, original input, references)` and returns the final scores - I think the equivalent in Captum would be allowing a user to provide a custom function that varies the logic in this line:
https://github.com/pytorch/captum/blob/2fea0d764a19e621e2e8da86aa6424f9cd74def2/captum/attr/_core/deep_lift.py#L207

[...]

PR #141, opened 2019-10-26
Compute deeplift contributions w.r.t. arbitrary layers

### LINKED ISSUE #91: Computing contributions w.r.t. logits rather than final activations
Often, in practice, we wish to compute the contributions w.r.t. the logits of the final sigmoid/softmax, rather than w.r.t. the final network output itself. This is to avoid artifacts that can be caused by the saturating nature of the sigmoid/softmax, and comes into play when comparing attributions between examples. It is particularly relevant if gradient*input is used as an attribution method, because for examples with very confident predictions, the sigmoid/softmax outputs tend to saturate and the gradients will approach zero. I'm wondering if it may be worth mentioning this in the documentation - in the current "getting started", the toy model has a sigmoid output:

Screenshot 2019-10-08 at 2 19 52 PM

I'm concerned that a naive user may try to compare the magnitudes of attributions across different examples without realizing that, for sigmoid/softmax outputs, it may be worth removing the final nonlinearity before doing such a comparison. We discuss this in [Section 3.6](https://arxiv.org/pdf/1704.02685.pdf) of the deeplift paper. Ideally there would be an option in Captum to ignore the final nonlinearity, but I realize it may not be trivial to add that option. Sorry if this is already addressed and I missed it.

gabrieltseng:
Closes #91 

Allows an additional parameter `target_output_layer` to be passed to the DeepLift constructor. If one is passed, the contributions are calculated with respect to the output of this layer instead of the final outputs of the model itself.

Thanks!

### DIFF
--- a/tests/attr/test_deeplift_basic.py
+++ b/tests/attr/test_deeplift_basic.py
@@ -3,6 +3,8 @@
 import torch
 from torch import nn
 
+import pytest
+
 from inspect import signature
 
 from captum.attr._core.deep_lift import DeepLift, DeepLiftShap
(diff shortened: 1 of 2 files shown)


NarineK:
Hi @gabrieltseng, thank you for the PR! We actually, implement the layers separately. Similar to IG, we prefer not to pass a separate parameter to keep code clean and have a separate file. I have already an implementation for it. I haven't pushed the PR yet.Sorry, I wasn't probably clear about it in the issue.

In the issue, I mentioned that I have a branch for layer already, I'm testing it for Dataparallel and  was about to push the PR soon.

NarineK:
sorry, I think I confused  #91 with  #35 and I somehow had in mind that you wanted to implement #35 .

tests/attr/test_deeplift_basic.py
```diff
@@ -71,6 +74,80 @@ def test_relu_linear_deeplift(self):
...
+                return self.additional(output)
+
+        model = NewModel()
+        main_model = ReLULinearDeepLiftModel()
+
+        dl_truncated = DeepLiftShap(model, target_output_layer=model.main_model)
+        dl_main = DeepLiftShap(main_model)
+
+        attrs_truncated = dl_truncated.attribute(inputs, baselines)
+        attrs_main = dl_main.attribute(inputs, baselines)
```
NarineK: @gabrieltseng , I think your implementation is similar to `NeuronDeeplift` but we fix neuron index, In order to support your implementation, we'll probably need to allow `neuron_index=None` and compute the gradients of the layer (all neuron in that layer) with respect to the inputs.

 https://github.com/pytorch/captum/pull/142/files#diff-ac1add4f968d4c9d5fc3edd1141da856R25 

Your tests passed for me as I tried:

```
        dl_truncated = NeuronDeepLift(model, model.main_model)
        dl_main = DeepLift(main_model)

        attrs_truncated = dl_truncated.attribute(inputs, 0, baselines)
       ...
```

gabrieltseng: Ah awesome! I think this means this PR can be closed.

Thanks!

PR #145, opened 2019-10-30
Add back auto_deploy_site for CircleCI master

orionr:
This roughly reverts https://github.com/pytorch/captum/commit/d327a247f3b7e52e160e725910fdd45872681505

and makes it so the captum.ai website is automatically updated when changes land in master.

cc @jlin27 @caraya10 @NarineK @vivekmig


orionr:
Merging this - hopefully it doesn't break gh-pages! If it does, I'll rollback.

PR #146, opened 2019-10-30
Redirect website to real root domain

zpao:
For "reasons", we don't actually have a CNAME set up in GitHub. The
result is that GitHub doesn't know that pytorch.org/captum should
actually redirect to captum.ai. This change puts a bit of JS on every
page in prod that will do that redirect. We should fix it properly but
at least we'll have something in place.

### DIFF
--- a/website/core/Footer.js
+++ b/website/core/Footer.js
@@ -92,6 +92,20 @@ class Footer extends React.Component {
           )}{" "}
           Copyright &copy; {currentYear} Facebook Inc.
         </section>
+        {process.env.NODE_ENV !== 'development' &&
+          <script dangerouslySetInnerHTML={{__html:`
+            (function() {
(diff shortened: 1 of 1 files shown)


zpao:
FYI: don't forget to rebuild & deploy the website to pick this up (generally should treat that gh-pages branch as a build artifact, I won't send any PRs against it)

orionr:
Looking to have automatic deploys to captum.ai from master, but running into some snags at https://circleci.com/gh/pytorch/captum/1867  cc'ed @jlin27 and @JoelMarcey on this, since I'm not  sure what we need to setup in CircleCI to have `https://docusaurus-bot@github.com` password set

JoelMarcey:
I think we are on our way to solving this - might have been a misnamed env variable in the config.

PR #158, opened 2019-11-05
Add underscores in front of the helper functions in `common.py` and support scalar-valued baselines

NarineK:
I should have split this into two PRs. Originally, I thought that this will be a small change but turned out to be bigger. To avoid merge conflicts I'm pushing this into one PR. 

This PR:
1. Underscores helper functions from `common.py` for formatting and validation
2. Enables baselines to be scalars or tuples of scalars
   2.1. Added documentation change which is available only in:
https://github.com/pytorch/captum/compare/master...NarineK:fix_underscored_functions_and_support_scalar_baselines?expand=1#diff-73cf88fd88dbc0794143fda23d832009R106
         Once this gets reviewed I'll apply it to everywhere
   2.2 Adds test cases with scalar baselines or tuples of scalar for some algorithms and for some algorithms which already had many test cases, I changed baselines to also take scalar values.


tests/attr/test_deeplift_basic.py
```diff
@@ -141,7 +154,9 @@ def _deeplift_assert(self, model, attr_method, inputs, baselines):
             else:
                 baselines = baselines()
 
-        baseline_bsz = len(baselines[0])
+        baseline_bsz = (
+            len(baselines[0]) if isinstance(baselines[0], torch.Tensor) else 1.0
```
vivekmig: nit: Can this be just 1 in the else since it corresponds to a size?

captum/attr/_utils/common.py
```diff
@@ -166,7 +171,22 @@ def _zeros(inputs):
     shape as the `inputs`
 
     """
-    return tuple(0 * input for input in inputs)
+    return tuple(0.0 for input in inputs)
+
+
+def _tensorize_baseline(inputs, baselines):
+    assert isinstance(inputs, tuple) and isinstance(baselines, tuple), (
```
vivekmig: Can we potentially combine this with _expand_baselines_based_on_inputs to perform all baseline expansions in one call?

tests/attr/test_deeplift_basic.py
```diff
@@ -62,12 +62,19 @@ def test_relu_linear_deeplift(self):
         x1 = torch.tensor([[-10.0, 1.0, -5.0]], requires_grad=True)
         x2 = torch.tensor([[3.0, 3.0, 1.0]], requires_grad=True)
 
-        b1 = torch.tensor([[0.0, 0.0, 0.0]], requires_grad=True)
-        b2 = torch.tensor([[0.0, 0.0, 0.0]], requires_grad=True)
-
         inputs = (x1, x2)
-        baselines = (b1, b2)
+        baselines = (0.0, 0.0001)
```
vivekmig: nit: Could make the first one int 0 to test both int and float?

tests/attr/test_layer_conductance.py
```diff
@@ -20,6 +20,15 @@ def test_simple_input_conductance(self):
         inp = torch.tensor([[0.0, 100.0, 0.0]])
         self._conductance_test_assert(net, net.linear0, inp, [[0.0, 390.0, 0.0]])
 
+    def test_simple_input_with_scalar_baseline_conductance(self):
+        net = BasicModel_MultiLayer()
+        inp = torch.tensor([[0.0, 100.0, 0.0]])
+        cond = LayerConductance(net, net.linear0)
+        attributions = cond.attribute(inp, 0.0001, target=0)
+        assertArraysAlmostEqual(
+            attributions.squeeze(0).tolist(), [-0.00039, 390.0, -0.00039]
```
vivekmig: nit: Could maybe make the scalar baseline a little larger, since this expected conductance is within the delta of 0.05?

captum/attr/_core/deep_lift.py
```diff
@@ -102,16 +104,32 @@ def attribute(
...
+                        computes the differences between the inputs and references
+                        and corresponding outputs.
+                        Baselines can be provided either as :
+
+                        - a single tensor, if inputs is a single tensor, with
+                            exactly the same dimensions as inputs or the first
+                            dimension is one and the remaining dimensions match
+                            with inputs.
+
+                        - a tuple of tensors, if inputs is a tuple of tensors,
```
vivekmig: nit: This probably should be updated as well to say that the first dimension of any of them in the tuple could be 1? Also, from the implementation, it looks like you can also mix tensors (with first dimension 1 or full dimension) and scalars, can also document that if it's true?

captum/attr/_core/noise_tunnel.py
```diff
@@ -184,28 +184,36 @@ def add_noise_to_input(input, stdev):
...
+                    if isinstance(baseline, torch.Tensor)
+                    else baseline
+                    for baseline in baselines
+                )
             else:
                 baselines = tuple(
                     baseline.repeat_interleave(n_samples, dim=0)
-                    for baseline in baselines
+                    if isinstance(baseline, torch.Tensor)
+                    and baseline.shape[0] == input.shape[0]
```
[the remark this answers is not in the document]
NarineK: actually, I don't need if only `baseline.shape[0] == 1`

Issue #156, opened 2019-11-03
Undesirable behavior of LayerActivation in networks with inplace ReLUs


mrsalehi:
Hi,
I was trying to use `captum.attr._core.layer_activation.LayerActivation` to get the activation of the first convolutional layer in a simple model. Here is my code:
```Python
torch.manual_seed(23)
np.random.seed(23)
model = nn.Sequential(nn.Conv2d(3, 4, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1)),
                      nn.ReLU(inplace=True),
                      nn.Conv2d(4, 4, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1)),
                      nn.ReLU(inplace=True))

layer_act = LayerActivation(model, model[0])
input = torch.randn(1, 3, 5, 5)
mylayer = model[0]
print(torch.norm(mylayer(input) - layer_act.attribute(input), p=2))
```
In fact, I have computed the activation in two different ways and compared them afterwards. Obviously, I expected a value close to zero to be printed as the output, however, this is what I got:
```Python
tensor(3.4646, grad_fn=<NormBackward0>)
```
I hypothesize that the inplace `ReLU` layer after the convolutional layer acts on its output since there were many zeros in the activation computed by Captum ( i.e. `layer_act.attribute(input)`). In fact, when I changed the architecture of the network to the following: 

```Python
model = nn.Sequential(nn.Conv2d(3, 4, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1)),
                      nn.ReLU(),
                      nn.Conv2d(4, 4, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1)),
                      nn.ReLU(inplace=True))
```
then the outputs matched.

**System information**
*  Python 3.7.0
*  torch 1.3.0
* Captum 0.1.0

vivekmig:
Hi @mrsalehi, yes, this is a bug, thanks for pointing it out! We will push a fix for this soon.

vivekmig:
Fix has been merged here: https://github.com/pytorch/captum/commit/5bf06ba94c3ea1c992b5cc2b29daa06fae527d34

PR #165, opened 2019-11-07
Fixing Layer / Neuron Attributions with In-place Operations

vivekmig:
This PR fixes neuron / layer attributions with in-place operations by keeping appropriate clones of intermediate values to ensure that they are not modified by future operations.

Addresses Issue: #156

### DIFF
--- a/captum/attr/_utils/gradient.py
+++ b/captum/attr/_utils/gradient.py
@@ -160,7 +160,7 @@ def _forward_layer_eval_with_neuron_grads(
     # get layer output tensor(s).
     # For DataParallel models, each partition adds entry to dictionary
     # with key as device and value as corresponding Tensor.
-    def forward_hook(module, inp, out):
+    def forward_hook(module, inp, out=None):
         eval_tsr = inp if attribute_to_layer_input else out
 
         # if `inp` or `out` is a tuple of one tensor, assign that tensor to `eval_tsr`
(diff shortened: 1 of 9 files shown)


NarineK:
Nice! Can we maybe create a similar test model from https://github.com/pytorch/captum/issues/156
with `nn.Sequential` and access the first layer from the module like in the issue ?

vivekmig:
Sure, I've added the test!

vivekmig:
Yup, the backward hook issue only affects the last layer / module before the hook (only for layer output). An example and test for this case is test_deconv_matching, which checks that NeuronDeconvolution matches Deconvolution output, which is not true if the hook doesn't execute on the last layer (which is a ReLU) in the neuron gradient computation.

The two functions do compute different things, but both can compute neuron gradients when given gradient_neuron_index. Since compute_layer_gradients_and_eval computes the layer gradients as well, it needs to hook and clone a little differently (overriding the module output with the clone) which inhibits the last layer's module hook. So if this is necessary, _forward_layer_eval_with_neuron_grads should be used to obtain neuron gradients. Sorry, this probably wasn't very clear in the comment, can update to make it clearer!

captum/attr/_utils/gradient.py
```diff
@@ -175,9 +175,12 @@ def forward_hook(module, inp, out):
...
             # doesn't support a tuple of inputs
-            saved_layer[eval_tsr.device] = eval_tsr
+            saved_layer[eval_tsr.device] = eval_tsr.clone()
 
-    hook = layer.register_forward_hook(forward_hook)
+    if attribute_to_layer_input:
+        hook = layer.register_forward_pre_hook(forward_hook)
+    else:
+        hook = layer.register_forward_hook(forward_hook)
     _run_forward(forward_fn, inputs, additional_forward_args=additional_forward_args)
```
NarineK: I was just looking into this. Doesn't `_run_forward ` also need `target_ind` to be passed as an argument ?

PR #166, opened 2019-11-08
Update the code in docs

PingjunChen:
The demo codes in the docs is not very accurate. These two are very obvious errors that should be better updated.


PingjunChen:
In addtion, the parameter `return_convergence_delta` need to be set as `True` if `ig.attribute` would like to obtain the value of  `delta`.

Issue #163, opened 2019-11-07
Toy Example breaks with CUDA on compute_convergence_delta for Integrated Gradients


suragnair:
For the toy example with cuda

```python
model = ToyModel()
model = model.cuda()
model.eval()

input = torch.rand(2, 3).cuda()
baseline = torch.zeros(2, 3).cuda()

ig = IntegratedGradients(model)
attributions, delta = ig.attribute(input, baseline, target=0, return_convergence_delta=True)
```

fails with the error 
```python
~/anaconda3/envs/heterokaryon/lib/python3.7/site-packages/captum/attr/_utils/attribution.py in compute_convergence_delta(self, attributions, start_point, end_point, target, additional_forward_args)
    232         row_sums = [_sum_rows(attribution) for attribution in attributions]
    233         attr_sum = torch.tensor([sum(row_sum) for row_sum in zip(*row_sums)])
--> 234         return attr_sum - (end_point - start_point)
    235 
    236 

RuntimeError: expected device cpu and dtype Float but got device cuda:0 and dtype Float
```

presumably since `attr_sum` is not on GPU. Turning `return_convergence_delta` to `False` results in no error.

Similar issues may arise in other places, though I haven't checked.

vivekmig:
Hi @suragnair, thanks for pointing out this bug! We will push a fix for this soon.

PR #169, opened 2019-11-11
Fixing CUDA delta issue

vivekmig:
Convergence delta was failing on GPUs since the total attribution tensor was not being created on the same device, this fixes that bug. Also, adds data parallel / CUDA tests for computing deltas whenever available.

Issue: #163

### DIFF
--- a/tests/attr/test_data_parallel.py
+++ b/tests/attr/test_data_parallel.py
@@ -420,6 +420,8 @@ class Test(BaseGPUTest):
             attr_dp = algorithm(dp_model)
 
         batch_sizes = [None]
+        delta_orig = None
+        delta_dp = None
         if test_batches:
             batch_sizes = [None, 1, 8]
         for batch_size in batch_sizes:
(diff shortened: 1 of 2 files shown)


tests/attr/test_data_parallel.py
```diff
@@ -428,13 +428,23 @@ def _data_parallel_test_assert(
                     internal_batch_size=batch_size, **kwargs
                 )
             else:
-                attributions_orig = attr_orig.attribute(**kwargs)
+                if attr_orig.has_convergence_delta():
+                    attributions_orig = attr_orig.attribute(
+                        return_convergence_delta=True, **kwargs
```
NarineK: Looks great! Do we want to return `attributions_orig, delta = ` here ?

tests/attr/test_data_parallel.py
```diff
@@ -428,13 +428,23 @@ def _data_parallel_test_assert(
...
+                else:
+                    attributions_orig = attr_orig.attribute(**kwargs)
             if batch_size:
                 attributions_dp = attr_dp.attribute(
                     internal_batch_size=batch_size, **kwargs
                 )
             else:
-                attributions_dp = attr_dp.attribute(**kwargs)
+                if attr_orig.has_convergence_delta():
+                    attributions_dp = attr_dp.attribute(
```
NarineK: nit: also here: `attributions_dp, delta` ?

vivekmig:
Thanks, good point, updated!

PR #170, opened 2019-11-12
Add support for custom attribution function in DeepLift/DeepLiftShap

NarineK:
This PR adds:

1. support for custom attribution function in DeepLift based on the logic described in feature request: https://github.com/pytorch/captum/issues/35.
In case custom function is not provided the default logic is used.

2. test cases. I also added a test case using Avanti's logic on hypothetical input:
https://github.com/AvantiShri/shap/blob/0b0350ba3a42af275f6e99ca2e3c5877d7d94f8a/notebooks/deep_explainer/PyTorch%20Deep%20Explainer%20DeepSEA%20example.ipynb

3. documentation in DeepLift and DeepLiftSHAP. Once this is reviewed I'll add it in Layer and Neuron implementations as well.

### DIFF
--- a/captum/attr/_core/deep_lift.py
+++ b/captum/attr/_core/deep_lift.py
@@ -183,11 +183,12 @@ class DeepLift(GradientAttribution):
                             - custom_attribution_func(multipliers)
                             - custom_attribution_func(multipliers, inputs)
                             - custom_attribution_func(multipliers, inputs, baselines)
-                        In case this function is not provided we use the default
+                        In case this function is not provided, we use the default
                         logic defined as: multipliers * (inputs - baselines)
-                        It is assumed that `custom_attribution_func` returns a
-                        tuple of attribution tensors that have the same shape and
-                        dimensionality as the `inputs`.
+                        It is assumed that all input arguments, `multipliers`,
+                        `inputs` and `outputs` are provided in tuples of same length.
+                        `custom_attribution_func` returns a tuple of attribution
+                        tensors that have the same length as the `inputs`.
 
                         Default: None
 
@@ -205,10 +206,14 @@ class DeepLift(GradientAttribution):
                 This is computed using the property that
                 the total sum of forward_func(inputs) - forward_func(baselines)
                 must equal the total sum of the attributions computed
-                based on Deeplift's rescale rule.
+                based on DeepLift's rescale rule.
                 Delta is calculated per example, meaning that the number of
                 elements in returned delta tensor is equal to the number of
                 of examples in input.
+                Note that the logic described for deltas is guaranteed when the
+                default logic for attribution computations is used, meaning that the
+                `custom_attribution_func=None`, otherwise it is not guaranteed and
+                depends on the specifics of the `custom_attribution_func`.
 
         Examples::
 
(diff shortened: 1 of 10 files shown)


tests/attr/test_deeplift_basic.py
```diff
@@ -199,3 +261,35 @@ def _deeplift_assert(self, model, attr_method, inputs, baselines):
                 ig = IntegratedGradients(model)
                 attributions_ig = ig.attribute(inputs, baselines)
                 assertAttributionComparision(self, attributions, attributions_ig)
+
+
+def _hypothetical_contrib_func(multipliers, inputs, baselines):
+    r"""
+    Implements hypothetical input contributions based on the logic described here:
+    https://github.com/kundajelab/deeplift/pull/36/files
+    This is not using a dummy model for test purposes
```
NarineK: `This is using`

tests/attr/test_deeplift_basic.py
```diff
@@ -199,3 +261,35 @@ def _deeplift_assert(self, model, attr_method, inputs, baselines):
                 ig = IntegratedGradients(model)
                 attributions_ig = ig.attribute(inputs, baselines)
                 assertAttributionComparision(self, attributions, attributions_ig)
+
+
+def _hypothetical_contrib_func(multipliers, inputs, baselines):
```
NarineK: The expanded/repeated inputs are passed here for the SHAP version but we can potentially pass original inputs and baselines to the function.
cc: @AvantiShri

AvantiShri: Passing the expanded inputs sounds fine to me and is also much cleaner from an API perspective! Thank you!

captum/attr/_core/deep_lift.py
```diff
@@ -174,6 +176,20 @@ def attribute(
...
+                        following signature:
+                            - custom_attribution_func(multipliers)
+                            - custom_attribution_func(multipliers, inputs)
+                            - custom_attribution_func(multipliers, inputs, baselines)
+                        In case this function is not provided we use the default
+                        logic defined as: multipliers * (inputs - baselines)
+                        It is assumed that `custom_attribution_func` returns a
+                        tuple of attribution tensors that have the same shape and
+                        dimensionality as the `inputs`.
+
```
vivekmig: It might also be worth documenting that the deltas may not be meaningful depending on this function, e.g. with a hypothetical input, since the delta should then be based on the hypothetical input evaluation (and even then, the completeness property would also not necessarily hold)? We could potentially allow this function to return overriden inputs and baselines to compute a delta based on that, but that may be unnecessarily complicated, and a user could do it themselves later if they really want it with compute_convergence_delta. Alternatively, we could just disallow delta computation in conjunction with custom_attribution_func and just return the output of the function, which lets the user do whatever they want with the multiplier, inputs, and baselines. What do you think?

NarineK: Good point! I think we can leave the logic of computing deltas and add documentation about convergence deltas. 
Too many checks will make the code harder to maintain.

captum/attr/_core/deep_lift.py
```diff
@@ -174,6 +176,20 @@ def attribute(
...
+                        at least one and at most three arguments with the
+                        following signature:
+                            - custom_attribution_func(multipliers)
+                            - custom_attribution_func(multipliers, inputs)
+                            - custom_attribution_func(multipliers, inputs, baselines)
+                        In case this function is not provided we use the default
+                        logic defined as: multipliers * (inputs - baselines)
+                        It is assumed that `custom_attribution_func` returns a
+                        tuple of attribution tensors that have the same shape and
+                        dimensionality as the `inputs`.
```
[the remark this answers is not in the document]
NarineK: if they don't compute convergence delta, they can even return a scalar.

captum/attr/_utils/common.py
```diff
@@ -313,6 +313,22 @@ def _expand_target(target, n_steps, expansion_type=ExpansionTypes.repeat):
     return target
 
 
+def _call_custom_attribution_func(
+    custom_attribution_func, multipliers, inputs, baselines
+):
+    custom_attr_func_params = signature(custom_attribution_func).parameters
```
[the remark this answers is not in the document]
NarineK: Adding, hopefully after type hints we can avoid all this checks

PR #176, opened 2019-11-18
Adding Layer Ablation, Neuron Ablation, and Occlusion

vivekmig:
Adding variants of basic Feature Ablation including Layer Ablation, Neuron Ablation and Occlusion along with necessary basic and data parallel tests.


captum/attr/_utils/attribution.py
```diff
@@ -20,6 +20,16 @@ class Attribution:
     to extend and override core `attribute` method.
     """
 
+    def __init__(self, forward_func):
```
vivekmig: @NarineK The changes here are to create a class hierarchy that works for everything, once we decide on a final structure for layer, neuron, gradient and perturbation attribution, can make those changes in a separate PR and clean this up.

vivekmig:
Closing and replacing with separate PRs for each method

PR #177, opened 2019-11-19
Update the demo codes output to be fully consistent with practical

PingjunChen:
Update the demo attributions and convergence delta output to be fully consistent with the practical testing on the terminal.


README.md
```diff
@@ -171,12 +171,13 @@ scores to each input feature with respect to the first target output.
...
 ig = IntegratedGradients(model)
 attributions, delta = ig.attribute(input, baseline, target=0, return_convergence_delta=True)
-print('IG Attributions: ', attributions, ' Convergence Delta: ', delta)
+print('IG Attributions: ', attributions)
+print('Convergence Delta: ', delta)
 ```
 Output:
 ```
-IG Attributions:  tensor([[-0.5922, -1.5497, -1.0067],
-                          [ 0.0000, -0.2219, -5.1991]])
```
NarineK: In similar cases I'd keep the original indention.

PingjunChen: I do agree that keep the originally aligned indention is more clear for the user to read. 

Updated the output format for the `attributions `.

PingjunChen:
Keep the aligned sub-arrays do look nicer. I updated the output of `attributions` to the original status.

PR #180, opened 2019-11-20
Feature/aggregator

miguelmartin75:
Need to add type hints


tests/attr/test_stat.py
```diff
@@ -0,0 +1,155 @@
...
+            actual_var = torch.from_numpy(np.array(actual_var))
+
+            var = graph.summary["variance"]
+
+            assertTensorAlmostEqual(self, var, actual_var)
+            self.assertTrue((var > 0).all())
+
+    def test_stats_random_data(self):
+        N = 1000
+        BIG_VAL = 100000
```
miguelmartin75: As `N` or `BIG_VAL` grows, rounding errors occur and causes a difference in numpy; note numpy has it implemented as `mean((x-m)**2)` (i.e. without a rolling variance).

Hence why these values are chosen to be low (in order for the tests to pass), but this is a bit hacky, not sure what else I can do though.

NarineK:
Thank you, @miguelmartin75 ! Would you, please, add a brief description of the PR, besides the `Need to add type hints` ?

orionr:
Can we close this PR, since you have the other simplified version? Thanks!

PR #181, opened 2019-11-20
Adding ReLU flag for GradCAM implementation

vivekmig:
This adds a flag to perform the final ReLU operation, which allows results to match the original algorithm exactly.

### DIFF
--- a/captum/attr/_core/layer/grad_cam.py
+++ b/captum/attr/_core/layer/grad_cam.py
@@ -130,7 +130,7 @@ class LayerGradCam(LayerAttribution):
                             apply a ReLU operation on the final attribution,
                             returning only non-negative attributions. Setting this
                             flag to True matches the original GradCAM algorithm,
-                            otherwise by default, both positive and negative
+                            otherwise, by default, both positive and negative
                             attributions are returned.
                             Default: False
 
@@ -187,5 +187,4 @@ class LayerGradCam(LayerAttribution):
         scaled_act = torch.sum(summed_grads * layer_eval, dim=1, keepdim=True)
         if relu_attributions:
             return F.relu(scaled_act)
-        else:
-            return scaled_act
+        return scaled_act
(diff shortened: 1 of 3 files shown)


captum/attr/_core/layer/grad_cam.py
```diff
@@ -123,6 +126,13 @@ def attribute(
                             attribute to the input or output, are single tensors.
                             Support for multiple tensors will be added later.
                             Default: False
+                relu_attributions (bool, optional): Indicates whether to
+                            apply a ReLU operation on the final attribution,
+                            returning only non-negative attributions. Setting this
+                            flag to True matches the original GradCAM algorithm,
+                            otherwise by default, both positive and negative
```
NarineK: nit: `otherwise, by default, both positive and` ... ?

captum/attr/_core/layer/grad_cam.py
```diff
@@ -174,6 +184,8 @@ def attribute(
...
         )
 
-        non_neg_scaled_act = torch.sum(summed_grads * layer_eval, dim=1, keepdim=True)
-
-        return non_neg_scaled_act
+        scaled_act = torch.sum(summed_grads * layer_eval, dim=1, keepdim=True)
+        if relu_attributions:
+            return F.relu(scaled_act)
+        else:
+            return scaled_act
```
NarineK: nit: maybe to save one line ;) 

```
if relu_attributions:
     return F.relu(scaled_act)
return scaled_act

```

tests/attr/layer/test_grad_cam.py
```diff
@@ -25,6 +25,33 @@ def test_simple_input_conv_relu(self):
         inp = 1.0 * torch.arange(16).view(1, 1, 4, 4).type(torch.FloatTensor)
         self._grad_cam_test_assert(net, net.relu1, inp, [[0.0, 4.0], [28.0, 32.5]])
 
+    def test_simple_input_conv_without_final_relu(self):
+        net = BasicModel_ConvNet_One_Conv()
+        inp = 1.0 * torch.arange(16).view(1, 1, 4, 4).type(torch.FloatTensor)
```
NarineK: nit: we could also do `inp.float()` instead of `type(torch.FloatTensor)`?
This applies also to other tests. Actually, do we need to multiply with `1.0` ?

PR #183, opened 2019-11-21
Fixing Class Hierarchy for Multiple Inheritance

vivekmig:
Summary: Remove all instances of super in class hierarchy and switch to direct initializations to support multiple / diamond inheritance appropriately.

Differential Revision: D18642075

### DIFF
--- a/captum/attr/_core/layer/layer_deep_lift.py
+++ b/captum/attr/_core/layer/layer_deep_lift.py
@@ -43,8 +43,8 @@ class LayerDeepLift(LayerAttribution, DeepLift):
             )
             model = model.module
 
-        super(LayerAttribution, self).__init__(model, layer)
-        super(DeepLift, self).__init__(model)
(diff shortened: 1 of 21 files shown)


captum/attr/_core/layer/layer_deep_lift.py
```diff
@@ -308,8 +308,8 @@ def __init__(self, model, layer):
                           Currently, it is assumed that both inputs and ouputs of
                           the layer can only be a single tensor.
         """
-        super(DeepLiftShap, self).__init__(model)
-        super(LayerDeepLift, self).__init__(model, layer)
+        DeepLiftShap.__init__(self, model)
```
NarineK: nit: here the order of calls looks different from previous ones. 
`
LayerDeepLift.__init__(self, model, layer)
DeepLiftShap.__init__(self, model)
`

vivekmig: Will switch, thanks!

PR #184, opened 2019-11-22
Initial Support for Captum Type Hints

vivekmig:
This adds type hints to a few methods in Captum, as well as setting up mypy static type checking in CircleCI tests for captum.attr and tests.


tests/attr/test_feature_ablation.py
```diff
@@ -183,7 +183,7 @@ def test_error_ablations_per_eval_limit_batch_scalar(self):
         inp = torch.tensor([[2.0, 10.0, 3.0], [20.0, 50.0, 30.0]], requires_grad=True)
         ablation = FeatureAblation(lambda inp: torch.sum(net(inp)).item())
         with self.assertRaises(AssertionError):
-            _ = ablation.attribute(inp, ablations_per_eval=2,)
+            _ = ablation.attribute(inp, ablations_per_eval=2)
```
vivekmig: Had to include this lint change as well as the test_gradient_shap change, since the conda test on CircleCI was not passing without it

captum/attr/_core/integrated_gradients.py
```diff
@@ -27,15 +30,21 @@ def __init__(self, forward_func):
...
-        target=None,
-        additional_forward_args=None,
-        n_steps=50,
-        method="gausslegendre",
-        internal_batch_size=None,
-        return_convergence_delta=False,
-    ):
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
+        baselines: Optional[
+            Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
```
edward-io: Should we define a type alias for this, e.g. 

```
Baseline = Union[Tensor, int, float]
```

?

We can create a new file just for defining Captum types.

captum/attr/_core/integrated_gradients.py
```diff
@@ -27,15 +30,21 @@ def __init__(self, forward_func):
...
+        target: Optional[
+            Union[int, Tuple[int, ...], Tensor, List[Tuple[int, ...]]]
+        ] = None,
+        additional_forward_args: Any = None,
+        n_steps: int = 50,
+        method: str = "gausslegendre",
+        internal_batch_size: Optional[int] = None,
+        return_convergence_delta: bool = False,
+    ) -> Union[
+        Tensor, Tuple[Tensor, ...], Tuple[Union[Tensor, Tuple[Tensor, ...]], Tensor]
```
edward-io: Ditto, type aliases could improve readability here and can be reused for inputs.

captum/attr/_core/layer/layer_activation.py
```diff
@@ -90,11 +102,12 @@ def attribute(
                 >>> attribution = layer_cond.attribute(input)
         """
         with torch.no_grad():
-            return _forward_layer_eval(
+            layer_eval = _forward_layer_eval(
```
edward-io: is this needed to cover all paths for pyre/mypy?

vivekmig: Yup, there was no return outside the with block before.

captum/attr/_core/integrated_gradients.py
vivekmig: Yes, this is a good point! I debated this a bit when I wrote the type hints initially. There are some repetitive patterns like these, which would be more concise to move to a separate file, but the advantage of having it here is that someone just reading the signature has the explicit type information readily available without having to look at any other files to find the type definition of aliases. The type information is often provided in the docstrings, but sometimes not as explicitly as the type hints, for instance, in this case of baselines. I'm open to either approach, what do you think @edward-io @NarineK ?

NarineK: I agree that `Baseline` will improve readability but users need to navigate to `Baseline ` type to figure out what it really is. We could have a comment explaining what it is but that might create redundancy with the documentation. How do other well known python libraries that support type hints do ? Maybe we can do a short research on that ?

tests/attr/test_gradient_shap.py
```diff
@@ -58,10 +58,7 @@ def test_basic_multi_input(self):
     def test_basic_scalar_baselines_multi_input(self):
         inputs = (torch.ones(10, 3), torch.rand(10, 4))
         baselines_scalars = (0.0, 0.0)
-        baselines_tensors = (
-            torch.zeros(1, 3),
-            torch.zeros(10, 4),
-        )
+        baselines_tensors = (torch.zeros(1, 3), torch.zeros(10, 4))
```
NarineK: Was this because we had different types of `black` ?

captum/attr/_utils/attribution.py
```diff
@@ -20,20 +20,21 @@ class Attribution:
     to extend and override core `attribute` method.
     """
 
-    def attribute(self, inputs, **kwargs):
+    def attribute(self, inputs):
```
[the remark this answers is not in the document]
NarineK: I see, but if it gets overridden then the signature changes. Maybe there is a different way of specifying multiple arbitrary arguments ?

PR #185, opened 2019-11-23
Feature/permutation feature importance

miguelmartin75:
Implementation of blackbox feature permutation importance algorithm: https://christophm.github.io/interpretable-ml-book/feature-importance.html

Pseudo-code of the above, for a single batch of data:

```
perm_feature_importance(batch):
     feature_importance = {}
     initial_err = f(batch)
     for feature in features:
           batch_copy = permute_feature_across(batch, feature)
           err = f(batch_copy)
           diff = initial_err - err
           feature_importance[feature] = diff

     return feature_importance
           
permute_feature_across(batch, feature):
      batch_copy = batch.clone()
      batch_copy[:, feature] = batch[random_perm(n), feature] 
      return batch_copy
```

The basic idea is to permute each feature in a batch and see how different the output (or error/loss) is from the output/error/loss of the original batch. If the error increases => we know that this feature is important. If the error remains about the same => we know that the feature is not that important since the model was not very reliant on it. Finally, if the error decreases => we know likely know that the feature was negatively impacting the model's performance and hence was detrimental.

error/loss functions must be provided in the `forward_func`. See the test cases for examples.

### DIFF
--- a/captum/attr/_core/perm_feature_importance.py
+++ b/captum/attr/_core/perm_feature_importance.py
@@ -13,8 +13,9 @@ def _permute_feature(x, feature_mask):
     while (perm == no_perm).all():
         perm = torch.randperm(n)
 
-    return (x[perm] * feature_mask.to(dtype=x.dtype)) +\
-           (x * feature_mask.bitwise_not().to(dtype=x.dtype))
+    return (x[perm] * feature_mask.to(dtype=x.dtype)) + (
+        x * feature_mask.bitwise_not().to(dtype=x.dtype)
+    )
 
 
 class PermutationFeatureImportance(FeatureAblation):
--- a/captum/attr/_core/perm_feature_importance.py
+++ b/captum/attr/_core/perm_feature_importance.py
@@ -4,7 +4,7 @@ import torch
 from .feature_ablation import FeatureAblation
 
 
-def permute_feature(x, feature_mask):
+def _permute_feature(x, feature_mask):
     n = x.size(0)
     assert n > 1, "cannot permute features with batch_size = 1"
 
@@ -21,7 +21,7 @@ def permute_feature(x, feature_mask):
 
 
 class PermutationFeatureImportance(FeatureAblation):
-    def __init__(self, forward_func=None, perm_fn=permute_feature):
+    def __init__(self, forward_func=None, perm_fn=_permute_feature):
         FeatureAblation.__init__(self, forward_func=forward_func)
         self.perm_fn = perm_fn
 
--- a/captum/attr/_core/perm_feature_importance.py
+++ b/captum/attr/_core/perm_feature_importance.py
@@ -13,11 +13,7 @@ def _permute_feature(x, feature_mask):
     while (perm == no_perm).all():
         perm = torch.randperm(n)
 
-    out = x.clone()
-    for i, j in enumerate(perm):
-        out[i, feature_mask] = x[j, feature_mask]
-
-    return out
+    return (x[perm] * feature_mask) + (x * feature_mask.bitwise_not())
 
 
 class PermutationFeatureImportance(FeatureAblation):
--- a/captum/attr/_core/perm_feature_importance.py
+++ b/captum/attr/_core/perm_feature_importance.py
@@ -32,9 +32,9 @@ class PermutationFeatureImportance(FeatureAblation):
         additional_forward_args=None,
         feature_mask=None,
         ablations_per_eval=1,
-        abs_attributions=False,
     ):
-        attribs = super().attribute(
+        return FeatureAblation.attribute(
+            self,
             inputs,
             baselines=None,
             target=target,
@@ -42,13 +42,6 @@ class PermutationFeatureImportance(FeatureAblation):
             feature_mask=feature_mask,
             ablations_per_eval=ablations_per_eval,
         )
-
-        f = torch.abs if abs_attributions else lambda x: x
-
-        if isinstance(attribs, tuple):
-            return tuple([f(a) for a in attribs])
-
-        return f(attribs)
 
     def _construct_ablated_input(
         self, feature_tensor, input_mask, baseline, start_feature, end_feature
--- a/captum/attr/_core/perm_feature_importance.py
+++ b/captum/attr/_core/perm_feature_importance.py
@@ -26,11 +26,12 @@ class PermutationFeatureImportance(FeatureAblation):
 
     def attribute(
         self,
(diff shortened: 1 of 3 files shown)


NarineK:
Thank you, @miguelmartin75 ! Would you, please, also add a brief description of the PR before the TODO list ?

captum/attr/_core/perm_feature_importance.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class PermutationFeatureImportance(FeatureAblation):
+    def __init__(self, forward_func=None, perm_fn=permute_feature):
+        super().__init__(forward_func=forward_func)
+        self.perm_fn = perm_fn
+
+    def attribute(
+        self,
+        inputs=None,
```
NarineK: We assume that input must be provided in all attribution algorithms, meaning that we can have `inputs` without defaulting it to `None`

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, x, y):
+        return x[:, 0] * x[:, 1] + y[:, 0] * y[:, 1]
+
+
+class Test(BaseTest):
+    def test_perm_fn(self):
+        for bs in [2, 10, 100]:
```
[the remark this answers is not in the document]
miguelmartin75: Should I simply assert, or throw an exception in this case? If I assert there will be no need to add a test case.

NarineK:
Thank you for the update, @miguelmartin75 ! I'll take a look. It looks like the linter is failing. Also, it would be good to add a short description, in case someone in the open source community opens this PR(even after merging or closing) and wants to understand what it is about described in couple sentences.

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+        self.assertTrue(attribs[1].squeeze(0).size() == input_size)
+
+        for i in range(2):
+            self.assertTrue((attribs[0][:, i] != 0).all())
+            if i != 1:
+                self.assertTrue((attribs[1][:, i] != 0).all())
+
+        self.assertTrue((attribs[1][:, 1] == 0).all())
+
+    def test_mulitple_ablations(self):
```
NarineK: nit: multiple permutations, perhaps ? Because we are testing permutation.

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+            unpermuted_features = mask.bitwise_not()
+            self.assertTrue((inp[:, mask] != perm_inp[:, mask]).any())
+            self.assertTrue(
+                (inp[:, unpermuted_features] == perm_inp[:, unpermuted_features]).all()
+            )
+
+        sizes_to_test = [(10,), (4, 5), (3, 4, 5), (6, 7, 8, 9)]
+        for batch_size in [2, 10]:
+            for inp_size in sizes_to_test:
+                inp = torch.randn((batch_size,) + inp_size)
```
NarineK: The nested for -loops might lead to test timeouts. It is better to avoid nested for-loops in the tests and instead split it up into more tests.

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+                flat_mask = torch.zeros_like(inp[0]).flatten().bool()
+
+                # test random set of single features
+                num_features = inp.numel() // batch_size
+                num_features_to_test = min(num_features, random.randint(2, 30))
```
NarineK: it is better to use a fixed number instead of `random.randint(2, 30)`

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+                    self.assertTrue(torch.sum(flat_mask) == 1)
+
+                    # permute the feature
+                    mask = flat_mask.view_as(inp[0])
+                    perm_inp = permute_feature(inp, mask)
+                    check_features_are_permuted(inp, perm_inp, mask)
+
+                    flat_mask[feature_idx] = 0
+
+                # test random set of features
```
NarineK: This could be a separate test.

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+class Test(BaseTest):
+    def test_perm_fn(self):
+        def check_features_are_permuted(inp, perm_inp, mask):
+            unpermuted_features = mask.bitwise_not()
+            self.assertTrue((inp[:, mask] != perm_inp[:, mask]).any())
+            self.assertTrue(
+                (inp[:, unpermuted_features] == perm_inp[:, unpermuted_features]).all()
+            )
+
+        sizes_to_test = [(10,), (4, 5), (3, 4, 5), (6, 7, 8, 9)]
```
NarineK: Can we also add a separate test case that calls `PermutationFeatureImportance(...).attribute(...) ` for the len(input.shape) > 2 ?

tests/attr/test_perm_feature_importance.py
```diff
@@ -0,0 +1,186 @@
...
+        for masks in masks_to_test:
+            target = random.randint(0, 2)
+            attribs = feature_importance.attribute(
+                (inp1, inp2), feature_mask=masks, target=target
+            )
+            self.assertTrue(isinstance(attribs, tuple))
+
+            features_in_mask = set()
+            for mask in masks:
+                fvs = mask
```
NarineK: nit: It would be good to simplify this test case

captum/attr/_core/perm_feature_importance.py
```diff
@@ -0,0 +1,59 @@
...
+    assert n > 1, "cannot permute features with batch_size = 1"
+
+    perm = torch.randperm(n)
+    no_perm = torch.arange(n)
+    while (perm == no_perm).all():
+        perm = torch.randperm(n)
+
+    out = x.clone()
+    for i, j in enumerate(perm):
+        out[i, feature_mask] = x[j, feature_mask]
```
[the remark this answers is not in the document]
miguelmartin75: Using arithmetic now, but my code will require torch >=1.4 (maybe >=1.3) unless I convert the mask to the same dtype as the input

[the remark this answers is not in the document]
miguelmartin75: Using a .to in order for tests to pass (for older versions of pytorch)

PR #186, opened 2019-11-23
Adding Layer Ablation

vivekmig:
This PR adds layer ablation, which computes the attribution of the input / output of a layer, by setting each neuron or group of neurons to a given baseline and measuring the difference in output value. This is analogous to feature ablation for layer attribution, with baselines and feature masks given with respect to the layer input / output size.

Differential Revision: D18670637

### DIFF
--- a/captum/attr/_core/layer/layer_feature_ablation.py
+++ b/captum/attr/_core/layer/layer_feature_ablation.py
@@ -1,0 +1,272 @@
+#!/usr/bin/env python3
+import torch
+from torch.nn.parallel.scatter_gather import scatter
+
+from ..._utils.attribution import LayerAttribution, PerturbationAttribution
+from ..._utils.common import (
+    _format_input,
+    _format_additional_forward_args,
+    _run_forward,
+    _extract_device,
+)
+from ..._utils.gradient import _forward_layer_eval
+
+from ..feature_ablation import FeatureAblation
+
+
+class LayerFeatureAblation(LayerAttribution, PerturbationAttribution):
+    def __init__(self, forward_func, layer, device_ids=None):
+        r"""
+        Args:
+
+            forward_func (callable):  The forward function of the model or any
+                          modification of it
+            layer (torch.nn.Module): Layer for which attributions are computed.
+                          Output size of attribute matches this layer's input or
+                          output dimensions, depending on whether we attribute to
+                          the inputs or outputs of the layer, corresponding to
+                          attribution of each neuron in the input or output of
+                          this layer.
+                          Currently, it is assumed that the inputs or the outputs
+                          of the layer, depending on which one is used for
+                          attribution, can only be a single tensor.
+            device_ids (list(int)): Device ID list, necessary only if forward_func
+                          applies a DataParallel model. This allows reconstruction of
+                          intermediate outputs from batched results across devices.
+                          If forward_func is given as the DataParallel model itself
+                          (or otherwise has a device_ids attribute with the device
+                          ID list), then it is not necessary to provide this
+                          argument.
+        """
+        LayerAttribution.__init__(self, forward_func, layer, device_ids)
+        PerturbationAttribution.__init__(self, forward_func)
+
+    def attribute(
+        self,
+        inputs,
+        layer_baselines=None,
+        target=None,
+        additional_forward_args=None,
+        layer_mask=None,
+        attribute_to_layer_input=False,
+        ablations_per_eval=1,
+    ):
+        r"""
+            A perturbation based approach to computing layer attribution, involving
+            replacing values in the input / output of a layer with a given baseline /
+            reference, and computing the difference in output. By default, each
+            neuron (scalar input / output value) within the layer is replaced
+            independently.
+            Passing a layer mask allows grouping neurons to be
+            ablated together.
+            Each neuron in the group will be given the same attribution value
+            equal to the change in target as a result of ablating the entire neuron
+            group.
+
+            Args:
+
+                inputs (tensor or tuple of tensors):  Input for which layer
+                            attributions are computed. If forward_func takes a single
+                            tensor as input, a single input tensor should be provided.
+                            If forward_func takes multiple tensors as input, a tuple
+                            of the input tensors should be provided. It is assumed
+                            that for all given input tensors, dimension 0 corresponds
+                            to the number of examples, and if multiple input tensors
+                            are provided, the examples must be aligned appropriately.
+                layer_baselines (scalar, tensor, tuple of scalars or tensors, optional):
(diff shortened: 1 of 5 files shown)


NarineK:
Let's add a brief description to this one and Neuron Ablation as well.

captum/attr/_core/layer/layer_ablation.py
```diff
@@ -0,0 +1,273 @@
...
+            >>> # group (0 - 2) per example are the same.
+            >>> # The attributions can be calculated as follows:
+            >>> # layer mask has dimensions 1 x 3 x 3
+            >>> layer_mask = torch.tensor([[[0,0,0],[1,1,1],
+            >>>                             [2,2,2]]])
+            >>> attr = ablator.attribute(input, target=1,
+            >>>                          layer_mask=layer_mask)
+        """
+
+        def layer_forward_func(*args):
```
NarineK: nit: could this be passed in 2 arguments, like ?
```
layer_input =inputs
original_inputs = additional_forward_args
```

captum/attr/_core/layer/layer_ablation.py
```diff
@@ -0,0 +1,273 @@
...
+            if (
+                device_ids is None
+                and isinstance(self.forward_func, torch.nn.DataParallel)
+                and self.forward_func.device_ids is not None
+            ):
+                device_ids = self.forward_func.device_ids
+
+            all_layer_inputs = {}
+            if device_ids is not None:
+                scattered_layer_input, _ = scatter_kwargs(
```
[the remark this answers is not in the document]
vivekmig: Good point, will update!

captum/attr/_core/layer/layer_ablation.py
```diff
@@ -0,0 +1,273 @@
...
+            layer_input = args[0]
+            original_inputs = args[1:]
+
+            device_ids = self.device_ids
+            if (
+                device_ids is None
+                and isinstance(self.forward_func, torch.nn.DataParallel)
+                and self.forward_func.device_ids is not None
+            ):
+                device_ids = self.forward_func.device_ids
```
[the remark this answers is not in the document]
vivekmig: Good point, the getattr syntax is shorter! I think it might be good to still leave the DataParallel instance check though, it's possible a model might have that attribute and not be DataParallel / not need to be scattered?

captum/attr/_core/layer/layer_ablation.py
vivekmig: This would work if layer_input is guaranteed to be 1 tensor / element as it is now, but once that becomes variable, I'm not sure how to cleanly separate it, since _run_forward in feature ablation will concatenate the inputs and additional_args tuples?

captum/attr/_core/layer/layer_ablation.py
```diff
@@ -0,0 +1,273 @@
...
+            (inputs + additional_forward_args)
+            if additional_forward_args is not None
+            else inputs
+        )
+
+        ablator = FeatureAblation(layer_forward_func)
+
+        return ablator.attribute(
+            layer_eval,
+            baselines=baselines,
```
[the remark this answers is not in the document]
vivekmig: Makes sense, yeah, it's a different here from baselines in other algorithms. I think since we want to evaluate essentially "turning off" each layer input / output, it's not clear what the layer values would be for a given input baseline. Maybe we can document the difference between this baseline and others more clearly? Or rename the argument to something like layer_baseline instead?

vivekmig:
Thanks for the review! Addressed most of the comments.
1. I've renamed to LayerFeatureAblation and also refer to ablating groups of layer input / output elements. I think the confusion in naming is based on defining each neuron as a unique set of weights as opposed to each scalar in the layer as an input / output "neuron", so hopefully this is clearer!

2. Makes sense, yes, this isn't completely consistent with baselines for Layer IG and Conductance, but I do think it's preferable to keep it this way here, since we want to essentially "turn off" each layer input / output neuron, and that's more difficult with an input baseline. We could rename the baselines argument to layer_baselines to make the distinction clearer, what do you think?

captum/attr/_core/layer/layer_feature_ablation.py
```diff
@@ -0,0 +1,270 @@
...
+            all_layer_inputs = {}
+            if device_ids is not None:
+                scattered_layer_input = scatter(layer_input, target_gpus=device_ids)
+                for tensor in scattered_layer_input:
+                    all_layer_inputs[tensor.device] = tensor
+            else:
+                all_layer_inputs[layer_input.device] = layer_input
+
+            def forward_hook(module, inp, out=None):
+                if inp[0].device not in all_layer_inputs:
```
NarineK: I moved `_extract_device` to common - we could use it here too ? https://github.com/pytorch/captum/pull/222/files#diff-3ff60ae8a182cf817df47a82964d8730R438

captum/attr/_core/layer/layer_feature_ablation.py
```diff
@@ -0,0 +1,270 @@
...
+
+        def layer_forward_func(*args):
+            layer_input = args[0]
+            original_inputs = args[1:]
+
+            device_ids = self.device_ids
+            if device_ids is None and isinstance(
+                self.forward_func, torch.nn.DataParallel
+            ):
+                device_ids = getattr(self.forward_func, "device_ids", None)
```
NarineK: I'm not able to comment in our previous conversation thread about the `device_ids ` - I'll comment here:
I think currently we have some places where we check with `torch.nn.DataParallel` and some cases where we don't (e.g in `gradient.py`).
If we don't explicitly check for `torch.nn.DataParallel` then it is more generic but the user might not always want it as you mentioned. 
But from other point of view they can always set `self.forward_func.device_ids=None` if they don't want the computations to be performed in a distributed manner ?

captum/attr/_core/layer/layer_feature_ablation.py
vivekmig: Sounds good, I will wait for that PR to land and then update

captum/attr/_core/layer/layer_feature_ablation.py
vivekmig: Makes sense, sounds good, it would be good to choose one approach and stay consistent. Should we not check for torch.nn.DataParallel everywhere when accessing device_ids and update documentation to reflect this?

NarineK: Yeah, probably that would be good - in case we want to also support ModelParallel, DistributedDataParallel in the future and custom forward functions.

vivekmig: Sounds good

PR #188, opened 2019-11-23
Adding Neuron Ablation

vivekmig:
This PR adds neuron ablation, which allows computing feature ablation for the input or output of a particular neuron. Each input feature (or group) is replaced with the corresponding baseline and the difference in neuron value is computed.

Differential Revision: D18670638

### DIFF
--- a/captum/attr/__init__.py
+++ b/captum/attr/__init__.py
@@ -17,6 +17,7 @@ from ._core.layer.internal_influence import InternalInfluence  # noqa
 from ._core.layer.grad_cam import LayerGradCam  # noqa
 from ._core.layer.layer_deep_lift import LayerDeepLift, LayerDeepLiftShap  # noqa
 from ._core.layer.layer_gradient_shap import LayerGradientShap  # noqa
+from ._core.neuron.neuron_feature_ablation import NeuronFeatureAblation  # noqa
 from ._core.neuron.neuron_conductance import NeuronConductance  # noqa
 from ._core.neuron.neuron_gradient import NeuronGradient  # noqa
 from ._core.neuron.neuron_integrated_gradients import NeuronIntegratedGradients  # noqa
@@ -60,6 +61,7 @@ __all__ = [
     "LayerDeepLift",
     "LayerDeepLiftShap",
     "NeuronConductance",
+    "NeuronFeatureAblation",
     "NeuronGradient",
     "NeuronIntegratedGradients",
     "NeuronDeepLift",
--- a/captum/attr/_core/neuron/neuron_feature_ablation.py
+++ b/captum/attr/_core/neuron/neuron_feature_ablation.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+import torch
+
+from ..._utils.attribution import NeuronAttribution, PerturbationAttribution
+from ..._utils.common import _verify_select_column
+from ..._utils.gradient import _forward_layer_eval
+
+from ..feature_ablation import FeatureAblation
+
+
+class NeuronFeatureAblation(NeuronAttribution, PerturbationAttribution):
+    def __init__(self, forward_func, layer, device_ids=None):
+        r"""
+        Args:
+
+            forward_func (callable):  The forward function of the model or any
+                          modification of it
+            layer (torch.nn.Module): Layer for which attributions are computed.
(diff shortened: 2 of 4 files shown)


captum/attr/_core/neuron/neuron_ablation.py
```diff
@@ -0,0 +1,221 @@
...
+import torch
+
+from ..._utils.attribution import NeuronAttribution, PerturbationAttribution
+from ..._utils.common import _verify_select_column
+from ..._utils.gradient import _forward_layer_eval
+
+from ..feature_ablation import FeatureAblation
+
+
+class NeuronAblation(NeuronAttribution, PerturbationAttribution):
```
NarineK: I think the name, `NeuronAblation `, might be a little misleading, since people might understand that it is performing neuron ablation in sense of neuron deletion. Like deleting some of the neurons and observing its impact on the prediction. Would `neuron_feature_ablation.py (NeuronFeatureAblation` be a better name ?

captum/attr/_core/neuron/neuron_ablation.py
```diff
@@ -0,0 +1,221 @@
...
+            >>> feature_mask = torch.tensor([[[0,0,1,1],[0,0,1,1],
+            >>>                             [2,2,3,3],[2,2,3,3]]])
+            >>> attr = ablator.attribute(input, neuron_index=(4,1,2),
+            >>>                          feature_mask=feature_mask)
+        """
+
+        def neuron_forward_func(*args):
+            with torch.no_grad():
+                layer_eval = _forward_layer_eval(
+                    self.forward_func,
```
NarineK: Is this function also supporting `additional_forward_args` ?

captum/attr/_core/neuron/neuron_ablation.py
```diff
@@ -0,0 +1,221 @@
...
+            >>> # group (0, 1, 2, and 3) per example are the same.
+            >>> # The attributions can be calculated as follows:
+            >>> # feature mask has dimensions 1 x 4 x 4
+            >>> feature_mask = torch.tensor([[[0,0,1,1],[0,0,1,1],
+            >>>                             [2,2,3,3],[2,2,3,3]]])
+            >>> attr = ablator.attribute(input, neuron_index=(4,1,2),
+            >>>                          feature_mask=feature_mask)
+        """
+
+        def neuron_forward_func(*args):
```
NarineK: Will additional `kwargs` also be supported here

captum/attr/_core/neuron/neuron_ablation.py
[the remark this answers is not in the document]
vivekmig: Good point, will add!

vivekmig: Makes sense, sounds good, sure NeuronFeatureAblation would work!

captum/attr/_core/neuron/neuron_ablation.py
vivekmig: Yes, I think this should work fine with additional_forward_args, this function is called with concatenated inputs and forward args, so those should all be contained in the args tuple. There are a few tests which have additional_forward args.

captum/attr/_core/neuron/neuron_ablation.py
```diff
@@ -0,0 +1,221 @@
...
+            reference, and computing the difference in the neuron's input / output.
+            By default, each scalar value within
+            each input tensor is taken as a feature and replaced independently. Passing
+            a feature mask, allows grouping features to be ablated together. This can
+            be used in cases such as images, where an entire segment or region
+            can be ablated, measuring the importance of the segment (feature group).
+            Each input scalar in the group will be given the same attribution value
+            equal to the change in target as a result of ablating the entire feature
+            group.
+
```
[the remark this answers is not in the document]
vivekmig: I think I've seen it used as a baseline / comparison in some papers, but not as a primary method. It would be nice to include one though, so if you know of any papers on this, let me know!

captum/attr/_core/neuron/neuron_ablation.py
vivekmig: I don't think we need kwargs here, since inputs + additional_forward_args will always form a tuple, without keyword arguments?

PR #191, opened 2019-11-30
Extend interpretable embedding with multiple args

NarineK:
This PR extends InterpretableEmbeddingBase to be more generic and support nested embeddings of any types and number of inputs.
I've also added test cases and documentation related to this PR.

In addition to that it looks like in the IMDB tutorial the device was missing and some minor fixes. Adding that as well.

### DIFF
--- a/captum/attr/_models/base.py
+++ b/captum/attr/_models/base.py
@@ -5,10 +5,10 @@ import torch
 import warnings
 
 from functools import reduce
-from torch.nn import Embedding
+from torch.nn import Module
 
 
-class InterpretableEmbeddingBase(Embedding):
+class InterpretableEmbeddingBase(Module):
     r"""
         Since some embedding vectors, e.g. word are created and assigned in
         the embedding layers of Pytorch models we need a way to access
@@ -21,11 +21,18 @@ class InterpretableEmbeddingBase(Embedding):
     """
 
     def __init__(self, embedding, full_name):
-        super().__init__(embedding.num_embeddings, embedding.embedding_dim)
+        Module.__init__(self)
+        self.num_embeddings = (
+            embedding.num_embeddings if hasattr(embedding, "num_embeddings") else None
+        )
+        self.embedding_dim = (
+            embedding.embedding_dim if hasattr(embedding, "embedding_dim") else None
+        )
+
         self.embedding = embedding
         self.full_name = full_name
 
-    def forward(self, input):
+    def forward(self, input, *other_inputs, **kwargs):
         r"""
          The forward function of a wrapper embedding layer that takes and returns
          embedding layer. It allows embeddings to be created outside of the model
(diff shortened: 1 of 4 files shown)


captum/attr/_models/base.py
```diff
@@ -5,10 +5,10 @@
 import warnings
 
 from functools import reduce
-from torch.nn import Embedding
+from torch.nn import Module
 
 
-class InterpretableEmbeddingBase(Embedding):
+class InterpretableEmbeddingBase(Module):
```
orionr: Should this now be `InterpretableModuleBase` or is it still only Embedding even though you are inheriting from Module?

captum/attr/_models/base.py
```diff
@@ -21,20 +21,34 @@ class InterpretableEmbeddingBase(Embedding):
     """
 
     def __init__(self, embedding, full_name):
-        super().__init__(embedding.num_embeddings, embedding.embedding_dim)
+        Module.__init__(self)
+        self.num_embeddings = (
+            embedding.num_embeddings if hasattr(embedding, "num_embeddings") else None
+        )
+        self.embedding_dim = (
+            embedding.embedding_dim if hasattr(embedding, "embedding_dim") else None
```
orionr: Can you just use `getattr` for both of these with the None default?

tutorials/IMDB_TorchText_Interpret.ipynb
```diff
@@ -15,14 +15,14 @@
    "source": [
     "This notebook loads pretrained CNN model for sentiment analysis on IMDB dataset. It makes predictions on test samples and interprets those predictions using integrated gradients method.\n",
     "\n",
-    "The model was trained using an open source sentiment analysis tutorials described in: https://github.com/bentrevett/pytorch-sentiment-analysis/blob/master/4%20-%20Convolutional%20Sentiment%20Analysis.ipynb\n  ",
-    "\n  ",
```
orionr: Doesn't seem  like we'd want these changes?

tutorials/IMDB_TorchText_Interpret.ipynb
```diff
@@ -15,14 +15,14 @@
...
+    "The model was trained using an open source sentiment analysis tutorials described in:https://github.com/bentrevett/pytorch-sentiment-analysis/blob/master/4%20-%20Convolutional%20Sentiment%20Analysis.ipynb\n",
+    "  \n",
     "**Note:** Before running this tutorial, please install the spacy package, and its NLP modules for English language."
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 24,
```
orionr: Same? Don't want these?

tutorials/IMDB_TorchText_Interpret.ipynb
```diff
@@ -467,7 +467,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 18,
+   "execution_count": 20,
```
orionr: Maybe just run straight through and save that?

captum/attr/_models/base.py
orionr: Actually, what's the specific motivation for inheriting differently? Seems like you'd want to still just have Embedding per the code. If you still need this please add a good comment about why.

NarineK: Good point! In a general case Embedding doesn't have to extend `nn.Embedding` but an `nn.Module`. An example of it is `BertEmbedding`:https://github.com/huggingface/transformers/blob/master/transformers/modeling_bert.py#L151

captum/attr/_models/base.py
NarineK: Good point! Fixed!

PR #193, opened 2019-12-02
Refactor layer forward functions

NarineK:
There was a common logic in `compute_layer_gradients_and_eval ` and  `_forward_layer_eval_with_neuron_grads` that I need to use also in `LayerIntegratedGradients`.
More specifically, I would like to access `saved_layer` dictionary or be able to externally gather/concatenate the values in `saved_layer`.
I moved some of the common logic to helper functions so that I don't need to duplicate code.

Creating a separate PR so that it doesn't get too convoluted with LayerIntegratedGradients.

### DIFF
--- a/captum/attr/_utils/gradient.py
+++ b/captum/attr/_utils/gradient.py
@@ -142,7 +142,7 @@ def _forward_layer_distributed_eval(
     pass and returns intermediate layer results, stored in a dictionary,
     and optionally also the output of the forward function. The keys in the
     dictionary are the device ids and the values are corresponding intermediate layer
-    results, either the inputs or the outputs of the layer, depending on, whether we set
+    results, either the inputs or the outputs of the layer depending on whether we set
     `attribute_to_layer_input` to True or False.
     This is especially useful when we execute forward pass in a distributed setting,
     using `DataParallel`s for example.
@@ -168,6 +168,10 @@ def _forward_layer_distributed_eval(
             # TODO we need to think what will be the best way of storing eval
             # tensors per device for each input per example. This implementation
             # doesn't support a tuple of inputs
+
+            # Note that cloning behaviour of `eval_tsr` is different
+            # when `forward_hook_with_return` is set to True. This is because
+            # otherwise `backward()` on the last output layer won't execute.
             if forward_hook_with_return:
                 saved_layer[eval_tsr.device] = eval_tsr
                 return eval_tsr.clone()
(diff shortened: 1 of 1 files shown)


captum/attr/_utils/gradient.py
```diff
@@ -128,41 +128,34 @@ def _forward_layer_eval(
...
-    More information regarding the behavior of forward hooks with DataParallel models
-    can be found in the PyTorch data parallel documentation. We maintain the separate
-    evals in a dictionary protected by a lock, analogous to the gather implementation
-    for the core PyTorch DataParallel implementation.
+    r"""
+    A helper function that allows to set a hook on model's `layer`, run the forward
+    pass and returns intermediate layer results, stored in a dictionary,
+    and optionally also the output of the forward function. The keys in the
+    dictionary are the device ids and the values are corresponding intermediate layer
+    results, either the inputs or the outputs of the layer, depending on, whether we set
```
orionr: Nit:  `layer, depending on, whether` ->`layer depending on whether`

captum/attr/_utils/gradient.py
```diff
@@ -175,32 +168,106 @@ def forward_hook(module, inp, out=None):
...
+
     if len(saved_layer) == 0:
         raise AssertionError("Forward hook did not obtain any outputs for given layer")
 
+    if forward_hook_with_return:
+        return saved_layer, output
+    return saved_layer
+
+
+def _gather_distributed_tnsrs(saved_layer, device_ids=None, key_list=None):
```
orionr: Nit: Can we change `tnsrs` to `tensors` since it's only two extra characters?

captum/attr/_utils/gradient.py
```diff
@@ -175,32 +168,106 @@ def forward_hook(module, inp, out=None):
...
+    r"""
+    A helper function to concatenate intermediate layer results stored on
+    different devices in `saved_layer`. `saved_layer` is a dictionary that
+    that contains `device_id` as a key and intermediate layer results (either
```
orionr: Nit: `that contains` -> `contains`, since  extra that on the previous line

captum/attr/_utils/gradient.py
```diff
@@ -175,32 +168,106 @@ def forward_hook(module, inp, out=None):
...
+    the input or the output of the layer) stored on the device corresponding to
+    the key.
+    `key_list` is a list of devices in appropriate ordering for concatenation
+    and in case it is not provided, it is identified based on device ids.
```
orionr: Nit: `and in case it is not provided, it is identified based on device ids.` ->  `if not provided, keys are sorted based on device ids.`

captum/attr/_utils/gradient.py
```diff
@@ -175,32 +168,106 @@ def forward_hook(module, inp, out=None):
...
-        ):
+        if hasattr(forward_fn, "device_ids") and forward_fn.device_ids is not None:
             device_ids = forward_fn.device_ids
         else:
             raise AssertionError(
-                "DataParallel Model Detected, device ID list or DataParallel model"
-                " must be provided for identifying device batch ordering."
+                "Layer tensors are saved on multiple devices, however unable to access"
+                " device ID list from the `forward_fn`. Device ID list must be"
+                " accessible from `forward_fn`. For example, we can be retrieved"
```
orionr: Nit: `For example, we can be retrieved` -> `For example, they can be retrieved`

captum/attr/_utils/gradient.py
```diff
@@ -273,46 +340,20 @@ def compute_layer_gradients_and_eval(
...
-        # Remove unnecessary forward hook.
-        hook.remove()
-
-        if len(saved_layer) == 0:
-            raise AssertionError(
-                "Forward hook did not obtain any outputs for given layer"
-            )
 
         # Multiple devices / keys implies a DataParallel model, so we look for
         # device IDs if given or available from forward function
```
vivekmig: nit: Can replace the below code with _extract_device_ids as well?

PR #207, opened 2019-12-17
Ablation Type Fix

vivekmig:
Adding fix for issue where input type does not necessarily match output type (e.g. int -> float) along with additional tests for this case in both FeatureAblation and Occlusion.

Thanks @miguelmartin75 for finding this bug!


captum/attr/_utils/common.py
```diff
@@ -241,7 +241,7 @@ def _zeros(inputs):
     shape as the `inputs`
 
     """
-    return tuple(0.0 for input in inputs)
+    return tuple(0 for input in inputs)
```
vivekmig: This change was necessary since a long input tensor was being converted to a float tensor when ablated with baseline, which was causing issues.

miguelmartin75: I think the documentation should also be updated, it's a bit mis-leading, I'd personally expect `tuple(torch.zeros_like(input) for input in inputs)` or  at least `(torch.tensor(0) for input in inputs)`. But since this seems to be a utility function, I guess it doesn't matter much.

vivekmig: Good point, will update the documentation!

PR #211, opened 2019-12-19
Add support for input forward arguments for embeddings that are only named

NarineK:
It looks like some of the embedding modules support only named arguments such as the master version of Bert models: https://github.com/huggingface/transformers/blob/855ff0e91d8b3bd75a3b1c1316e2efd814373764/transformers/modeling_bert.py#L733

Making `InterpretableEmbeddingBase` more generic to support arguments which are only named.

### DIFF
--- a/captum/attr/_models/base.py
+++ b/captum/attr/_models/base.py
@@ -28,7 +28,7 @@ class InterpretableEmbeddingBase(Module):
         self.embedding = embedding
         self.full_name = full_name
 
-    def forward(self, input, *other_inputs, **kwargs):
+    def forward(self, *inputs, **kwargs):
         r"""
          The forward function of a wrapper embedding layer that takes and returns
          embedding layer. It allows embeddings to be created outside of the model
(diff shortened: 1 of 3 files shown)


tests/attr/helpers/basic_models.py
```diff
@@ -212,9 +213,9 @@ def __init__(
         self.relu = nn.ReLU()
         self.linear2 = nn.Linear(hidden_dim, output_dim)
 
-    def forward(self, input, another_input=None):
-        embedding1 = self.embedding1(input)
```
NarineK: I was using the same inputs for different embeddings. Changing it to use different inputs for each embedding.

captum/attr/_models/base.py
```diff
@@ -28,32 +28,42 @@ def __init__(self, embedding, full_name):
...
+                    Note that if inputs are specified, it is assumed that the first
+                    argument is the embedding tensor generated using the
+                    `self.embedding` layer using all imput arguments provided in
+                    `inputs` and `kwargs`.
+            **kwargs (Any, optional): Similar to `inputs` we want to make sure
                     that our forward pass supports arbitrary number and type of
-                    key-value arguments
+                    key-value arguments. If `inputs` is not provided, `kwargs` must
+                    be provded and the first argument corresponds to the embedding
+                    tensor generated using the `self.embedding`.
```
vivekmig: It might be good to add a note here that the ordered dict behavior for kwargs which is relied on here is new in python 3.6 and not guaranteed to remain consistent for future versions based on this spec: https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep468

NarineK: added

PR #213, opened 2019-12-19
Add a tutorial section on Occlusion

bilalsal:
As requested in T58506541, demonstrate the Occlusion algorithm on a torchvision model

Test Plan: run the notebook and check the images computed by the new section on Occlusion


NarineK:
Nice! Thank you!
I think `rand_img_dist = torch.cat([input * 0, input * 1])` maybe got copied from the previous cells ?

and this one: `(a defined via strides)`, maybe `a` is not needed ?

Issue #122, opened 2019-10-16
Could I use captum for object localisation?


batrlatom:
Hello,
Can I use this library for object localisation tasks? Would you think you could prepare some very easy tutorial for this? I bet that this would be very helpful for many people since labelling images with bounding boxes or polygons is really time consuming as you know.

NarineK:
Thank you @batrlatom ! I understand! In case of attribution we do not necessarily need to attribute to the labeled classes , we can also attribute to the predicted ones. Do you have any examples and links that you think will be most useful, we will find time and prioritize it ?
Thank you :)

batrlatom:
Hi,
this is great that you find such usecase usefull! The tutorial could be as easy as making gradcam from input image and thresholding inputs so I can obtain most likely bounding box for the predicted 1-n classes to more advanced like using captum in conjuction with attention dropout layer ( https://github.com/junsukchoe/ADL ).

craymichael:
Hi, existing image tutorials now cover the relevant workflow for object-level attribution. In particular, the ResNet feature ablation tutorial uses segmentation masks as feature groups to identify influential object regions, and the segmentation tutorial covers dense image attribution. Closing as documented. https://captum.ai/tutorials/Resnet_TorchVision_Ablation https://captum.ai/tutorials/Segmentation_Interpret

PR #214, opened 2019-12-20
Add feature ablation to the list of tutorials

bilalsal:
@NarineK proposed this tutorial could in part answer issue #122

### DIFF
--- a/website/tutorials.json
+++ b/website/tutorials.json
@@ -13,7 +13,7 @@
     },
     {
       "id": "Resnet_TorchVision_Interpret",
-      "title": "Interpreting vision with Resnet"
+      "title": "Interpreting vision with ResNet"
     },
     {
       "id": "Resnet_TorchVision_Ablation",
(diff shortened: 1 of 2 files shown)


website/tutorials.json
```diff
@@ -13,7 +13,11 @@
     },
     {
       "id": "Resnet_TorchVision_Interpret",
-      "title": "Interpreting vision with Resnet"
+      "title": "Interpreting vision with ResNet"
+    },
+    {
+      "id": "Resnet_TorchVision_Ablation",
+      "title": "Feature ablation on images with ResNet"
```
vivekmig: It might be good to also add a section and short description to website/pages/tutorials/index.js, which will also include it on the main tutorials page, I think this will only add it to the sidebar. (as an example, #198 does both changes for the BERT tutorial).

NarineK: Good point! and actually [master-only] tag would be good too.cc: @vivekmig , @bilalsal 
`
"title": "[master only] Interpreting question answering with BERT"
--
Show all 11 lines

`

PR #217, opened 2019-12-22
Add layer integrated gradients

NarineK:
This PR adds vanilla integrated gradients implementation for layers. This is especially useful for computing IG for word embeddings and it will help us to avoid pre-computing embedding layers separately before the forward pass and wrapping that layer, e.g. in the IMDB tutorial.

The following things will be added in separate PRs:
1. Documentation
2. Tutorial changes using LayerIntegratedGradients for IMDB and Bert models

### DIFF
--- a/captum/attr/__init__.py
+++ b/captum/attr/__init__.py
@@ -18,6 +18,7 @@ from ._core.layer.internal_influence import InternalInfluence  # noqa
 from ._core.layer.grad_cam import LayerGradCam  # noqa
 from ._core.layer.layer_deep_lift import LayerDeepLift, LayerDeepLiftShap  # noqa
 from ._core.layer.layer_gradient_shap import LayerGradientShap  # noqa
+from ._core.layer.layer_integrated_gradients import LayerIntegratedGradients  # noqa
 from ._core.neuron.neuron_feature_ablation import NeuronFeatureAblation  # noqa
 from ._core.neuron.neuron_conductance import NeuronConductance  # noqa
 from ._core.neuron.neuron_gradient import NeuronGradient  # noqa
@@ -35,7 +36,6 @@ from ._models.base import (
     configure_interpretable_embedding_layer,
     remove_interpretable_embedding_layer,
 )  # noqa
-from ._utils import visualization  # noqa
 from ._utils.attribution import Attribution  # noqa
 from ._utils.attribution import GradientAttribution  # noqa
 from ._utils.attribution import LayerAttribution  # noqa
@@ -69,6 +69,7 @@ __all__ = [
     "LayerDeepLift",
     "LayerDeepLiftShap",
     "LayerGradientShap",
+    "LayerIntegratedGradients",
     "NeuronConductance",
     "NeuronFeatureAblation",
     "NeuronGradient",
--- a/captum/attr/__init__.py
+++ b/captum/attr/__init__.py
@@ -70,6 +70,7 @@ __all__ = [
     "LayerDeepLiftShap",
     "LayerGradientShap",
     "LayerIntegratedGradients",
+    "LayerGradientShap",
     "NeuronConductance",
     "NeuronFeatureAblation",
     "NeuronGradient",
--- a/captum/attr/__init__.py
+++ b/captum/attr/__init__.py
@@ -70,7 +70,6 @@ __all__ = [
     "LayerDeepLiftShap",
     "LayerGradientShap",
     "LayerIntegratedGradients",
-    "LayerGradientShap",
     "NeuronConductance",
     "NeuronFeatureAblation",
     "NeuronGradient",
--- a/captum/attr/_core/layer/layer_conductance.py
+++ b/captum/attr/_core/layer/layer_conductance.py
@@ -49,7 +49,7 @@ class LayerConductance(LayerAttribution, GradientAttribution):
         target=None,
         additional_forward_args=None,
         n_steps=50,
-        method="riemann_trapezoid",
+        method="gausslegendre",
         internal_batch_size=None,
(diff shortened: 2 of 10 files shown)


captum/attr/__init__.py
```diff
@@ -35,7 +36,6 @@
     configure_interpretable_embedding_layer,
     remove_interpretable_embedding_layer,
 )  # noqa
-from ._utils import visualization  # noqa
```
NarineK: Somehow there were 2 definitions of this in the `__init__.py`

captum/attr/_core/layer/layer_conductance.py
```diff
@@ -49,7 +49,7 @@ def attribute(
         target=None,
         additional_forward_args=None,
         n_steps=50,
-        method="riemann_trapezoid",
+        method="gausslegendre",
```
NarineK: Changing to `gausslegendre` in order to be consistent with IG and documentation.

captum/attr/_core/layer/layer_integrated_gradients.py
```diff
@@ -0,0 +1,214 @@
...
+            start_point, end_point = baselines, inputs
+            # computes approximation error based on the completeness axiom
+            delta = self.compute_convergence_delta(
+                attributions,
+                start_point,
+                end_point,
+                additional_forward_args=additional_forward_args,
+                target=target,
+            )
+            return _format_attributions(is_inputs_tuple, attributions), delta
```
NarineK: `is_inputs_tuple ` for input is not necessary - will remove it

captum/attr/_core/layer/layer_integrated_gradients.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # gather inputs and baseline layer outputs similar to gradients
+        inputs_layer = _gather_distributed_tensors(
+            inputs_layer_dict, device_ids=self.device_ids
+        )
+        inputs_layer = (inputs_layer,)
+        baselines_layer = _gather_distributed_tensors(
+            baselines_layer_dict, device_ids=self.device_ids
+        )
+        baselines_layer = (baselines_layer,)
```
NarineK: This needs to be properly formatted

captum/attr/_core/layer/layer_integrated_gradients.py
```diff
@@ -0,0 +1,214 @@
...
+        baselines=None,
+        target=None,
+        additional_forward_args=None,
+        n_steps=50,
+        method="gausslegendre",
+        return_convergence_delta=False,
+    ):
+        # Keeps track whether original input is a tuple or not before
+        # converting it into a tuple.
+        is_inputs_tuple = isinstance(inputs, tuple)
```
NarineK: we probably don't need this

captum/attr/_core/layer/layer_integrated_gradients.py
```diff
@@ -0,0 +1,150 @@
...
+                for scattered_input in scattered_inputs
+            }
+
+            with torch.autograd.set_grad_enabled(True):
+
+                def layer_forward_hook(module, hook_inputs, hook_outputs=None):
+                    result = scattered_inputs_dict[hook_inputs[0].device]
+                    return (
+                        result[0]
+                        if isinstance(result, tuple) and hook_outputs is not None
```
vivekmig: With the current implementation of _forward_layer_eval always returning a tensor, will result ever be a tuple? We will also probably need to update this condition when the layer tuple code merges to return the tuple?

tests/attr/layer/test_layer_integrated_gradients.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_compare_with_layer_conductance_attr_to_inputs(self):
+        model = BasicModel_MultiLayer()
+        input = torch.tensor([[50.0, 50.0, 50.0]], requires_grad=True)
+        self._assert_compare_with_layer_conductance(model, input, True)
+
+    def _assert_compare_with_layer_conductance(
+        self, model, input, attribute_to_layer_input=False
+    ):
+        lc = LayerConductance(model, model.linear2)
```
vivekmig: nit: It might be nice to have a comment here explaining that although they are not theoretically equivalent, they are both integrating along paths between the same 2 points, and with only linear transformations prior to the layer they are equivalent? It might be confusing that this property should hold in general.

captum/attr/_core/layer/layer_integrated_gradients.py
```diff
@@ -0,0 +1,150 @@
...
+
+            scattered_inputs_dict = {
+                scattered_input[0].device: scattered_input
+                for scattered_input in scattered_inputs
+            }
+
+            with torch.autograd.set_grad_enabled(True):
+
+                def layer_forward_hook(module, hook_inputs, hook_outputs=None):
+                    result = scattered_inputs_dict[hook_inputs[0].device]
```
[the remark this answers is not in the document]
NarineK: yeah, I think that we can assume that it will never be a 0D tensor because we always assume that there is at least one dimension preserving which has to be at least for the number of examples ?
Let's say:
`hook_inputs  = torch.tensor([5])`
I can still access its device with
`hook_inputs[0].device`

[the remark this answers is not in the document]
vivekmig: Makes sense, ok, yeah this should work generally then!

captum/attr/_core/layer/layer_integrated_gradients.py
NarineK: Oh, no, that was an issue coming from scatter. Fixed it

PR #221, opened 2019-12-26
Add documentation for Layer Integrated Gradients

NarineK:
1. Adds documentation for Layer Integrated Gradients
2. Refactors Integrated Gradients - Moves algorithm description directly under the definition of the class
3. Fixes formatting for IG documentation - r"""  is now more consistent with other docs in IG

### DIFF
--- a/captum/attr/_core/integrated_gradients.py
+++ b/captum/attr/_core/integrated_gradients.py
@@ -16,6 +16,21 @@ from .._utils.attribution import GradientAttribution
 
 
 class IntegratedGradients(GradientAttribution):
+    r"""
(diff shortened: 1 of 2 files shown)


captum/attr/_core/integrated_gradients.py
```diff
@@ -37,142 +53,143 @@ def attribute(
         return_convergence_delta=False,
     ):
         r"""
```
NarineK: Indented the documentation under `r"""`

NarineK: Yeah, I noticed that also in other algorithms. So we can adjust the indention in other methods as well. Docstring will be consistently aligned with r""". What do you think  ?

PR #226, opened 2020-01-05
Fix issue in gradcam and layer_gradient_x_activation when model is in no_requires_grad mode

NarineK:
As I was playing with gradcam and layer_gradient_x_activation, noticed that if the require_grads is set to False for a model, model developers sometimes do that as an optimization, then those two methods fail. To fix that we need to apply `apply_gradient_requirements and undo_gradient_requirements`...

I created some test cases for those cases.
The grads are sometimes disabled with the following command:
```
        for param in net.parameters():
            param.requires_grad = False
```

### DIFF
--- a/captum/attr/_core/layer/layer_gradient_x_activation.py
+++ b/captum/attr/_core/layer/layer_gradient_x_activation.py
@@ -5,8 +5,11 @@ from ..._utils.common import (
     _format_additional_forward_args,
     _format_attributions,
 )
-from ..._utils.gradient import compute_layer_gradients_and_eval
-from ..._utils.gradient import apply_gradient_requirements, undo_gradient_requirements
+from ..._utils.gradient import (
+    compute_layer_gradients_and_eval,
+    apply_gradient_requirements,
+    undo_gradient_requirements,
+)
 
 
 class LayerGradientXActivation(LayerAttribution, GradientAttribution):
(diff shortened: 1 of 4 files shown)


tests/attr/layer/test_layer_gradient_x_activation.py
```diff
@@ -25,6 +25,18 @@ def test_simple_linear_gradient_activation(self):
             net, net.linear1, inp, [90.0, 101.0, 101.0, 101.0]
         )
 
+    def test_simple_linear_gradient_activation_no_grad(self):
+        net = BasicModel_MultiLayer()
+        inp = torch.tensor([[0.0, 100.0, 0.0]])
+        # this way we deactivate require_grad. Some models expictly do that
```
NarineK: explicitly

captum/attr/_core/layer/layer_gradient_x_activation.py
```diff
@@ -6,6 +6,7 @@
     _format_attributions,
 )
 from ..._utils.gradient import compute_layer_gradients_and_eval
+from ..._utils.gradient import apply_gradient_requirements, undo_gradient_requirements
```
NarineK: Combine with the above import

vivekmig:
Looks great, thanks for adding this :)!!

PR #233, opened 2020-01-13
Create config info for captum insights

J0Nreynolds:
Summary:
Creates a file config.py that holds information about Captum Insights's supported Attribution methods and their corresponding argument types/limits.

FilterConfig is refactored to contain a chosen attribution method class, as well as whatever keyword arguments need to be passed to the attribution method's `attribute` function.

The widget.py and server.py files now convert and pass the config info to the frontend in a readable format. They convert to the expected FilterConfig format upon receiving the chosen attribution method and arguments.

Differential Revision: D19254210

### DIFF
--- a/captum/attr/_utils/attribution.py
+++ b/captum/attr/_utils/attribution.py
@@ -3,13 +3,14 @@ import torch
 import torch.nn.functional as F
 
 from .common import (
-    _run_forward,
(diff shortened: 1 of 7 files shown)


edward-io:
lets check why the OSS tests aren't passing before landing.

PR #234, opened 2020-01-13
Support debugging mode in CaptumInsights widget

J0Nreynolds:
Summary: This diff makes it simple for users to enable debug mode in the CaptumInsights widget. I've modified the `AttributionVisualizer` to have a `debug` flag in the `render` function. When `debug` is enabled, any output from stdout is redirected to be displayed below the widget.

Reviewed By: edward-io

Differential Revision: D19319091

### DIFF
--- a/captum/attr/_utils/attribution.py
+++ b/captum/attr/_utils/attribution.py
@@ -5,13 +5,13 @@ import torch
 import torch.nn.functional as F
 
 from .common import (
-    _run_forward,
+    _format_additional_forward_args,
     _format_input_baseline,
     _format_tensor_into_tuples,
-    _format_additional_forward_args,
+    _run_forward,
+    _tensorize_baseline,
     _validate_input,
     _validate_target,
-    _tensorize_baseline,
 )
 from .gradient import compute_gradients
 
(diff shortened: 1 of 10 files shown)


captum/attr/_utils/attribution.py
```diff
@@ -111,6 +118,22 @@ def compute_convergence_delta(self, attributions, *args):
             "Deriving sub-class should implement" " compute_convergence_delta method"
         )
 
+    @classproperty
+    def display_name(self):
```
NarineK: `display_name ` sounds UI specific, can we name it more generic  ? Can we actually, look into PyTorch and use similar convention ? For the module they have `get_name()` https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module

Does it have to be a class property can we use static method like this ?

https://github.com/automl/Auto-PyTorch/blob/master/autoPyTorch/core/autonet_classes/autonet_image_classification.py#L20?

captum/attr/_utils/attribution.py
```diff
@@ -3,13 +3,13 @@
 import torch.nn.functional as F
 
 from .common import (
-    _run_forward,
+    _format_additional_forward_args,
     _format_input_baseline,
     _format_tensor_into_tuples,
-    _format_additional_forward_args,
+    _run_forward,
+    _tensorize_baseline,
```
NarineK: nit: it looks like restructured the order of imports. Did black formatting do it ?

captum/attr/_utils/attribution.py
```diff
@@ -111,6 +111,22 @@ def compute_convergence_delta(self, attributions, *args):
...
 
+    @classmethod
+    def get_name(cls):
+        r"""
+        Create readable class name by inserting a space before any capital
+        characters besides the very first.
+
+        Returns:
+            str: a readable class name
+        """
```
NarineK: nit: perhaps you can provide an example of the readable name here ?

captum/attr/_utils/attribution.py
J0Nreynolds: Yes, I believe one of the formatting linters requested this change.

captum/attr/_utils/attribution.py
J0Nreynolds: Updated

PR #235, opened 2020-01-13
Incorporate config in frontend

J0Nreynolds:
Summary:
Modify frontend to use the config information passed from the backend(s).
This means the frontend now
- Displays attribution method options.
- Displays attribution method arguments based on the selected method.
  - Create React components for different argument data types
- Updates and persists arguments even when selected method is changed
- Sends chosen arguments when Fetch button is pressed

Differential Revision: D19254187

### DIFF
--- a/captum/attr/_utils/attribution.py
+++ b/captum/attr/_utils/attribution.py
@@ -3,13 +3,14 @@ import torch
 import torch.nn.functional as F
 
 from .common import (
-    _run_forward,
+    _format_additional_forward_args,
     _format_input_baseline,
     _format_tensor_into_tuples,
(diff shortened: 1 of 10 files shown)


captum/attr/_utils/attribution.py
```diff
@@ -111,6 +118,22 @@ def compute_convergence_delta(self, attributions, *args):
             "Deriving sub-class should implement" " compute_convergence_delta method"
         )
 
+    @classproperty
+    def display_name(self):
```
NarineK: I commented regarding display_name in: #234

edward-io:
LGTM!

PR #240, opened 2020-01-15
add Insights API docs

edward-io:
Add documentation for `insights/features.py` and add it to the main Captum API docs page.

test plan:

Build docs (`./scripts/build_docs.sh`) and verify that they show up correctly on http://localhost:3000/api/.

### DIFF
--- a/captum/insights/features.py
+++ b/captum/insights/features.py
@@ -5,10 +5,10 @@ from collections import namedtuple
 from io import BytesIO
 from typing import Callable, List, Optional, Union
 
+import numpy as np
+
 from captum.attr._utils import visualization as viz
 from captum.attr._utils.common import safe_div
-
-import numpy as np
 
 FeatureOutput = namedtuple("FeatureOutput", "name base modified type contribution")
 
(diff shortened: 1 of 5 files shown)


captum/insights/features.py
```diff
@@ -93,6 +123,12 @@ def visualize(self, attribution, data, contribution_frac) -> FeatureOutput:
 
 
 class TextFeature(BaseFeature):
+    r"""
+    TextFeature is used to visualize text (e.g. sentences) in Insights.
+    It expects the visualization transform to convert the input data (e.g. index to
+    string) to the raw text.
```
vivekmig: This visualization transform hint is useful, might be good to add more particular descriptions / docstrings below for text.

edward-io:
Addressed comments from @vivekmig  and @jlin27. Thanks for the feedback!

captum/insights/features.py
```diff
@@ -44,13 +72,37 @@ def visualize(self, attribution, data, contribution_frac) -> FeatureOutput:
 
 
 class ImageFeature(BaseFeature):
+    r"""
+    ImageFeature is used to visualize image features in Insights. It expects an image in
+    NCHW format. If C has a dimension of 1, its assumed to be a black and white image.
```
vivekmig: nit: maybe greyscale instead of black and white?

PR #241, opened 2020-01-16
Update algorithms.md

aliborji:


edward-io:
@aliborji thanks for the PR. Captum is a repository for implementations of interpretability algorithms, we would be happy to have your new method added as a separate pull request. `Algorithms.md` is just a listing of existing algorithms we have implemented in the codebase.

NarineK:
Hi @aliborji, Thank you very much for the PR. Your work looks very interesting and we would like to consider it for Captum. 
Two important remarks:
1. As @edward-io  mentioned we would like to have the core implementation of your algorithm in Captum and additional jupyter notebooks as tutorials.
2. Currently, we only have attribution algorithms under `attr` package but we plan to add a new package under the `_core` where we will collect optimization algorithms. We'd like to propose you to add your approach under that package. We will create that package under the `_core`, most probably named `optim`, with all necessary configuration and design and we'll invite you to add the core logic of your implementation. Does that sound good to you ?

At this point we'd suggest to close the PR and we can think of adding the core logic into Captum. What do you think ?

edward-io:
Closing PR due to inactivity.

PR #244, opened 2020-01-17
Type Improvements and Initial Neuron Method Typing

vivekmig:
This PR makes improvements to type hints, particularly adding overload signatures to methods such as IG, simplifying the output types for specific input types (e.g. tensor -> tensor). This is necessary for dependencies to properly identify the returned type rather than considering a variety of type outputs.

We also remove the base attribute method, which was causing issues due to violating Liskov substitutability, and replaced it with a type hint for the attribute.

Finally, type hints were added for neuron methods, particularly neuron Gradient Shap (along with tests) and neuron integrated gradients.

### DIFF
--- a/captum/attr/_core/integrated_gradients.py
+++ b/captum/attr/_core/integrated_gradients.py
@@ -17,7 +17,7 @@ from .._utils.common import (
     _expand_target,
 )
 from .._utils.attribution import GradientAttribution
-from .._utils.typing import TensorOrTuple
+from .._utils.typing import TensorOrTupleOfTensors
 
 
 class IntegratedGradients(GradientAttribution):
@@ -53,7 +53,7 @@ class IntegratedGradients(GradientAttribution):
     @typing.overload
     def attribute(
         self,
-        inputs: TensorOrTuple,
+        inputs: TensorOrTupleOfTensors,
         baselines: Optional[
             Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
         ] = None,
@@ -64,13 +64,13 @@ class IntegratedGradients(GradientAttribution):
         n_steps: int = 50,
         method: str = "gausslegendre",
         internal_batch_size: Optional[int] = None,
-    ) -> TensorOrTuple:
+    ) -> TensorOrTupleOfTensors:
         ...
 
     @typing.overload
     def attribute(
         self,
-        inputs: TensorOrTuple,
+        inputs: TensorOrTupleOfTensors,
         baselines: Optional[
             Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
         ] = None,
@@ -82,7 +82,7 @@ class IntegratedGradients(GradientAttribution):
         method: str = "gausslegendre",
         internal_batch_size: Optional[int] = None,
         return_convergence_delta: bool = False,
-    ) -> Union[TensorOrTuple, Tuple[TensorOrTuple, Tensor]]:
+    ) -> Union[TensorOrTupleOfTensors, Tuple[TensorOrTupleOfTensors, Tensor]]:
         ...
 
     def attribute(
--- a/captum/attr/_core/integrated_gradients.py
+++ b/captum/attr/_core/integrated_gradients.py
@@ -17,6 +17,7 @@ from .._utils.common import (
     _expand_target,
 )
 from .._utils.attribution import GradientAttribution
+from .._utils.typing import TensorOrTuple
 
 
 class IntegratedGradients(GradientAttribution):
@@ -52,7 +53,7 @@ class IntegratedGradients(GradientAttribution):
     @typing.overload
     def attribute(
         self,
-        inputs: Tensor,
+        inputs: TensorOrTuple,
         baselines: Optional[
             Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
         ] = None,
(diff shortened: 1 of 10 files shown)


tests/attr/neuron/test_neuron_gradient_shap.py
```diff
@@ -53,9 +63,9 @@ def _assert_attributions(
...
 
         attrs_ig = []
-        for baseline in baselines:
+        for baseline in baselines.__iter__():
             attrs_ig.append(
                 nig.attribute(inputs, neuron_ind, baselines=baseline.unsqueeze(0))
             )
-        attrs_ig = torch.stack(attrs_ig, axis=0).mean(axis=0)
-        assertTensorAlmostEqual(self, attrs_gs, attrs_ig, 0.5)
+        combined_attrs_ig = torch.stack(attrs_ig, dim=0).mean(dim=0)
```
vivekmig: From PyTorch docs, seems like this parameter should have been dim not axis

scripts/install_via_pip.sh
```diff
@@ -50,7 +50,7 @@ fi
 
 # install pytorch nightly if asked for
 if [[ $PYTORCH_NIGHTLY == true ]]; then
-  sudo pip install torch_nightly -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+  sudo pip install --upgrade --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
```
edward-io: can you add what these changes are to your PR?

tests/attr/neuron/test_neuron_gradient_shap.py
```diff
@@ -53,9 +63,9 @@ def _assert_attributions(
             baselines = baselines(inputs)
 
         attrs_ig = []
-        for baseline in baselines:
+        for baseline in baselines.__iter__():
```
NarineK: is the `__iter__()` necessary here ?

vivekmig: Whoops, good point, I was testing because this was causing type errors, but it was an issue with torch 1.3, will remove this, thanks!!

captum/attr/_core/neuron/neuron_integrated_gradients.py
```diff
@@ -30,17 +40,45 @@ def __init__(self, forward_func, layer, device_ids=None):
         NeuronAttribution.__init__(self, forward_func, layer, device_ids)
         GradientAttribution.__init__(self, forward_func)
 
+    @typing.overload
```
NarineK: nit: in order to not repeat typing it would be good to import `typing` everywhere and just call `@overload`  ?

captum/attr/_core/neuron/neuron_integrated_gradients.py
```diff
@@ -30,17 +40,45 @@ def __init__(self, forward_func, layer, device_ids=None):
...
+        n_steps: int = 50,
+        method: str = "gausslegendre",
+        internal_batch_size: Optional[int] = None,
+        attribute_to_neuron_input: bool = False,
+    ) -> Tensor:
+        ...
+
+    @typing.overload
+    def attribute(
+        self,
```
NarineK: It looks like we have multiple copies of `def attribute`  and duplicated code. I think that the original solution was probably better. Why did we decided to change it ?

captum/attr/_core/neuron/neuron_integrated_gradients.py
vivekmig: Yes, I initially had that, the problem is that flake8 doesn't recognize the overloading appropriately and errors with just @overload, more details are here https://github.com/PyCQA/pyflakes/issues/320. The fix has been added to pyflakes, but isn't yet in the latest version of flake8, once it gets added we can switch back to just @overload.

captum/attr/_core/neuron/neuron_integrated_gradients.py
vivekmig: The original constructor / type hints are still there, this just adds variants that show that the return type will be a Tensor when a Tensor input is given, etc. The reason this was needed for Integrated Gradients was that the return type is quite complex, and this causes errors when type checking code calling IG that assumes a particular return type, while the generic one contains a union of a variety of options. The callers generally assume one of the known behaviors, e.g. providing a single tensor with no compute_convergence_delta flag always returns a single tensor. By including this, the calling code can be type checked with the appropriate return type as opposed to the union of all possible return types. This was causing issues in both insights as well as neuron IG, which is why I added it to IG as well as here. Unfortunately, it does require duplicating the method signature (but at least none of the implementation, so it's just a decoration that doesn't affect runtime), so in general, it might be good to include these signatures whenever possible, since they help to type check code using Captum methods more accurately.

vivekmig: For now, I've left the original type hints as well as the overload signatures to be consistent with other methods, but we can consider removing the main definition type hints with unions and just leaving the overload signatures if that looks cleaner?

scripts/install_via_pip.sh
vivekmig: The torch nightly build name changed, so this was actually still using an old nightly version from last August, which was still torch 1.3. I've updated this command, so it now obtains the latest nightly build.

captum/attr/_core/integrated_gradients.py
```diff
@@ -16,6 +17,7 @@
     _expand_target,
 )
 from .._utils.attribution import GradientAttribution
+from .._utils.typing import TensorOrTuple
```
NarineK: nit: let's rename it to `TensorOrTupleOfTensors` - then it is very explicit and users don't have to think what is in the tuple ?

vivekmig: Sounds good, yeah, I agree that would be better naming 👍!

vivekmig:
Sounds good, will add the comment, thanks!

PR #247, opened 2020-01-22
Add entropy, pdfs and various refactoring

NarineK:
1. Renamed `target_class` to `true_class` in the `visualization.py` for the text visualization because `target_class` was confused / mistaken by the target in the attribute method
2. Originally I was using `bert-base-cased` pre-trained model but it doesn't seem to work anymore, now I changed to `bert-base-uncased` and this slightly affected prediction / attribution results - updated those changes
3. Added plots for pdfs and Entropy


NarineK:
Good points! Thank you! Addressed the comments.

PR #250, opened 2020-01-23
Adding CircleCI Configs for PyTorch 1.2 and 1.3

vivekmig:
This PR adds test configurations on CircleCI for PyTorch 1.2 and 1.3, which we still support, and updates the existing Conda and PIP tests for torch nightly to use the latest nightly builds.

### DIFF
--- a/scripts/install_via_conda.sh
+++ b/scripts/install_via_conda.sh
@@ -30,7 +30,7 @@ fi
 
 if [[ $PYTORCH_NIGHTLY == true ]]; then
   # install CPU version for much smaller download
-  conda install -y -c pytorch pytorch-nightly-cpu
+  conda install -y pytorch cpuonly -c pytorch-nightly
 else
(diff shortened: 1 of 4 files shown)


scripts/install_via_conda.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 if [[ $PYTORCH_NIGHTLY == true ]]; then
   # install CPU version for much smaller download
-  conda install -y -c pytorch pytorch-nightly-cpu
+  conda install -y pytorch cpuonly -c pytorch-nightly
```
edward-io: unrelated to this change, but maybe we should rename install_via_conda to something like circle_ci_test_via_conda, in case users try to execute the script.

vivekmig: Sounds good, yeah that's a good point! I think this originated from copying Botorch's structure, but will change naming in a separate PR.

PR #251, opened 2020-01-23
Sparse Updates for Permutation Importance

vivekmig:
This PR adds support for sparse 1D tensors (which may not match the number of examples) and empty tensors (which will have empty attributions) to both feature ablation and permutation feature importance with corresponding test cases.

Differential Revision: D19539757

### DIFF
--- a/captum/attr/_core/feature_ablation.py
+++ b/captum/attr/_core/feature_ablation.py
@@ -284,6 +284,9 @@ class FeatureAblation(PerturbationAttribution):
 
             # Iterate through each feature tensor for ablation
             for i in range(len(inputs)):
+                # Skip any empty input tensors
+                if torch.numel(inputs[i]) == 0:
+                    continue
                 for (
                     current_inputs,
                     current_add_args,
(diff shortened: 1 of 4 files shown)


captum/attr/_core/feature_ablation.py
```diff
@@ -284,6 +284,9 @@ def attribute(
 
             # Iterate through each feature tensor for ablation
             for i in range(len(inputs)):
+                # Skip any empty input tensors
+                if torch.numel(inputs[i]) == 0:
```
NarineK: For the empty input, can't the use pass it as an additional forward args since the attribution is empty too ?

vivekmig: The issue is that this (sparse) tensor may not be empty for all batches, so to pass a consistent structure, it would be good to have this.

NarineK: I see,  makes sense

tests/attr/test_perm_feature_importance.py
```diff
@@ -193,3 +198,33 @@ def forward_func(x):
...
+        total_attr2 = 0
+        for _ in range(50):
+            attr1, attr2 = feature_importance.attribute((inp1, inp2))
+            total_attr1 += attr1
+            total_attr2 += attr2
+        total_attr1 /= 50
+        total_attr2 /= 50
+        self.assertEqual(total_attr2.shape, (1,))
+        assertTensorAlmostEqual(self, total_attr1, [0.0, 0.0, 0.0])
+        assertTensorAlmostEqual(self, total_attr2, [-6.0], delta=0.2)
```
NarineK: oh, so it is summing across all sparse indices `[1, 7, 2, 4, 5, 3, 6]` for input2 ? It sees the sparse tensor as one feature ?

vivekmig: Yup, it sees the sparse tensor as one feature, and the model uses the first index.

PR #255, opened 2020-01-28
Fix for Captum to be served in Colab notebooks

J0Nreynolds:
Summary:
This change uses TensorBoard's example of Colab serving to support serving Captum Insights in Colab. The main changes are:
- Display a script in Colab which does the following upon execution
  - Set the base URL to localhost:PORT where PORT is the port that the Insights server is using (this ensures requests are directed properly)
  - Fetch "." and get the HTML response
  - Insert the HTML into the DOM
  - For each script and css file in the HTML, remove and reinsert them in DOM to force them to be loaded.
- Replace request.json with request.get_json(force=True). This is necessary since Colab doesn't respect Content-Type header when forwarding requests through proxy, and force=True forces the request body to be parsed as JSON

Differential Revision: D19590439

### DIFF
--- a/captum/insights/api.py
+++ b/captum/insights/api.py
@@ -14,9 +14,6 @@ from typing import (
 )
 
 import torch
-from torch import Tensor
-from torch.nn import Module
-
 from captum.attr import IntegratedGradients
 from captum.attr._utils.batching import _batched_generator
 from captum.attr._utils.common import _run_forward, safe_div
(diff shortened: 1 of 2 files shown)


J0Nreynolds:
All failures are due to linter error:
./captum/insights/api.py:50:9: F401 'google.colab' imported but unused

However, we need to import google.colab in order to test for notebook context. All changes have been retested in Colab, and the widget functions properly.

Issue #253, opened 2020-01-27
GradientShap's `attribute` method `baselines` argument should be None


josejimenezluna:
https://github.com/pytorch/captum/blob/5231c2ec8c59f7ca48a7b421adb412ccf21c8364/captum/attr/_core/gradient_shap.py#L16-L35

According to the docs, the `baselines` parameter in the `attribute` method of `GradientShap` is optional, and is replaced with a zero-filled Tensor as the same size as the input if not provided. However at the moment it's a required argument.

NarineK:
Hi @josejimenezluna, thank you for finding the issue. It is a bug in the documentation. We decided to make baseline a mandatory field for GradientSHAP and let the user specify the distribution of baselines through that variable. If we default is to zero it might not be a very meaningful distribution.

NarineK:
Fixed with #256

PR #256, opened 2020-01-28
Fix documentation and input argument checks for gradientSHAP and …

NarineK:
1. Fixes the documentation issue mentioned in https://github.com/pytorch/captum/issues/253 and deepliftSHAP
2. Added additional checks in gradientSHAP to make sure that the input provided is a tensor corresponding to the distribution


captum/attr/_core/layer/layer_gradient_shap.py
```diff
@@ -96,40 +96,29 @@ def attribute(
                         that for all given input tensors, dimension 0 corresponds
                         to the number of examples, and if multiple input tensors
                         are provided, the examples must be aligned appropriately.
-            baselines (scalar, tensor, tuple of scalars or tensors, callable, optional):
+            baselines (tensor, tuple of tensors, callable):
                         Baselines define the starting point from which expectation
-                        is computed and can be provided as:
+                        is computed can be provided as:
```
vivekmig: nit: seems like leaving and is grammatically correct here? (also in GradientShap and NeuronGradientShap as well?). Or otherwise can make a separate sentence?

captum/attr/_core/neuron/neuron_gradient_shap.py
```diff
@@ -33,7 +33,7 @@ def attribute(
         self,
         inputs,
         neuron_index,
-        baselines=None,
+        baselines,
```
vivekmig: Can you merge with master and update the type hint here? Thanks!

captum/attr/_core/layer/layer_deep_lift.py
```diff
@@ -348,35 +348,31 @@ def attribute(
                         to the number of examples (aka batch size), and if
                         multiple input tensors are provided, the examples must
                         be aligned appropriately.
-            baselines (scalar, tensor, tuple of scalars or tensors, callable, optional):
+            baselines (tensor, tuple of tensors, callable):
```
vivekmig: Also, baselines still defaults to None here, do you want to remove that?

PR #259, opened 2020-01-30
Added type hints for captum integrated_gradients test cases

joshuadlouie:
A bit awkward in the helper functions that utilized attribute, as the return type from the attribute calls w/ return_convergence_delta was giving various type failures that I had to work around (inability to statically discern what type would be returned) and the fact that those helper functions always made the inputs/baselines into tuples to be used, resulting in additional helper variables being required.


tests/attr/test_integrated_gradients_basic.py
```diff
@@ -253,86 +264,101 @@ def _assert_batched_tensor_multi_input(
...
         self,
-        model,
-        inputs,
-        baselines=None,
-        target=None,
-        additional_forward_args=None,
-        type="vanilla",
-        approximation_method="gausslegendre",
-    ):
+        model: nn.Module,
```
vivekmig: nit: Can we add "from torch.nn import Module" and just use Module here? That's the way we've done it in other files, so it would be nice to remain consistent here.

tests/attr/test_integrated_gradients_basic.py
```diff
@@ -253,86 +264,101 @@ def _assert_batched_tensor_multi_input(
...
+        baselines_tuple: Tuple[Any, ...]
         if baselines is None:
-            baselines = _tensorize_baseline(inputs, _zeros(inputs))
+            baselines_tuple = _tensorize_baseline(inputs_tuple, _zeros(inputs_tuple))
+        elif not isinstance(baselines, tuple):
+            baselines_tuple = (baselines,)
+        else:
+            baselines_tuple = baselines
 
+        attributions: Tuple[Tensor, ...]
```
vivekmig: I think for this issue, it might be cleaner to avoid declaring the types in advance here and just use "cast" to cast the return value to the expected type. For instance, below, it would be something like:
`attributions, delta = cast(Tuple[Tuple[Tensor, ...], Tensor], ig.attribute ...`

You'll need to import cast from typing, it won't have any effect on runtime, just for type hinting.

tests/attr/test_integrated_gradients_basic.py
```diff
@@ -253,86 +264,101 @@ def _assert_batched_tensor_multi_input(
...
         r"""
             attrib_type: 'vanilla', 'smoothgrad', 'smoothgrad_sq', 'vargrad'
         """
         ig = IntegratedGradients(model)
-        if not isinstance(inputs, tuple):
-            inputs = (inputs,)
-
-        if baselines is not None and not isinstance(baselines, tuple):
-            baselines = (baselines,)
+        inputs_tuple: Tuple[Tensor, ...]
```
vivekmig: If we avoid the explicit type hints here, can we leave the original code? Or is there an error that requires these hints?

tests/attr/test_integrated_gradients_basic.py
```diff
@@ -17,91 +18,95 @@
...
-    def _assert_multi_variable(self, type, approximation_method="gausslegendre"):
-        model = BasicModel2()
+    def _assert_multi_variable(
+        self, type: str, approximation_method: str = "gausslegendre"
+    ) -> None:
+        model: nn.Module = BasicModel2()
 
-        input1 = torch.tensor([3.0])
-        input2 = torch.tensor([1.0], requires_grad=True)
+        input1: Tensor = torch.tensor([3.0])
```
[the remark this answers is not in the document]
joshuadlouie: I somehow screwed up what I checked in yesterday, I had removed all the local types when possible, but accidentally messed up and put in an old copy. Anyways, latest should be fixed.

tests/attr/test_integrated_gradients_basic.py
```diff
@@ -253,86 +264,101 @@ def _assert_batched_tensor_multi_input(
...
-        inputs,
-        baselines=None,
-        target=None,
-        additional_forward_args=None,
-        type="vanilla",
-        approximation_method="gausslegendre",
-    ):
+        model: nn.Module,
+        inputs: TensorOrTupleOfTensors,
+        baselines: Optional[Any] = None,
```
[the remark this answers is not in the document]
joshuadlouie: Done.

tests/attr/test_integrated_gradients_basic.py
joshuadlouie: Done.

tests/attr/test_integrated_gradients_basic.py
joshuadlouie: Casting was what I was missing to make life easier, when we "know" the right thing. I think with that, the latest changes are much more straight forward.

tests/attr/test_integrated_gradients_basic.py
joshuadlouie: Casting solved these annoyances, the changes are tighter now.

tests/attr/test_integrated_gradients_classification.py
```diff
@@ -83,26 +87,28 @@ def _assert_softmax_classification_batch(
...
         self,
-        model,
-        inputs,
-        target,
-        type="vanilla",
-        approximation_method="gausslegendre",
-        baseline=None,
-    ):
+        model: Module,
+        input: Tensor,
```
joshuadlouie: From the way this test is written, the only way it was able to work was if inputs was a singular Tensor, and not a TensorOrTupleOfTensors. So I changed the name of it and the type to reflect that.

tests/attr/test_integrated_gradients_basic.py
```diff
@@ -253,20 +266,22 @@ def _assert_batched_tensor_multi_input(
...
+        type: Optional[str] = "vanilla",
+        approximation_method: str = "gausslegendre",
+    ) -> Tuple[Tensor, ...]:
         r"""
             attrib_type: 'vanilla', 'smoothgrad', 'smoothgrad_sq', 'vargrad'
         """
         ig = IntegratedGradients(model)
         if not isinstance(inputs, tuple):
-            inputs = (inputs,)
+            inputs = cast(TensorOrTupleOfTensors, (inputs,))
```
vivekmig: This changes behavior, since cast doesn't have any effect at runtime. Can we just leave the original here?

vivekmig: Never mind, sorry, ignore above, this is fine!

PR #260, opened 2020-01-31
Add mypy type hints for captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py

fuquan-wang:
This is to add type hints in the file mentioned in the summary

The following tests have been run:
```
python -m unittest -v tests.attr.test_guided_backprop 
python -m unittest -v tests.attr.test_data_parallel
./scripts/run_mypy.sh
```

### DIFF
--- a/captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
+++ b/captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
@@ -1,11 +1,22 @@
 #!/usr/bin/env python3
+from typing import Any
+from typing import List
+from typing import Optional
+from typing import Tuple
+from typing import Union
+
+import torch.nn as nn
+from torch import Tensor
+
 from ..._utils.attribution import NeuronAttribution, GradientAttribution
 from ..._utils.gradient import construct_neuron_grad_fn
 from ..guided_backprop_deconvnet import GuidedBackprop, Deconvolution
 
 
 class NeuronDeconvolution(NeuronAttribution, GradientAttribution):
-    def __init__(self, model, layer, device_ids=None):
+    def __init__(
+        self, model: nn.Module, layer: nn.Module, device_ids: Optional[List] = None
+    ) -> None:
         r"""
         Args:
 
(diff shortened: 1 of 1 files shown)


captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
```diff
@@ -1,11 +1,22 @@
...
+
 from ..._utils.attribution import NeuronAttribution, GradientAttribution
 from ..._utils.gradient import construct_neuron_grad_fn
 from ..guided_backprop_deconvnet import GuidedBackprop, Deconvolution
 
 
 class NeuronDeconvolution(NeuronAttribution, GradientAttribution):
-    def __init__(self, model, layer, device_ids=None):
+    def __init__(
+        self, model: nn.Module, layer: nn.Module, device_ids: Optional[List] = None
```
vivekmig: nit: We usually have been using the type as just Module and importing as "from torch.nn import Module", can you switch to that for consistency? Thanks!

captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
```diff
@@ -31,11 +42,11 @@ def __init__(self, model, layer, device_ids=None):
 
     def attribute(
         self,
-        inputs,
-        neuron_index,
-        additional_forward_args=None,
-        attribute_to_neuron_input=False,
-    ):
+        inputs: Union[Tensor, Tuple[Tensor]],
```
vivekmig: You can actually use the TensorOrTupleOfTensors type here, and then the return type would be the same (this conveys that the return type matches the inputs type). That type can be imported from captum.attr.utils.typing.

captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
```diff
@@ -31,11 +42,11 @@ def __init__(self, model, layer, device_ids=None):
...
+        neuron_index: Union[int, Tuple[int]],
+        additional_forward_args: Optional[Any] = None,
```
vivekmig: Any should be sufficient, and include None, so Optional[Any] isn't necessary.

captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
```diff
@@ -130,7 +141,9 @@ def attribute(
 
 
 class NeuronGuidedBackprop(NeuronAttribution, GradientAttribution):
```
vivekmig: Same comments from above for this class.

captum/attr/_core/neuron/neuron_guided_backprop_deconvnet.py
```diff
@@ -31,11 +42,11 @@ def __init__(self, model, layer, device_ids=None):
 
     def attribute(
         self,
-        inputs,
-        neuron_index,
-        additional_forward_args=None,
-        attribute_to_neuron_input=False,
-    ):
+        inputs: Union[Tensor, Tuple[Tensor]],
+        neuron_index: Union[int, Tuple[int]],
```
vivekmig: This should be Tuple[int, ...], since it should take an arbitrary length tuple, not just a Tuple containing a single integer.

PR #261, opened 2020-01-31
Add type hint for occlusion.py and test_occlusion.py

chenyc15:
Add type hints for the following two files:
- captum/attr/_core/occlusion.py
- captum/tests/attr/test_occlusion.py

Test: 
`python -m unittest -v tests.attr.test_occlusion` OK
`./scripts/run_mypy.sh` shows no errors:
```bash
Success: no issues found in 44 source files
Success: no issues found in 48 source files
```


vivekmig:
Hi @chenyc15, thanks for working on this! I took a quick look, here are some suggestions:

1. For the error in line 29, you can just add the comment "# type: ignore" to the end of the line, and it'll suppress the warning. This error is caused by the child class not having a matching signature to the parent, and we may look into changing the class hierarchy, but for now, you can just add the ignore flag. Add a short comment above explaining the reason.
2. In line 297, change the 1 to "torch.ones(1, dtype=torch.long)", this should fix the next 2 errors. It seems that some of the type promotion possibilities may not have been type hinted in PyTorch.
3. In line 348, this seems to be caused by a mismatch with PyTorch type hints, which specify a List[int] for the parameter, while the documentation requests a tuple here: https://pytorch.org/docs/stable/nn.functional.html. This may likely require an upstream fix for the hint in PyTorch, so for now, just add a "type: ignore" comment with a short comment above to this line as well.

It seems that there are also some issues with the test files, which show up after these checks pass. Some seem to be due to incorrect hints in the test file (such as ablations_per_eval, which takes a tuple). See if you can resolve those by fixing the hints, if you have further issues feel free to update with the errors and I can take a look.

captum/attr/_core/occlusion.py
```diff
@@ -23,16 +26,22 @@ def __init__(self, forward_func):
...
+            Union[int, Tuple[int, ...], Tuple[Union[int, Tuple[int, ...]], ...]]
+        ] = None,
+        baselines: Optional[
+            Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
+        ] = None,
+        target: Optional[
+            Union[int, Tuple[int, ...], Tensor, List[Tuple[int, ...]]]
+        ] = None,
+        additional_forward_args: Any = None,
+        ablations_per_eval: Optional[int] = 1,
```
vivekmig: ablations_per_eval should be just int, since it cannot be None

captum/attr/_core/occlusion.py
```diff
@@ -245,8 +254,16 @@ def attribute(
         )
 
     def _construct_ablated_input(
-        self, expanded_input, input_mask, baseline, start_feature, end_feature, **kwargs
-    ):
+        self,
+        expanded_input: Tensor,
+        input_mask: Tensor,
```
vivekmig: input_mask should be an Optional[Tensor], since None is passed here.

captum/attr/_core/occlusion.py
```diff
@@ -330,9 +351,13 @@ def _occlusion_mask(
...
-        padded_tensor = torch.nn.functional.pad(sliding_window_tsr, tuple(pad_values))
+        padded_tensor = torch.nn.functional.pad(
+            sliding_window_tsr, tuple(pad_values)  # type: ignore
+        )
         return padded_tensor.reshape((1,) + padded_tensor.shape)
 
-    def _get_feature_range_and_mask(self, input, input_mask, **kwargs):
+    def _get_feature_range_and_mask(
+        self, input: Tensor, input_mask: Tensor, **kwargs: Any
+    ) -> Tuple[int, float, None]:
```
vivekmig: The second return type here should also be an int. You might need to use cast below if you have errors.

tests/attr/test_occlusion.py
```diff
@@ -265,16 +268,26 @@ def test_simple_multi_input_conv(self):
...
-        strides=None,
-    ):
+        model: Callable,
+        test_input: TensorOrTupleOfTensors,
+        expected_ablation: Union[
+            float,
+            List[float],
+            List[List[float]],
+            Tuple[Union[Tensor, List[float]], ...],
+            Tuple[List[List[float]], ...],
```
vivekmig: nit: Can you just add List[List[float]] to the above union?

captum/attr/_core/occlusion.py
```diff
@@ -245,8 +254,16 @@ def attribute(
         )
 
     def _construct_ablated_input(
-        self, expanded_input, input_mask, baseline, start_feature, end_feature, **kwargs
-    ):
+        self,
+        expanded_input: Tensor,
+        input_mask: Tensor,
+        baseline: Optional[
+            Union[Tensor, int, float, Tuple[Union[Tensor, int, float], ...]]
```
vivekmig: baseline here should just be Union[Tensor, int, float] (not optional), since we only apply to a single tensor.

tests/attr/test_occlusion.py
```diff
@@ -15,9 +18,9 @@
 
 
 class Test(BaseTest):
-    def test_improper_window_shape(self):
+    def test_improper_window_shape(self) -> None:
         net = BasicModel_ConvNet_One_Conv()
-        inp = torch.arange(16).view(1, 1, 4, 4).type(torch.FloatTensor)
+        inp = torch.arange(16).view(1, 1, 4, 4).to(torch.float)
```
vivekmig: nit: Can you actually update this (and below) to `torch.arange(16, dtype=torch.float).view(1, 1, 4, 4)`? That would be a little more concise.

PR #264, opened 2020-01-31
Type hint on input x gradient

luciawlli:
$ ./scripts/run_mypy.sh
360 passed, 50 skipped, 138 warnings in 106.13 seconds

$ black .
All done! ✨ 🍰 ✨
107 files left unchanged.

$ flake8 .
<no errors>

$ ./scripts/run_mypy.sh
Success: no issues found in 44 source files
Success: no issues found in 48 source files

### DIFF
--- a/captum/attr/_core/input_x_gradient.py
+++ b/captum/attr/_core/input_x_gradient.py
@@ -1,11 +1,14 @@
 #!/usr/bin/env python3
+from typing import Callable, List, Optional, Tuple, Union, Any
+from torch import Tensor
+
 from .._utils.common import _format_input, _format_attributions
 from .._utils.attribution import GradientAttribution
 from .._utils.gradient import apply_gradient_requirements, undo_gradient_requirements
 
 
 class InputXGradient(GradientAttribution):
-    def __init__(self, forward_func):
+    def __init__(self, forward_func: Callable) -> None:
         r"""
         Args:
 
@@ -14,7 +17,14 @@ class InputXGradient(GradientAttribution):
         """
         GradientAttribution.__init__(self, forward_func)
 
-    def attribute(self, inputs, target=None, additional_forward_args=None):
+    def attribute(
+        self,
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
+        target: Optional[
+            Union[int, Tuple[int, ...], Tensor, List[Tuple[int, ...]]]
+        ] = None,
+        additional_forward_args: Any = None,
+    ) -> Union[Tensor, Tuple[Tensor, ...]]:
         r""""
         A baseline approach for computing the attribution. It multiplies input with
         the gradient with respect to input.
--- a/captum/attr/_core/input_x_gradient.py
+++ b/captum/attr/_core/input_x_gradient.py
@@ -5,6 +5,7 @@ from torch import Tensor
 from .._utils.common import _format_input, _format_attributions
 from .._utils.attribution import GradientAttribution
 from .._utils.gradient import apply_gradient_requirements, undo_gradient_requirements
+from .._utils.typing import TensorOrTupleOfTensors
 
 
 class InputXGradient(GradientAttribution):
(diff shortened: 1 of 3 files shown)


tests/attr/test_input_x_gradient.py
```diff
@@ -96,14 +101,16 @@ def _input_x_gradient_classification_assert(self, nt_type="vanilla"):
             output = model(input)[:, target]
             output.backward()
             expercted = input.grad * input
-            self.assertEqual(
-                expercted.detach().numpy().tolist(),
-                attributions.detach().numpy().tolist(),
-            )
+            if isinstance(attributions, Tensor):
```
luciawlli: added if check here and below, otherwise ./scripts/run_mypy.sh gives errors on 
```

tests/attr/test_input_x_gradient.py:87: error: Item "Tuple[Tensor, ...]" of "Union[Tensor, Tuple[Tensor, ...]]" has no attribute "shape"
tests/attr/test_input_x_gradient.py:107: error: Item "Tuple[Tensor, ...]" of "Union[Tensor, Tuple[Tensor, ...]]" has no attribute "detach"
```

due to `inputs: Union[Tensor, Tuple[Tensor, ...]]`

vivekmig: The change above to TensorOrTupleOfTensors should avoid this issue, and you can remove this and the extra case below.

captum/attr/_core/input_x_gradient.py
```diff
@@ -14,7 +17,14 @@ def __init__(self, forward_func):
         """
         GradientAttribution.__init__(self, forward_func)
 
-    def attribute(self, inputs, target=None, additional_forward_args=None):
+    def attribute(
+        self,
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
```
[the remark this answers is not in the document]
luciawlli: Thanks for the review.

After this change, I also made the change in `test_input_x_gradient.py` line 52 to make `input TensorOrTupleOfTensors`, however, the`if isinstance ...` is still needed, else would run into the error:

`tests/attr/test_input_x_gradient.py:88: error: "Tuple[Tensor, ...]" has no attribute "shape"`

tests/attr/test_input_x_gradient.py
```diff
@@ -75,15 +78,18 @@ def _input_x_gradient_base_assert(
                     )
                 self.assertEqual(input.shape, attribution.shape)
         else:
```
vivekmig: Can you change this else to `elif isinstance(attributions, Tensor) `?
This captures the only other possibility for attributions and would not need any isinstance checks below.

tests/attr/test_input_x_gradient.py
```diff
@@ -106,4 +112,4 @@ def _input_x_gradient_classification_assert(self, nt_type="vanilla"):
                 input, nt_type=nt_type, n_samples=10, stdevs=1.0, target=target
             )
 
-        self.assertAlmostEqual(attributions.shape, input.shape)
+        assertArraysAlmostEqual(attributions.shape, input.shape)
```
vivekmig: Can you change this to just self.assertEqual? Thanks!

tests/attr/test_input_x_gradient.py
```diff
@@ -1,55 +1,58 @@
...
         self,
-        model,
-        inputs,
-        expected_grads,
-        additional_forward_args=None,
-        nt_type="vanilla",
-    ):
+        model: Module,
+        inputs: TensorOrTupleOfTensors,
+        expected_grads: Tensor,
```
vivekmig: This should also be TensorOrTupleOfTensors

luciawlli: Changing `expected_grads` from `Tensor` to `TensorOrTupleOfTensors` results in the following error in 3 places:
`tests/attr/test_input_x_gradient.py:35: error: Value of type variable "TensorOrTupleOfTensors" of "_input_x_gradient_base_assert" of "Test" cannot be "object"`

luciawlli:
> Looks great, thanks again! Just a few more small changes, and it should be good.

Thanks for reviewing! Addressed all the comments :)

tests/attr/test_input_x_gradient.py
```diff
@@ -74,16 +77,18 @@ def _input_x_gradient_base_assert(
                         attribution.reshape(-1), (expected_grad * input).reshape(-1)
                     )
                 self.assertEqual(input.shape, attribution.shape)
-        else:
-            if nt_type == "vanilla":
+        elif isinstance(attributions, Tensor):
+            if nt_type == "vanilla" and isinstance(attributions, Tensor):
```
vivekmig: I think this isinstance can also be removed now?

luciawlli:
> Thanks for updating! Just one more minor fix.

Yes, it should be cleaned up. Sorry missed this early.

PR #268, opened 2020-02-03
Add type hints for layer gradient SHAP

NarineK:
Adding type hints in layer gradient SHAP and related tests

### DIFF
--- a/captum/attr/_core/layer/layer_gradient_shap.py
+++ b/captum/attr/_core/layer/layer_gradient_shap.py
@@ -18,7 +18,7 @@ from ..._utils.common import (
     _format_callable_baseline,
     _compute_conv_delta_and_format_attrs,
 )
-from .._utils.typing import TensorOrTupleOfTensors
+from ..._utils.typing import TensorOrTupleOfTensors
 
 from ..noise_tunnel import NoiseTunnel
 
--- a/captum/attr/_core/layer/layer_gradient_shap.py
+++ b/captum/attr/_core/layer/layer_gradient_shap.py
@@ -60,7 +60,7 @@ class LayerGradientShap(LayerAttribution, GradientShap):
         target: Optional[
             Union[int, Tuple[int, ...], Tensor, List[Tuple[int, ...]]]
         ] = None,
-        additional_forward_args: Any = None,
+        additional_forward_args: Optional[Any] = None,
         attribute_to_layer_input: Optional[bool] = False,
     ) -> TensorOrTupleOfTensors:
         ...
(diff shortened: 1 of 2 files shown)


captum/attr/_core/layer/layer_gradient_shap.py
```diff
@@ -39,6 +50,37 @@ def __init__(self, forward_func, layer, device_ids=None):
         LayerAttribution.__init__(self, forward_func, layer, device_ids)
         GradientShap.__init__(self, forward_func)
 
+    @typing.overload
+    def attribute(
+        self,
+        inputs: TensorOrTupleOfTensors,
+        baselines: TensorOrTupleOfTensors,
```
NarineK: need to add callable here

tests/attr/layer/test_layer_gradient_shap.py
```diff
@@ -112,33 +117,33 @@ def test_basic_multi_input(self):
...
-            self, attrs, expected, delta=0.005,
-        )
+        assertTensorTuplesAlmostEqual(self, attrs, expected, delta=0.005)
         if expected_delta is None:
             _assert_attribution_delta(self, inputs, attrs, n_samples, delta, True)
         else:
-            for delta_i, expected_delta_i in zip(delta, expected_delta):
-                self.assertAlmostEqual(delta_i, expected_delta_i, delta=0.01)
+            for delta_i, expected_delta_i in zip(
+                torch.unbind(delta), torch.unbind(expected_delta)
```
vivekmig: Do we need the unbind calls here? I think this should type hint fine without it if testing with torch 1.4.

tests/attr/layer/test_layer_gradient_shap.py
```diff
@@ -112,33 +117,33 @@ def test_basic_multi_input(self):
...
-        n_samples=5,
-        attribute_to_layer_input=False,
-        add_args=None,
-    ):
+        model: Module,
+        layer: Module,
+        inputs: TensorOrTupleOfTensors,
+        baselines: Union[TensorOrTupleOfTensors, Callable],
+        target: Union[int, Tuple[int, ...], Tensor, List[Tuple[int, ...]]],
+        expected: Union[Tensor, Tuple[Tensor, ...], List, Tuple[List, ...]],
```
vivekmig: nit: Could make the List hint more specific? (List[float] or List[List[float]]?)

tests/attr/layer/test_layer_gradient_shap.py
NarineK: Cool, yeah, it works w/1.4 without unbinding.

tests/attr/layer/test_layer_gradient_shap.py
NarineK: It makes it a little more lengthy but added it

PR #270, opened 2020-02-03
Adding Type Hints to Captum: noise_tunnel.py

Vaishali-Garg:
Summary:Added type hints to the attribute method and added required import

Test Plan:Ran run_mypy.sh and there are no errors

Reviewers:vivekm

Subscribers:jomart

Tasks:T61337176

Tags:

### DIFF
--- a/captum/attr/_core/noise_tunnel.py
+++ b/captum/attr/_core/noise_tunnel.py
@@ -51,7 +51,7 @@ class NoiseTunnel(Attribution):
         stdevs: Union[float, Tuple[float, ...]] = 1.0,
         draw_baseline_from_distrib: bool = False,
         **kwargs: Any
-    ):
+    ) -> Union[Tensor, Tuple[Tensor, ...]]:
         r"""
         Adds gaussian noise to each input in the batch `n_samples` times
         and applies the given attribution algorithm to each of the samples.
@@ -167,7 +167,10 @@ class NoiseTunnel(Attribution):
                 for (input, stdev) in zip(inputs, stdevs_)
             )
 
-        def add_noise_to_input(input: Tensor, stdev: float) -> Tensor:
+        def add_noise_to_input(
+            input: Tensor,
+            stdev: float
+            ) -> Tensor:
             # batch size
             bsz = input.shape[0]
 
@@ -268,7 +271,7 @@ class NoiseTunnel(Attribution):
 
         _validate_noise_tunnel_type(nt_type, SUPPORTED_NOISE_TUNNEL_TYPES)
 
-        delta = None
+        delta = 0
         inputs_with_noise = add_noise_to_inputs()
         # if the algorithm supports targets, baselines and/or additional_forward_args
         # they will be expanded based on the n_steps and corresponding kwargs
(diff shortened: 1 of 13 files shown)


captum/attr/_core/noise_tunnel.py
```diff
@@ -1,11 +1,14 @@
...
+from typing import Callable, List, Optional, Tuple, Union, Any
 
 import torch
+from torch import Tensor
 
 import numpy as np
 from enum import Enum
 
 from .._utils.attribution import Attribution
+
```
vivekmig: nit: New line not necessary here?

captum/attr/_core/noise_tunnel.py
```diff
@@ -43,13 +46,13 @@ def __init__(self, attribution_method):
...
-        stdevs=1.0,
-        draw_baseline_from_distrib=False,
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
+        nt_type: str = "smoothgrad",
+        n_samples: int = 5,
+        stdevs: Union[float, Tuple[float, ...]] = 1.0,
+        draw_baseline_from_distrib: bool = False,
         **kwargs
-    ):
+    ) -> Union[Tensor, Tuple[Tensor]]:
```
vivekmig: This should include Tuple[Tensor, ...] not just Tuple[Tensor] to account for tuples with multiple tensors.

captum/attr/_core/noise_tunnel.py
```diff
@@ -43,13 +46,13 @@ def __init__(self, attribution_method):
...
-        n_samples=5,
-        stdevs=1.0,
-        draw_baseline_from_distrib=False,
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
+        nt_type: str = "smoothgrad",
+        n_samples: int = 5,
+        stdevs: Union[float, Tuple[float, ...]] = 1.0,
+        draw_baseline_from_distrib: bool = False,
         **kwargs
-    ):
```
vivekmig: Can also explicitly type hint **kwargs with Any

vivekmig:
It also looks like there's some lint errors from flake8, just run black . and flake8 . before pushing updated changes. Thanks!

captum/attr/_core/noise_tunnel.py
```diff
@@ -43,13 +45,13 @@ def __init__(self, attribution_method):
...
-        draw_baseline_from_distrib=False,
-        **kwargs
-    ):
+        inputs: Union[Tensor, Tuple[Tensor, ...]],
+        nt_type: str = "smoothgrad",
+        n_samples: int = 5,
+        stdevs: Union[float, Tuple[float, ...]] = 1.0,
+        draw_baseline_from_distrib: bool = False,
+        **kwargs: Any
+    ) -> Union[Tensor, Tuple[Tensor, ...]]:
```
vivekmig: I realized this return type will have some issues with other methods, so we will look into those later, you can just remove the return type hint here.

captum/attr/_core/noise_tunnel.py
```diff
@@ -326,7 +331,11 @@ def compute_expected_attribution_and_sq(attribution):
         )
 
     def _apply_checks_and_return_attributions(
-        self, attributions, is_attrib_tuple, return_convergence_delta, delta
+        self,
+        attributions: Union[Tensor, Tuple[Tensor, ...]],
+        is_attrib_tuple: bool,
+        return_convergence_delta: bool,
+        delta: Optional[float],
```
vivekmig: This should actually be Optional[Tensor]. You will need to change line 274 above which initializes delta, to `delta = None`.

PR #280, opened 2020-02-09
Add tutorial on regression models

NarineK:
Adding a tutorial on a regression model using Boston House Prices.


NarineK:
Thank you for reviewing the PR Vivek! 

I originally did some research and gathered information about the models that people have built for Boston House Prices dataset. Based on that and my experiments I observed that the performance of the model is noticeably better when we standardize the data. This, however, as you mentioned, can lead to the fact that the sum of those features will get close to 0 and as a result if we were to use it for the entire dataset it will lead to 0 attributions. In a general case, I don't think that this is principally wrong -it reflects the vulnerability of the attribution algorithms to the model inputs and baselines. But I agree that it would be better for the purpose of attributions not to normalize the data.

In terms of choosing different baselines, I think that if in a tutorial we end up choosing the baseline in a very specific way the users might copy that pattern for their models and datasets which might not be ideal. Perhaps comparing different baselines is a better approach here.

To make the tutorial less trivial, I'll use a more complex, non-linear model for attribution purposes and see how it turns out.

vivekmig:
Sounds great, thanks :) ! Also another option without changing baselines would be to just average absolute value of attributions (which would be comparable to absolute value of weights); this would be accurate for feature importance, and direction can be separately investigated by looking at the attribution sign for positive and negative feature values separately.

tutorials/House_Prices_Regression_Interpret.ipynb
```diff
@@ -0,0 +1,657 @@
...
...Kd3JbyfZVtbz2vQe2JozcsuE3NNlDutb9DrTE524rW5nEkckxlveDQnf1VshodeyxjzruTPb59Pf/vsyfwuZc7OP3RW/nnH8/gy/Nnu3SNHhSPiohIWOk7KjzyimszxJnzX3m3YvtB5zuxM9Ptf/nl3XnZI37NZ0+zYuLXTPFxoiZzZgiI0CTPjTaCt95y++x+//v95ffd536/8AKsXJnxX6fPWkzPoFG9nj7L9FmLsz5kKSb7xrU9CvK4lHpRhghzGpCyzrCfNw++8hU3A/bkk91PV5ebSVAltIJB4qLcy7Tb5izJGAwpLZ6UStjTSUo0lft9lZr9ly3tSqZUjZNnzM3Ypp5z88IB/+vFK/3jpG024OEX3s1Y7kuOPge197m9O2Ik//fpY/nTxBZOeuJOvrTkYS7+x+XQcLm7wcKFsOuuA/6nmuLRuKbLERGJq2r6jgqz9o5OPlq1Zkh5prjW69z9+nnL1o6f5ZMqsFD5puPP5/ZeMa/X/RaivaMz46AQBhrqE3R1D80YUGMMY6fOjk6MU1sLP/uZ+3nuORg92pVvt13/bR58ED7zmbVXMx13tvJ0hWzRkC6u7VGQx6UVZZJRWGfml32G/Q03wPPP919vaCjN/UaEVjBIHFRimXZcAxIJjzimk5Tglft9VejsP6+2s9faotrvl9/JfL9e5ena5iyhp9d7WZuf/dOkX/ewdfj9pKM4/Ku/4DuHntv/h912g6OOgttuc3tFACPrM6cE8ir3a91hmTeh9yovt7imyxERiTP1mQQv9f05eCBigxGJjHGtV5w5OMqL8kocP7EtFB6/ej3nAD29FmOgNkMmhl5rc8Y4YYvP1tphB5dxDODctNj1s5912cnSV54FJK7tUZDHpYEyiZSKpIXcbjt3oj51Ktx5pyubP98tw425MKfdFPGrEsu0wxSQZEtzJtEW1kkrEm3lfF95bTqebTNyyN52FtN+FzOpIdttFBv5U5cpdaUx3LXLvi7Wfu45OO88+Mc/3GDZppvCtdcyrHd1xvvLI6tQRonazKe+XuXlFtd0OSIicZbvfldSepm+PwFGDKsreN/ZlFwxa1j5iW2LiV+9nvOU91b20JtjUzGvGCds8VlGbW0udv3XvwaWAZsl+tjmndeG/IufveCK7cuJax+uV4rPShxXiN51IrlVdIb9JZfAwQfDqlWw556w2WbuZD
```
NarineK: This should be neuron

PR #281, opened 2020-02-11
Enable mypy for Insights and fix type errors

edward-io:


captum/insights/api.py
```diff
@@ -339,20 +346,21 @@ def _calculate_vis_output(
...
                 score=100, index=label[0], label=self.classes[label[0]]
             )
-        else:
-            actual_label_output = None
 
         predicted_scores = self._get_labels_from_scores(scores, predicted)
 
         # Filter based on UI configuration
-        if not self._should_keep_prediction(predicted_scores, actual_label_output):
+        if actual_label_output is None or not self._should_keep_prediction(
```
vivekmig: nit: Is this preferable over keeping this line the same and just changing the line in the function to return False if actual_label_output is None?

captum/insights/features.py
```diff
@@ -60,8 +61,8 @@ def __init__(
                             specified in ``captum/captum/insights/frontend/App.js``.
         """
         self.name = name
-        self.baseline_transforms = baseline_transforms
-        self.input_transforms = input_transforms
+        self.baseline_transforms = format_transforms(baseline_transforms)
+        self.input_transforms = format_transforms(input_transforms)
```
vivekmig: Was checking the usage of these transforms and noticed there are some checks in api.py verifying these (baseline_transforms and input_transforms) are not None, those shouldn't be necessary after this change since the format call will only return a List?

captum/insights/api.py
```diff
@@ -164,7 +166,7 @@ def _calculate_attribution_from_cache(
     def _calculate_attribution(
         self,
         net: Module,
-        baselines: Optional[List[Tuple[Tensor, ...]]],
+        baselines: Optional[List[Tuple[Optional[Tensor], ...]]],
```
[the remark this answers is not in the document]
edward-io: This is a better approach, thanks.

captum/insights/api.py
edward-io: Typechecker isn't happy later on when returning the `VisualizationOutput`

PR #290, opened 2020-02-15
Fixing LayerGradientShap Type Hints

vivekmig:
Fixes for LayerGradientShap type hints which is causing mypy to fail on master. LayerGradientShap previously extended from GradientShap, which has an inconsistent signature; this adjusts the inheritance hierarchy for layer methods to inherit directly from GradientAttribution and just access the necessary methods.

### DIFF
--- a/tests/attr/neuron/test_neuron_conductance.py
+++ b/tests/attr/neuron/test_neuron_conductance.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import unittest
-from typing import List, Optional, Tuple, Union, Any
+from typing import List, Optional, Tuple, Union, Any, cast
 import torch
 from torch import Tensor
 from torch.nn import Module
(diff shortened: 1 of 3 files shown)


tests/attr/neuron/test_neuron_conductance.py
```diff
@@ -155,12 +155,15 @@ def _conductance_input_sum_test_assert(
         ] = None,
     ):
         layer_cond = LayerConductance(model, target_layer)
-        attributions = layer_cond.attribute(
-            test_input,
-            baselines=test_baseline,
-            target=0,
-            n_steps=500,
-            method="gausslegendre",
+        attributions = cast(
```
NarineK: Is the cast necessary here ? I thought that the overrides for the attribute methods will solve the casting issue ?

vivekmig: The issue is that this is for a layer method. The overrides help for regular and neuron attribution, but for layer, whether it returns a tuple is based on the layer input / output and not necessarily dependent on the input, thus it can't be represented clearly in type hints.

NarineK: I see, makes sense!

PR #297, opened 2020-02-22
Shapley Values

vivekmig:
This PR adds a variant of Shapley Value sampling that iterates over all permutations to compute exact Shapley Values.

### DIFF
--- a/captum/attr/_core/shapley_value.py
+++ b/captum/attr/_core/shapley_value.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 
 import warnings
-from typing import Any, Callable, List, Optional, Tuple, Union, Iterable
+from typing import Any, Callable, List, Optional, Tuple, Union, Iterable, Sequence
 
+import itertools
 import torch
 from torch import Tensor
 
@@ -19,6 +20,11 @@ from .._utils.common import (
 )
 from .._utils.attribution import PerturbationAttribution
 from .._utils.typing import TensorOrTupleOfTensors
+
+
+def _all_perm_generator(num_features: int, num_samples: int) -> Iterable[Sequence[int]]:
+    for perm in itertools.permutations(range(num_features)):
+        yield perm
 
 
 def _perm_generator(num_features: int, num_samples: int) -> Iterable[Tensor]:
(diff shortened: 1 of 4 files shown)


captum/attr/_core/shapley_value.py
```diff
@@ -456,3 +472,49 @@ def construct_feature_mask(
...
+    More details regarding Shapley Values can be found in these papers:
+    https://www.sciencedirect.com/science/article/pii/S0305054808000804
+    https://pdfs.semanticscholar.org/7715/bb1070691455d1fcfc6346ff458dbca77b2c.pdf
+
+    NOTE: The method implemented here is very computationally intensive, and
+    should only be used with a very small number of features (e.g. < 7).
+    This implementation simply extends ShapleyValueSampling and
+    evaluates all permutations, leading to a total of n * n! evaluations for n
+    features. Shapley values can alternatively be computed with only 2^n
+    evaluations, and we plan to add this approach in the future.
```
NarineK: It would be good to check the size of the input tensor and warn the user using it or suggest to use the sampling one ... 

This will require overriding the attribute method and making a simple check ....

tests/attr/test_shapley.py
```diff
@@ -279,6 +285,19 @@ def _shapley_test_assert(
...
+                attributions = shapley_val.attribute(
+                    test_input,
+                    target=target,
+                    feature_mask=feature_mask,
+                    additional_forward_args=additional_input,
+                    baselines=baselines,
+                    perturbations_per_eval=batch_size,
+                )
+                assertTensorTuplesAlmostEqual(
+                    self, attributions, expected_attr, mode="max", delta=0.001
```
NarineK: could we use the same delta as above ?

vivekmig: We could, but that delta is almost always substantially larger since sampling has pretty wide errors, this is much closer, so I fixed a smaller value here.

captum/attr/_core/shapley_value.py
vivekmig: Good idea, adding!

vivekmig:
Thanks for reviewing! Updated to address comments.

PR #299, opened 2020-02-24
Add algorithm comparison matrix

NarineK:
This PR adds attribution algorithms comparison matrix.
It also renames the algorithms' name on the sidebar from `Algorithms` to `Algorithm Descriptions`


NarineK:
Thank you for the review comments, @vivekmig !

4. `Layer Internal Influence` sounds better, I'll rename to that name.
6. I think those columns are pretty close to each other, also OOMs were an important issue that we saw - I think that it is fine for now.
8. That's a constant but sure

The rest I fixed, thanks!

PR #300, opened 2020-02-24
Contributing Updates

vivekmig:
This updates CONTRIBUTING.md and setup.py with the additional requirement of isort as well as relevant instructions and requirements for type hinting.


NarineK:
Do we also need to mention about the extension package for the `Literals` ?

vivekmig:
Good point, will add that, thanks!

PR #310, opened 2020-03-03
Adding isort

vivekmig:
This applies isort to alphabetize import ordering. We also add an isort check to CircleCI to ensure that import ordering is maintained in future PRs. This required adding an isort config that is compatible with black, which was found here: https://github.com/psf/black

This also adds a type: ignore for an issue caused by merging Insights type fixes with type hints for common.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ pip using `pip install isort`, and run locally by calling
 ```bash
 isort
 ```
-from the repository root. No additional configuration should be needed.
+from the repository root. Configuration for isort is located in .isort.cfg.
 
 We feel strongly that having a consistent code style is extremely important, so
 CircleCI will fail on your PR if it does not adhere to the black or flake8 formatting style or isort import ordering.
--- a/captum/insights/api.py
+++ b/captum/insights/api.py
@@ -504,15 +504,19 @@ class AttributionVisualizer(object):
 
         # Type ignore for issue with passing union to function taking generic
         # https://github.com/python/mypy/issues/1533
-        for inputs, additional_args, label in _batched_generator(  # type: ignore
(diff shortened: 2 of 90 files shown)


captum/insights/api.py
```diff
@@ -502,15 +502,17 @@ def _get_outputs(self) -> List[Tuple[VisualizationOutput, SampleCache]]:
         batch_data = next(self._dataset_iter)
         vis_outputs = []
 
-        for inputs, additional_forward_args, label in _batched_generator(
+        # Type ignore for issue with passing union to function taking generic
+        # https://github.com/python/mypy/issues/1533
+        for inputs, additional_args, label in _batched_generator(  # type: ignore
```
vivekmig: These changes were necessary to fix mypy failures on master due to merging with the common type hints. additional_forward_args was renamed to additional_args since the "type: ignore" made the line too long causing flake8 errors.
CC: @edward-io

edward-io: could we do:

```
for (
  inputs,
  additional_forward_args,
  label,
) in _batched_generator( # type: ignore
```

instead?

CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ pip using `pip install isort`, and run locally by calling
 ```bash
 isort
 ```
-from the repository root. No additional configuration should be needed.
+from the repository root. Configuration for isort is located in .isort.cfg.
```
edward-io: Where is this file?

vivekmig: It's in root .isort.cfg.

captum/insights/api.py
vivekmig: Sure, I figured a single line is more concise, but this works too, will update, thanks!

PR #312, opened 2020-03-04
Use relative paths for Insights

edward-io:
This change allows Insights to work with [Jupyter server proxy](https://github.com/jupyterhub/jupyter-server-proxy).

However, Sagemaker is broken due to https://forums.aws.amazon.com/message.jspa?messageID=914445#914445.

### DIFF
--- a/captum/insights/frontend/package.json
+++ b/captum/insights/frontend/package.json
@@ -2,6 +2,7 @@
   "name": "frontend",
   "version": "0.2.0",
   "private": true,
+  "homepage": ".",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


captum/insights/frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "frontend",
   "version": "0.2.0",
   "private": true,
+  "homepage": ".",
```
J0Nreynolds: Was this necessary to get relative paths working?

edward-io: Yup, this rewrites the urls from `/something.css` to `./something.css`.

Issue #316, opened 2020-03-07
Bug in *.attribute -- New tensor in gathering ignoring used device


kai-tub:
Hi, first of all, thanks for providing and working on such a neat library!
I think I found a "bug" in the [`common.py`](https://github.com/pytorch/captum/blob/master/captum/attr/_utils/common.py#L447) file, which is used for most attributions. The problem arises when a Cuda device is used and a non-tensor target is given, for example, a simple list. 

I don't know if you officially support CUDA or not, but as I couldn't find any hints indicating otherwise, I've used the Saliency function without any problems on a CUDA device, while I was using a tensor as a target. After some refactoring, I changed the target to a simple list and the error occurred. I traced it down to the following line:
[`return torch.gather(output, 1, torch.tensor(target).reshape(len(output), 1))`](https://github.com/pytorch/captum/blob/master/captum/attr/_utils/common.py#L470)

As it can be seen a tensor is created but no device information is used. A simple fix would be to look where the output tensor lives:
```python
device = "cuda" if output.is_cuda else "cpu"
return torch.gather(output, 1, torch.tensor(target).reshape(len(output), 1).to(device))
# EDIT alternative:
device = output.device
return torch.gather(output, 1, torch.tensor(target, device=device).reshape(len(output), 1))
```
(I don't know if this could cause problems if the data lies on a different GPU, as I have no experience with multiple GPUs)

The tensor version works, as it can be moved before on the user side and only reshapes the used one:
`return torch.gather(output, 1, target.reshape(len(output), 1))`

I've included a minimal example highlighting the problem:

```python
# minimal_example.py
import torch
import torchvision
import numpy as np
from captum.attr import Saliency

# device = "cpu" works!
device = "cuda"
model = torchvision.models.alexnet(pretrained=False)
model.to(device)
model.eval()
X = torch.rand(3, 3, 224, 224, dtype=torch.float)
X = X.to(device)
y_pred = model(X)
saliency = Saliency(model)
X.requires_grad = True
grads = saliency.attribute(X, target=[0, 1])
# working derivations:
# grads = saliency.attribute(X, target=torch.tensor([0, 1, 2]).to(device))
grads.shape
# The reason is the gathering command WITHOUT regarding the current device.
```

vivekmig:
Hi @kai-tub , thanks for the detailed information, you are right, this is definitely a bug! We will push a fix for this soon.

kai-tub:
Yes I would like to create a PR. :)
[...]

PR #317, opened 2020-03-08
Select targets gpu fix

kai-tub:
Hey,
Here is my proposed fix for [#316](https://github.com/pytorch/captum/issues/316).
I've added a test case that uses the BaseGPUTest class and simply runs the saliency target tests again, with everything on the GPU. As I did not fully understand the `_target_batch_test_assert` function and what the requirements for the tests are, I simply copied the function for now. I assume there may be an obvious way to integrate these tests. :)

Another question is:
Do we want to enforce the move if we receive a target tensor, which is not on the right device?
We could print a User Warning similar to automatically setting require gradients as in gradient.py:33.
Or should the user be responsible to move the target tensor to the correct device?

Best regards,
Kai

### DIFF
--- a/captum/attr/_utils/common.py
+++ b/captum/attr/_utils/common.py
@@ -450,6 +450,7 @@ def _select_targets(output: Tensor, target: TargetType) -> Tensor:
 
     num_examples = output.shape[0]
     dims = len(output.shape)
+    device = output.device
     if isinstance(target, (int, tuple)):
         return _verify_select_column(output, target)
     elif isinstance(target, torch.Tensor):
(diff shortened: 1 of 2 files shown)


kai-tub:
On my machine isort is changing it wrong.
I am running isort version 4.3.21
```
> python -m isort
> Skipped 1 files
> python -m isort --settings-path=./.isort.cfg
> Skipped 1 files
```
I am running it on Windows if this has any impact.

NarineK:
> Looks great, thanks so much for adding this fix @kai-tub ! The `_target_batch_test_assert` method works well there, just one small change to avoid code duplication.
> 
> Regarding the question of whether to automatically move the target tensor, I think it might be better for the user to be responsible for this when they provide a tensor. Throughout the algorithms, there are multiple tensor arguments, such as baselines, masks, etc., which are not automatically moved to the appropriate device, so for consistency, it would likely be best to leave the behavior as is for now. What do you think @NarineK ?

Thank you for working on the PR, @kai-tub  and reviewing @vivekmig ! I agree that we don't want to do that check here. If we were, we would have to do it in other places as well and I don't think that it is a critical check. The user will get an error about the wrong device at some point if they provide it incorrectly.

kai-tub:
> Looks great, thanks so much for adding this fix @kai-tub ! The _target_batch_test_assert method works well there, just one small change to avoid code duplication.
Thanks! I've changed the tests accordingly. I probably could've figured that one out on my own.

But I would like to ask what version of isort you are using. It makes a lot of different changes to the current repo and will be rejected. Maybe you could add a note for what specifc version has to be used?

PR #318, opened 2020-03-09
ensure gridlines are turned off in the output

bilalsal:
T61524356


captum/attr/_utils/visualization.py
```diff
@@ -230,6 +230,7 @@ def visualize_image_attr(
     plt_axis.yaxis.set_ticks_position("none")
```
vivekmig: nit: Can possibly either update the comment above (or add a comment below) that grid lines are disabled?

PR #321, opened 2020-03-12
Feature/class summarizer

miguelmartin75:
Basic usage:

```
# ClassSummarizer(stats=common_stats())
summarizer = CommonSummarizer(ClassSummarizer) 

# one input
summarizer.update(torch.Tensor([1, 2, 3]), 1)
# two inputs in one
summarizer.update(torch.Tensor([[1, 2, 3], [4, 5, 6]]), [1, 2])

# aggregate stats
print(summarizer.summary)

# class-based stats
print(summarizer.class_summaries)
```

### DIFF
--- a/captum/attr/_utils/class_summarizer.py
+++ b/captum/attr/_utils/class_summarizer.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
 from collections import defaultdict
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Union
 
 from torch import Tensor
 
 from captum.attr._utils.common import _format_tensor_into_tuples
 from captum.attr._utils.stat import Stat
 from captum.attr._utils.summarizer import Summarizer
+from captum.attr._utils.typing import TargetType, TensorOrTupleOfTensorsGeneric
 
 
 class ClassSummarizer(Summarizer):
@@ -24,9 +25,7 @@ class ClassSummarizer(Summarizer):
         )
 
     def update(
-        self,
-        x: Union[Tensor, Tuple[Tensor, ...]],
-        labels: Optional[Union[Any, List[Any]]] = None,
+        self, x: TensorOrTupleOfTensorsGeneric, labels: TargetType = None,
     ):
         r"""
         Updates the stats of the summarizer, optionally associated to classes.
@@ -38,7 +37,7 @@ class ClassSummarizer(Summarizer):
                 The input tensor to be summarised. The first
                 dimension of this input must be associated to
                 the batch size of the inputs.
-            labels (Any, List[Any], optional):
+            labels (int, tuple, tensor or list, optional):
                 The associated labels for `x`. If Any, we
                 assume `labels` represents the label for all inputs in `x`.
 
@@ -50,17 +49,32 @@ class ClassSummarizer(Summarizer):
 
         x = _format_tensor_into_tuples(x)
 
-        if not isinstance(labels, list):
-            labels = [labels]
+        num_labels = 1
 
-        for tensor in x:
-            assert tensor.size(0) == len(labels)
+        labels_typed: Union[List[Any], Tensor]
+        if isinstance(labels, list) or isinstance(labels, Tensor):
+            labels_typed = labels
+            num_labels = len(labels)  # = labels.size(0) if tensor
+        else:
+            labels_typed = [labels]
 
-        for i, label in enumerate(labels):
+        # mypy doesn't realise I have made the int a list
+        if len(labels_typed) > 1:
+            for x_i in x:
+                assert x_i.size(0) == num_labels, (
+                    "batch size does not equal amount of labels; "
+                    "please ensure length of labels is equal to 1 or the number"
+                )
+
+        batch_size = x[0].size(0)
+
+        for i in range(batch_size):
             tensors_to_summarize = tuple(tensor[i] for tensor in x)
+            tensors_to_summarize_copy = tuple(tensor[i].clone() for tensor in x)
+            label = labels_typed[0] if len(labels_typed) == 1 else labels_typed[i]
 
             self.summaries[label].update(tensors_to_summarize)
-            super().update(tensors_to_summarize)
+            super().update(tensors_to_summarize_copy)
 
     @property
     def class_summaries(self) -> Dict[Any, Dict[str, Tensor]]:
--- a/captum/attr/_utils/class_summarizer.py
+++ b/captum/attr/_utils/class_summarizer.py
@@ -39,7 +39,6 @@ class ClassSummarizer:
             if not isinstance(x, tuple):
                 x = (x,)
 
-            # TODO: how to check for list
             should_resqueeze = torch.zeros(len(x))
             if not isinstance(labels, list):
                 # we need to support the data having
--- a/captum/attr/_utils/class_summarizer.py
+++ b/captum/attr/_utils/class_summarizer.py
@@ -22,7 +22,11 @@ class ClassSummarizer:
         self.all_summary = Summarizer(stats=stats)
         self.summaries = defaultdict(lambda: Summarizer(stats=stats))
 
-    def update(self, x: Union[Tensor, Tuple[Tensor, ...]], labels: Optional[Union[Any, List[Any]]] = None):
+    def update(
+        self,
+        x: Union[Tensor, Tuple[Tensor, ...]],
+        labels: Optional[Union[Any, List[Any]]] = None,
+    ):
         """
         Updates the stats of the summarizer, optionally associated to classes.
 
--- a/captum/attr/_utils/class_summarizer.py
+++ b/captum/attr/_utils/class_summarizer.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import torch
 from torch import Tensor
-from typing import List, Any, Dict, Union, Optional
+from typing import List, Any, Dict, Union, Optional, Tuple
 from collections import defaultdict
 from captum.attr._utils.summarizer import Summarizer
 from captum.attr._utils.stat import Stat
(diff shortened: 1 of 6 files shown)


captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,81 @@
...
+
+class ClassSummarizer:
+    """
+    Used to keep track of summaries for associated classes. The
+    classes/labels can be of any type that is supported by `dict`.
+
+    This also keeps track of an aggregate of all class summaries.
+    """
+
+    def __init__(self, stats: List[Stat] = None):
```
miguelmartin75: remove `= None`

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,81 @@
...
+        self.stats = stats
+        self.all_summary = Summarizer(stats=stats)
+        self.summaries = defaultdict(lambda: Summarizer(stats=stats))
+
+    def update(self, x: Tensor, labels: Optional[Union[Any, List[Any]]] = None):
```
miguelmartin75: `x` should be `Union[Tensor, Tuple[Tensor, ...]]`

captum/attr/_utils/summarizer.py
```diff
@@ -140,14 +140,14 @@ def _reorder_stats(stats):
     return stats, summary_stat_indexs
 
 
-def CommonSummarizer():
+def CommonSummarizer(summarizer_class=Summarizer):
     r"""
     Returns a summarizer with common summary statistics, specifically with:
         Mean, Sample Variance, Sample Std Dev, Min, Max
     """
     from captum.attr._utils.stat import Mean, Var, StdDev, Min, Max
```
NarineK: It looks like the summarizer has the imports everywhere inside the classes and functions. To be consistent with other code can we, please, move the imports to the beginning of the file, @miguelmartin75  ?

captum/attr/_utils/summarizer.py
```diff
@@ -140,14 +140,14 @@ def _reorder_stats(stats):
     return stats, summary_stat_indexs
 
 
-def CommonSummarizer():
+def CommonSummarizer(summarizer_class=Summarizer):
```
NarineK: This feels a bit unusual. Could we probably Instead of having `CommonSummarizer` default all those arguments for the `Summarizer` ? The user can provide another list if they prefer to override it.
```
class Summarizer(stats=[Mean(), Var(order=1), StdDev(order=1), Min(), Max()]):

```
ClassSummarizer will then extend Summarizer since both are Summarizers and it makes sense that `ClassSummarizer` extends `Summarizer`. Let me know.

tests/attr/test_class_summarizer.py
```diff
@@ -0,0 +1,100 @@
...
+import torch
+
+from captum.attr import ClassSummarizer, CommonSummarizer
+
+from .helpers.utils import BaseTest
+
+
+class Test(BaseTest):
+    def class_test(self, data, classes, x_sizes):
+        summarizer = CommonSummarizer(ClassSummarizer)
```
NarineK: I think it would be nice to have something like summarizer = ClassSummarizer(...)

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,89 @@
...
+                the first dimension can optionally be unsqueezed).
+            labels (Any, List[Any], optional):
+                The associated labels for `x`. If Any, we
+                assume x does not represent a batch of inputs.
+
+                If this is None we simply aggregate the total summary.
+        """
+        if labels is not None:
+            if not isinstance(x, tuple):
+                x = (x,)
```
NarineK: nit: you can use `_format_tensor_into_tuples` function from `common.py`

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,89 @@
...
+                dimension of this input must be associated to
+                the amount of `labels` specified; unless there
+                is one labels represents one label (in that case
+                the first dimension can optionally be unsqueezed).
+            labels (Any, List[Any], optional):
+                The associated labels for `x`. If Any, we
+                assume x does not represent a batch of inputs.
+
+                If this is None we simply aggregate the total summary.
+        """
```
NarineK: nit: to avoid nested-ness, I'd probably do something like:
```
if labels is None:
    self.all_summary.update(x)
    return

# The rest of the code goes here
if not isinstance(x, tuple):

.... 

```

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,89 @@
...
+            if not isinstance(labels, list):
+                # we need to support the data having
+                # size (1, ...) or (...)
+                labels = [labels]
+                for i in range(len(x)):
+                    should_resqueeze[i] = len(x[i].size()) == 0 or x[i].size(0) != 1
+                x = tuple(y.unsqueeze(0) for y in x)
+
+            for i, label in enumerate(labels):
+                xx = tuple(
```
NarineK: nit: Can we, please, give more meaningful name to this variable ?

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,89 @@
...
+                    y[i].squeeze(0) if resqueeze else y[i]
+                    for y, resqueeze in zip(x, should_resqueeze)
```
NarineK: nit: can we instead of calling y call it: `x_elem` ?

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,89 @@
...
+            if not isinstance(x, tuple):
+                x = (x,)
+
+            should_resqueeze = torch.zeros(len(x))
+            if not isinstance(labels, list):
+                # we need to support the data having
+                # size (1, ...) or (...)
+                labels = [labels]
+                for i in range(len(x)):
+                    should_resqueeze[i] = len(x[i].size()) == 0 or x[i].size(0) != 1
```
NarineK: nit:  I think that `should_resqueeze` variable sounds a bit unnatural. 
Can we possibly check it differently ?Is this check only necessary for Class Summarizer  ? We didn't have it for the original `Summarizer` ?

captum/attr/_utils/summarizer.py
miguelmartin75: I think I'll convert this function to return the stats used to construct the summarizer, i.e. `CommonStats()`; in-fact I'll move it to `stats.py`

I'll try to make ClassSummarizer extend Summarizer. I decided not to because it seemed like I could just use composition over inheritance.

captum/attr/_utils/class_summarizer.py
miguelmartin75: This is done because the `.update` method accepts a single label or multiple labels associated to a batch of tensors. To make the code work for both, for the single label case the input may need to be unsqueezed in order to do the indexing in the for loop; hence I want to potentially re-perform the squeeze to transform it back to what it was.

The alternative is to make `.update` only accept a single label and single tensor, remove the for loop and squeeze/unsqueeze operations.

NarineK: In `Captum` we always assume that the first dimension is the number of examples in the batch. With `torch.numel` you can check if there is only one element in the label tensor.

Otherwise it becomes more tricky and it might make the code more difficult to maintain. For the labels, I'll actually follow the logic we had for target:

https://github.com/pytorch/captum/blob/master/captum/attr/_utils/common.py#L447

captum/attr/_utils/class_summarizer.py
```diff
@@ -0,0 +1,87 @@
...
+            # for single input we need to ensure
+            # the input is (1, ...) in order to index it
+            # thus we do-so with a squeeze and after we're
+            # done we need to re-squeeze the input to
+            # where it was such that it seems like we did not
+            # touch the given tensor
+            labels = [labels]
+            for i in range(len(x)):
+                should_resqueeze[i] = len(x[i].size()) == 0 or x[i].size(0) != 1
+            batched_x = tuple(y.unsqueeze(0) for y in x)
```
NarineK: I think that we can make an assumption that the first dimension of input is the number of examples ?
We can document it...

tests/attr/test_class_summarizer.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_classes(self):
+        sizes_to_test = [
+            ((1,),),
+            ((3, 2, 10, 3), (1,)),
+            ((20,)),
+        ]
+        classeses = [list(range(100)), ["%d" % i for i in range(100)]]
+        for batch_size in [None, 1, 4]:
+            for sizes, classes in zip(sizes_to_test, classeses):
```
vivekmig: It looks like sizes_to_test and classeses aren't the same length, so testing with size (20,) is skipped, did you intend to add another element to classeses?

nit: Also maybe a name instead of classeses (e.g. class_lists_to_test) might be easier to read?

tests/attr/test_class_summarizer.py
```diff
@@ -0,0 +1,100 @@
...
+
+                data = [
+                    (
+                        tuple(torch.randn(si) for si in sizes_plus_batch),
+                        create_batch_labels(batch_idx),
+                    )
+                    for batch_idx in range(num_batches)
+                ]
+
+                with self.subTest(
```
vivekmig: One thing to note is that pytest doesn't support subTest (which is what we use in CircleCI and sometimes when running locally). I think this will still run all the subtests sequentially as a single test, so it's not harmful to leave this, but might be good to keep in mind if you want the subTest information to show up in CircleCI failures / need them to run independently.

tests/attr/test_class_summarizer.py
miguelmartin75: Nice catch. Will do the suggested nit as well 👍

PR #324, opened 2020-03-15
Fix internal batching bug for cases when attribution is called  `with no_grad()` block

NarineK:
This problem came up as we were using `internal_batch_size` for BERT models.
Autograd was failing to compute the gradients with respect to the inputs (both for IG and layerIG) because grad wasn't enabled for the input slice, this happened because attribute was called in  `with no_grad()` block. Thanks @vivekmig for pointing out to the general issue.

```
inp = torch.rand(3,4)
inp.requires_grad_()
print(inp.requires_grad) # True

with torch.no_grad():
   print((inp*1).requires_grad) # False

```

### DIFF
--- a/captum/attr/_utils/batching.py
+++ b/captum/attr/_utils/batching.py
@@ -126,8 +126,9 @@ def _batched_generator(
     else:
         for current_total in range(0, num_examples, internal_batch_size):
             with torch.autograd.set_grad_enabled(True):
-               inputs_splice = _tuple_splice_range(inputs, current_total,
-                                            current_total + internal_batch_size)
+                inputs_splice = _tuple_splice_range(
+                    inputs, current_total, current_total + internal_batch_size
+                )
             yield inputs_splice, _tuple_splice_range(
                 additional_forward_args,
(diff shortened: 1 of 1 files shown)


captum/attr/_utils/batching.py
```diff
@@ -125,9 +125,11 @@ def _batched_generator(
...
         for current_total in range(0, num_examples, internal_batch_size):
-            yield _tuple_splice_range(
-                inputs, current_total, current_total + internal_batch_size
-            ), _tuple_splice_range(
+            with torch.autograd.set_grad_enabled(True):
+                inputs_splice = _tuple_splice_range(
+                    inputs, current_total, current_total + internal_batch_size
+                )
+            yield inputs_splice, _tuple_splice_range(
                 additional_forward_args,
```
edward-io: why is this not included in the context?

NarineK: This will happen in a general case when people call attribute in a `with no_grad()` block. IG works normally since we activate grads internally but slicing fails.

PR #325, opened 2020-03-16
Move generic commons to upper package

NarineK:
This PR moves generic common functions from `../attr/_utils/common.py` to `../_utils/common.py`.
It also moves type hinting functions from `../attr/_utils/typing.py` to `../_utils/typing.py`.

This is necessary when we want to expand `captum` beyond attribution.


tests/attr/test_common.py
```diff
@@ -37,43 +31,3 @@ def test_validate_nt_type(self) -> None:
         _validate_noise_tunnel_type("smoothgrad", SUPPORTED_NOISE_TUNNEL_TYPES)
         _validate_noise_tunnel_type("smoothgrad_sq", SUPPORTED_NOISE_TUNNEL_TYPES)
         _validate_noise_tunnel_type("vargrad", SUPPORTED_NOISE_TUNNEL_TYPES)
-
-    def test_select_target_2d(self) -> None:
```
vivekmig: It looks like these tests are being deleted, could we possibly move them to a test class corresponding to the generic utils?

NarineK: That one I moved to `test/utils/test_common.py`. It looks like I forgot to push it into PR

NarineK:
> Looks great, thanks for moving these! From a quick look, it seems that safe_div and _extract_device in common might also be generic enough to move up, but we can also move more methods in separate PRs if you prefer.

Thank you! Added those too and moved `tests/attr/helpers/utils.py` to `tests/helpers/basic.py`

PR #326, opened 2020-03-16
Feature/stats summary typehints

miguelmartin75:
Documentation and typehints for `stats.py` and `summarizer.py`

### DIFF
--- a/captum/attr/_utils/stat.py
+++ b/captum/attr/_utils/stat.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 import torch
+from torch import Tensor
+from typing import Dict, Any, Optional, Callable, List
 
 
 class Stat:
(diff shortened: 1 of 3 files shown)


captum/attr/_utils/summarizer.py
```diff
@@ -171,16 +190,38 @@ def __init__(self, stats=None, summary_stats_indices=None):
...
+            x (Tensor):
+                The tensor to summarize
+        """
         for stat in self._stats:
             stat.update(x)
 
-    def get(self, stat):
+    def get(self, stat: Stat) -> Optional[Stat]:
+        r"""
+        Retrieves `stat` from cache if this summarizer contains it
```
vivekmig: Can you clarify in this description / args that the input can be a new stat object with the same class and additional parameters and if the summarizer has a corresponding stat class and parameters it is returned if it exists? From a quick look, it feels that the input and output are the same form the construction of _stat_to_stat, since the overridden hashing / equality logic isn't obvious here.

captum/attr/_utils/summarizer.py
```diff
@@ -145,24 +165,23 @@ def CommonSummarizer():
...
         A simple class that summarizes a single tensor. The basic functionality
         of this class is two operations .update and .summary
+    """
 
+    def __init__(self, stats: List[Stat], summary_stats_indices: List[int]):
+        r"""
         Args:
-            summary_stats (list of Stat): A list of Stat objects you
-                want to show in the .summary property.
             stats (list of Stat): A list of all the Stat objects that
```
vivekmig: It might be good to clarify here that this list needs to already contain all dependencies in the appropriate order / after applying _reorder_stats. It also might be good to not recommend users try to access this class directly and avoid generating documentation for this class on the website.

captum/attr/_utils/stat.py
```diff
@@ -170,21 +214,26 @@ def __init__(self, name=None, order=0):
...
-    def get(self):
+    def get(self) -> Optional[Tensor]:
         var = self.var.get()
         return var ** 0.5 if var is not None else None
 
 
 class GeneralAccumFn(Stat):
-    def __init__(self, fn, name=None):
+    """
+    Performs updaate(x): result = fn(result, x)
```
vivekmig: nit: update

captum/attr/_utils/summarizer.py
miguelmartin75: I'm not sure how to avoid generating documentation for this class, unless I don't document it.

vivekmig: We'll have to add these classes to rst files in order for documentation to be generated, so just a suggestion for that. We'll do that in a separate PR, so don't worry about it for merging this.

PR #330, opened 2020-03-20
Test Refactoring

vivekmig:
This PR switches to dynamic test generation for target tests and Data Parallel tests and adds new JIT tests. All these tests share a common test configuration, which defines valid input settings and the algorithms to which the test case applies to. This format makes it much easier to add tests for new algorithms, generally simply adding a test config for the new algorithm or adding the algorithm to an existing config is sufficient for JIT, target and DataParallel tests to be generated and run.

These additional tests verify a wider range of inputs on all algorithms, which surfaced some new issues. Bugs and issues fixed in this PR include:
1. Additional forward arguments were not working appropriately with DeepLift on a DataParallel model. This is because the device split of expanded additional forward args didn't necessarily match that of inputs. The behavior has been changed to expand the additional args in the hook function (after device split), which ensures the additional args and inputs remain matched.
2. Different targets per example was not working appropriately with DeepLift on a DataParallel model. This is because the model output concatenated the outputs of the devices in DataParallel, which mixed input / baseline outputs, inhibiting appropriate matching between input example and target. Additional forward hooks have been added to appropriately return the output with all inputs followed by all baselines.
3. GradCAM is primarily intended for layers with >= 3 dimensions, since it computes average gradient for each example / channel. For layers with 2 dimensions, the mean gradient over all dimensions was being taken. This has been updated to use the layer gradients directly in this case, which better aligns with the behavior for >= 3 dimensions.
4. DeepLiftShap (and Neuron / Layer variants) were incorrectly repeating additional forward args, this has been fixed to use repeat interleave instead.

In addition to these fixes, some minor changes and additional type hints were added to test models which was necessary for JIT tests, since scripting with non-tensor arguments requires type hints. Some parameters and deltas in other tests are required small changes, since the test model initialization changed, which caused some changes in the seeded random behavior.


NarineK:
@vivekmig , do we have the equivalent command for pytest -ra ? Would you, please, share it in the docs / readme, etc ?

vivekmig:
This approach should work with both pytest and unittest, so `pytest -ra` should still work fine :)

tests/helpers/basic.py
```diff
@@ -1,11 +1,27 @@
...
 import random
 import unittest
+from functools import reduce
 
 import numpy as np
 import torch
 
 
+def get_nested_attr(obj, layer_name):
+    return reduce(getattr, layer_name.split("."), obj)
```
NarineK: Actually, we have exact same function here: 
https://github.com/pytorch/captum/blob/master/captum/attr/_models/base.py#L124
We should be able to reuse it.

NarineK: typehints would be good here also fir the `deep_copy_args`

tests/attr/test_targets.py
```diff
@@ -1,668 +1,205 @@
...
+        args: Dict[str, Any],
+        target_delta: float,
+        noise_tunnel: bool,
+        baseline_distr: bool,
+    ) -> Callable:
+        """
+        This method creates a single target test for the given algorithm and parameters.
+        """
+
+        target_layer = get_nested_attr(model, layer) if layer is not None else None
```
NarineK: The layer is a module here, right ? But it looks like `get_nested_attr ` is taking string ...

tests/attr/test_targets.py
```diff
@@ -1,668 +1,205 @@
...
+    list or tensor.
+    Attribution of each example is computed independently with the appropriate target
+    and compared to the corresponding result of attributing to a batch with a tensor
+    / list of targets.
+    """
+
+    def __new__(cls, name: str, bases: Tuple, attrs: Dict):
+        for test_config in config:
+            algorithms = cast(List[Type[Attribution]], test_config["algorithms"])
+            model = test_config["model"]
```
NarineK: It looks like this piece is being repeated in couple places. I wonder if we could generalize it ?

tests/attr/test_targets.py
```diff
@@ -1,668 +1,205 @@
...
+        target_delta: float,
+        noise_tunnel: bool,
+        baseline_distr: bool,
+    ) -> Callable:
+        """
+        This method creates a single target test for the given algorithm and parameters.
+        """
+
+        target_layer = get_nested_attr(model, layer) if layer is not None else None
+        original_inputs = args["inputs"]
```
NarineK: It looks like that we didn't need to copy the inputs for the data parallel and for the jit but we need it for targets. Maybe you can explain a little bit.

tests/attr/layer/test_grad_cam.py
```diff
@@ -22,7 +22,7 @@ def test_simple_input_non_conv(self) -> None:
     def test_simple_multi_input_non_conv(self) -> None:
         net = BasicModel_MultiLayer(multi_input_module=True)
         inp = torch.tensor([[0.0, 6.0, 0.0]], requires_grad=True)
-        self._grad_cam_test_assert(net, net.relu, inp, ([21.0], [21.0]))
+        self._grad_cam_test_assert(net, net.multi_relu, inp, ([21.0], [21.0]))
```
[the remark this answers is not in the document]
vivekmig: Initially, the relu attribute was either single input or multi-input module based on a flag. This didn't work well with JIT scripting, since it seems to expect a module attribute to have a consistent signature in all paths. I changed it to name the multi-input ReLU module differently which resolved the scripting issue.

tests/attr/helpers/test_config.py
```diff
@@ -0,0 +1,1050 @@
...
+* "dp_delta": Delta for comparison in test_data_parallel
+"""
+
+# Set random seeds to ensure deterministic behavior
+random.seed(1234)
+np.random.seed(1234)
+torch.manual_seed(1234)
+torch.cuda.manual_seed_all(1234)  # type: ignore
+torch.backends.cudnn.deterministic = True  # type: ignore
+
```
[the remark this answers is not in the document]
vivekmig: Good point, sure, will add, thanks!

tests/attr/helpers/test_config.py
```diff
@@ -0,0 +1,1050 @@
...
+            "inputs": torch.arange(400).view(4, 1, 10, 10).float(),
+            "perturbations_per_eval": 8,
+            "sliding_window_shapes": (1, 4, 2),
+            "target": 0,
+        },
+    },
+    {
+        "name": "basic_multi_input_with_perturbations_per_eval_occlusion",
+        "algorithms": [Occlusion],
+        "model": ReLULinearDeepLiftModel(),
```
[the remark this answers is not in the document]
vivekmig: Yeah, I added this to be consistent / cover all cases in the current DataParallel tests, which were testing Occlusion with this model. But I agree, this naming is misleading, will rename this model to ReLULinearModel instead.

tests/attr/test_data_parallel.py
```diff
@@ -1,898 +1,226 @@
...
-                    delta=delta,
+                    # LayerDeepLift and LayerDeepLiftShap do not take device ids
+                    # as a parameter, since they must always have the DataParallel
+                    # model object directly.
+                    # Some neuron methods and GuidedGradCAM also require the
+                    # model and cannot take a forward function.
+                    if issubclass(
+                        internal_algorithm,
+                        (
+                            LayerDeepLift,
```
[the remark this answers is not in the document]
vivekmig: We could for the device_ids, but some methods need the model rather than just a forward function and still take device_ids (NeuronDeconvolution and NeuronGuidedBackprop), which makes this messier.

tests/attr/test_data_parallel.py
```diff
@@ -1,898 +1,226 @@
...
+                assertTensorTuplesAlmostEqual(
+                    self, attributions_1, attributions_2, mode="max", delta=dp_delta
                 )
 
+        return data_parallel_test_assert
+
 
-if __name__ == "__main__":
-    unittest.main()
+class DataParallelTest(BaseGPUTest, metaclass=DataParallelMeta):
```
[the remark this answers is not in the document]
vivekmig: This is necessary because of how pytest collects tests. It doesn't seem to execute the init of a class before collecting the available test methods, so simply adding test attributes in the class init method doesn't work with pytest. This solution uses a meta-class (which is usually type) to dynamically create the attributes, essentially a class factory. The metaclass should extend from "type". The attributes are then available on creating DataParallelTest with the metaclass DataParallelMeta, which creates the tests.

tests/helpers/basic.py
vivekmig: Yeah, I got it from there, but good point, will just access it directly, thanks!

tests/attr/test_data_parallel.py
```diff
@@ -1,898 +1,226 @@
...
+                        raise AssertionError(
+                            "Trying to overwrite existing test with name: %r"
+                            % test_name
+                        )
+                    attrs[test_name] = test_method
+
+        return super(DataParallelMeta, cls).__new__(cls, name, bases, attrs)
+
+    @classmethod
+    @deep_copy_args
```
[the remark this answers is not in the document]
vivekmig: This is primarily needed for test_targets, since we modify the args each time we run each example independently. We can reset it at the end, but if the test errors, it leaves the modified args for the next test. I thought it's safer to deep copy all arguments to ensure that all tests are independent and not dependent on any modifications made in a previous test, so I included this in all methods. Will add a short comment explaining this.

tests/attr/test_targets.py
vivekmig: Targets needs to change the inputs / additional args etc. to test each example independently, will add a comment to make it clearer.

tests/attr/test_jit.py
```diff
@@ -0,0 +1,187 @@
...
+                self.setUp()
+                attributions_2 = attr_method_2.attribute(**args)
+                assertTensorTuplesAlmostEqual(
+                    self, attributions_1, attributions_2, mode="max"
+                )
+
+        return jit_test_assert
+
+
+class JITTest(BaseTest, metaclass=JITMeta):
```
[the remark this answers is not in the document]
vivekmig: This is the same metaclass issue with the other tests, we need this to ensure the test attributes are collected by pytest.

tests/attr/test_targets.py
vivekmig: Yes, needed to this, since I couldn't find a way to reference the model, which is created in one test_config dictionary attribute and access it's attributes to get the layer while constructing the dictionary config. Instead, I'm just including the string path to the module in the config and finding the corresponding module within the test case. The models could possibly be initialized in advance of defining the config dictionary, but it felt cleaner to define the model within each config.

tests/attr/test_data_parallel.py
```diff
@@ -1,898 +1,226 @@
...
+                        algorithm,
+                        model,
+                        target_layer,
+                        args,
+                        dp_delta,
+                        noise_tunnel,
+                        baseline_distr,
+                        mode,
+                    )
+                    test_name = (
```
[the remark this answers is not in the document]
vivekmig: Good point, will create a helper method for this.

tests/attr/test_targets.py
vivekmig: Whoops, looks like the type hint here is wrong, will update, thanks!

vivekmig:
Thanks for the review! I've addressed all comments.

For documentation, I have some detailed explanations of the schema and structure in test_config.py. I've added more detail to it and also added pointers from the other generated test classes to refer to the documentation there on info regarding how to add new tests. Let me know if I should add anything else to make it clearer.

Yes, I agree there's definitely a tradeoff between making the config more concise with more assumptions or leaving it more generic but more repetitive. I considered and iterated through a few structures of the config. When adding layer methods to the same config as general methods, it's not as clear what parameters are passed to which algorithm, so I preferred to keep it structured such that all algorithms in a given test case share the same arguments. I think this shouldn't affect adding new algorithms, since we can usually just add them to existing configs with relevant arguments.

And there shouldn't be any additional commands for running the generated tests, just running with the standard pytest or unittest commands should work fine!

vivekmig:
That's a good point! Right now, for testing, I've been creating a config only including that test and renaming the main config, which is a bit hacky. It would be nice to have a command line argument to take the name, will have to add that feature and test with pytest and unittest. I'll merge this PR and create a separate PR adding that functionality?

PR #333, opened 2020-03-24
Internal Batching Improvements

vivekmig:
This PR improves internal batching to avoid creating an expanded tensor initially and simply recursively compute each subset of steps and add the resulting attributions.

This adds the extra condition that internal_batch_size must be at least equal to the number of examples, and the documentation has been updated appropriately.


captum/attr/_core/integrated_gradients.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, forward_func: Callable) -> None:
                           modification of it
         """
         GradientAttribution.__init__(self, forward_func)
+        self.predefined_step_size_alphas = None  # used for internal batching
```
NarineK: I think this can be dangerous if we call attribute multiple times for different inputs. Does it have to be an instance variable ?

vivekmig: This attribute is only modified in `_batched_attribution` for recursive sub-calls and always reset to None after each operation, so it shouldn't affect any separate attribution calls. The other option is to add another argument to attribute that is only used for this, but it could be confusing to users looking at the method signature, so I preferred this approach.

NarineK: I think that if this gets executed in a multi-threaded environment there can be some issues... 

One way to avoid it could be to move the core logic of ig into an auxiliary function such as: _attribute that also takes start and ends kwargs. `_batched_attribution ` can call  `_attribute` instead of `attribute` ?
https://github.com/pytorch/captum/pull/333/files#diff-67ff19e8dcc2510648379db71c2ee9fbR275

vivekmig: Sounds good, I can switch to that, it would avoid adding a user-facing parameter which would be good! Thanks for the suggestion!

I think calling attribute from different threads on the same attribution object isn't broadly supported by our methods, since we have other cases where we store temp info such as hook lists as an object attribute, which would have the same issue. But I agree that approach is cleaner, will update to it, thanks!

NarineK: Thank you :) yes, that's a good point. Ideally we need to make those hook holders local variables as well. IG is in a more sophisticated shape and more likely to be used than others. We can keep this in mind and make necessary adjustments in other approaches as well.

.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ commands:
     steps:
       - run:
           name: "Check import order with isort"
-          command: isort --check-only
+          command: isort --check-only -v
```
vivekmig: Switching this to verbose to have more details on incorrect orderings in CircleCI

captum/attr/_utils/batching.py
```diff
@@ -12,6 +12,76 @@
...
+    """
+    # Number of steps for each batch
+    step_count = internal_batch_size // num_examples
+    assert (
+        step_count > 0
+    ), "Internal batch size must be at least equal to the number of input examples."
+    if include_endpoint:
+        assert (
+            step_count > 1
+        ), "Internal batch size must be at least twice the number of input examples."
```
NarineK, line 48: Perhaps explaining `why` here would be better. Also, perhaps defaulting like the previous suggestion would be better

NarineK: Does this mean that if the `n_steps=1` then this will always fail because `internal_batch_size` can only be 1 in that case ?
Perhaps making a recommendation message would be helpful.

vivekmig: Yes, that will fail (and probably should), since it would've also failed in the implementation (LayerConductance) where this argument is used, because we take the difference between consecutive steps to estimate gradient, and wouldn't be able to do this with 1 step.

captum/insights/config.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from typing import List, NamedTuple, Optional, Tuple, Dict
+from typing import Dict, List, NamedTuple, Optional, Tuple
```
vivekmig, line 2: Fixing isort failure on master

PR #335, opened 2020-03-25
DataParallel DeepLift Fixes

vivekmig:
This PR adds only the bug fixes identified from refactoring and adding dynamic tests.

1. Additional forward arguments were not working appropriately with DeepLift on a DataParallel model. This is because the device split of expanded additional forward args didn't necessarily match that of inputs. The behavior has been changed to expand the additional args in the hook function (after device split), which ensures the additional args and inputs remain matched.
2. Different targets per example was not working appropriately with DeepLift on a DataParallel model. This is because the model output concatenated the outputs of the devices in DataParallel, which mixed input / baseline outputs, inhibiting appropriate matching between input example and target. Additional forward hooks have been added to appropriately return the output with all inputs followed by all baselines.
3. GradCAM is primarily intended for layers with >= 3 dimensions, since it computes average gradient for each example / channel. For layers with 2 dimensions, the mean gradient over all dimensions was being taken. This has been updated to use the layer gradients directly in this case, which better aligns with the behavior for >= 3 dimensions.
4. DeepLiftShap (and Neuron / Layer variants) were incorrectly repeating additional forward args, this has been fixed to use repeat interleave instead.

### DIFF
--- a/captum/attr/_core/deep_lift.py
+++ b/captum/attr/_core/deep_lift.py
@@ -543,20 +543,6 @@ class DeepLift(GradientAttribution):
                 self.model.register_forward_hook(forward_hook),
             ]  # type: ignore
 
-    def _hook_data_parallel_model(self) -> List[RemovableHandle]:
-        if isinstance(self.model, nn.DataParallel):
-
-            def _dp_hook(module, inputs, outputs):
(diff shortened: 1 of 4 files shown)


vivekmig:
Updated to use 1 hook instead of 2.

NarineK:
> Updated to use 1 hook instead of 2.

Awesome! Thank you! I'll take a look!

PR #337, opened 2020-03-31
Error Message Fix for Class Summarizer

miguelmartin75:


captum/attr/_utils/class_summarizer.py
```diff
@@ -63,7 +63,8 @@ def update(
             for x_i in x:
                 assert x_i.size(0) == num_labels, (
                     "batch size does not equal amount of labels; "
-                    "please ensure length of labels is equal to 1 or the number"
+                    "please ensure length of labels is equal to 1 "
+                    "or the `batch_size` corresponding to the input(s)"
```
NarineK: nit: 
```
 "please ensure that the length of  the labels is equal to 1 "
 "or to the `batch_size` corresponding to the number of examples in the input(s)"
```
How about this ? Minor changes.

PR #341, opened 2020-04-04
fix node version for circleci pt2

edward-io:
move nodejs install to bottom of build script to see if that fixes the problem.


NarineK:
@vivekmig , @edward-io  can we merge this ? I still see build failures related to node.

Issue #143, opened 2019-10-29
Support for LRP/DeepTaylor


EricSpeidel:
Hello,

firstly, thank you for this outstanding library! 
I was wondering whether direct support for [Layer-Wise Relevance Propagation](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0130140) or [Deep Taylor Decomposition](https://www.sciencedirect.com/science/article/pii/S0031320316303582?via%3Dihub) in general is planned or being worked on. 

Since these methods are rather well known, surely they would make a good addition. [Standalone](https://github.com/sebastian-lapuschkin/lrp_toolbox)/[Keras](https://github.com/albermax/innvestigate) implementations are available on GitHub, too.

Thank you for considering this feature request!

Best,
Eric

NarineK:
Hi @EricSpeidel, thank you for bringing this up. Yes, this is on my radar. 

There was also an issue created in pytorch a while ago regarding LRP. 
 https://discuss.pytorch.org/t/layer-wise-relevance-propagation-lrp-in-pytorch/46730

AvantiShri:
@EricSpeidel for what it’s worth, it has been shown that for networks that have only ReLU or maxpooling operations, the “z-rule” of LRP is equivalent to gradient-times-input when biases are included as inputs and the “epsilon” value used for numerical stability is set to 0. This equivalence was (to my knowledge) first noted in the precursor to the DeepLIFT paper (https://arxiv.org/pdf/1605.01713.pdf) and was also shown in a paper by Kindermans et al (https://arxiv.org/pdf/1611.07270.pdf). I think the Kindermans et al proof is more concise/polished, and I have pasted it below for convenience. Of course, it is worth noting that there are many other variants of LRP that are not equivalent to gradient*input and are of interest, particularly the variants involving RNNs.

![9850DFBB-760B-4EE6-86D0-C3B855FD5C03](https://user-images.githubusercontent.com/2302598/69712582-ef6a4480-10b7-11ea-923d-2b151857f3a9.jpeg)

nanohanno:
Hi @NarineK ,
we started working in a research project on attribution methods and specifically LRP. We need an implementation in PyTorch and consider aligning or even adding it to captum. What is your status on the LRP implementation and would you be open to a contribution from our side or some form of collaboration? We would be very happy to add value to your brilliant project.
Best regards,
Hanno

NarineK:
[...]

PR #342, opened 2020-04-06
Add feature Layer-wise Relevance Propagation

nanohanno:
This PR adds the attribution method [Layer-wise relevance propagation](https://doi.org/10.1371/journal.pone.0130140) as discussed in issue #143 . The implementation is inspired by the [tutorial of the inventor's group](https://doi.org/10.1016/j.dsp.2017.10.011) and the publication by [Ancona et al.](https://openreview.net/forum?id=Sy21R9JAW).

So far, three propagation rules are provided (`EpsilonRule`, `GammaRule`, `Alpha1_Beta0_Rule`) inheriting from a Base class `PropagationRule` which should facilitate the implementation of further rules.

This work is part of the government-sponsored project [KI Absicherung](https://www.ki-absicherung.vdali.de/).


nanohanno:
Known open points that may need discussion are:
1. Skip-connections are not supported yet because relevance propagation does not work correctly with tensor addition operations.
2. Inplace ReLUs cause strange behaviour for `LayerLRP`. Shown in testcase `test_lrp_simple_inplaceReLU`
3. Apparently, the backward hook on the input tensor is not called (https://github.com/pytorch/pytorch/issues/35802). This results in problems with additional_forward_args shown in testcase `test_lrp_multi`
4. The returned relevance is normalized instead of amounting to the output score.
5. The lists of supported layers `SUPPORTED_LAYERS_WITH_RULES` and `SUPPORTED_NON_LINEAR_LAYERS` are not complete.

NarineK:
Thank you very much for all the hard work and the PR, @nanohanno ! I'll look into it this week!

NarineK:
Responded on the issue

nanohanno:
Hey @NarineK ,
Sorry for replying late, I was busy with another project and worked on your suggestions during the last days.

I guess we can extend from `GradientAttribution` even though it can be argued that LRP is not really a Gradient method. However, `compute_convergence_delta` would be overwritten, so it should work.
Regarding the deviations to Marco's implementation, I guess the difference is rather the output value than the input values. As mentioned in https://github.com/pytorch/captum/issues/143#issuecomment-608474670 point 1, the output value is not in the multiplication by chain rule because the last module gets 1 as grad value instead of the score.

I tried out your suggestion using `input + 0` as a workaround for https://github.com/pytorch/pytorch/issues/35802%E2%80%A8 . It does work but I had to adjust the `compute_gradients` function to have the adjusted input as the input of the model but calculate gradients w.r.t. the input. The adjusted function is for the beginning in the LRP module, called `compute_gradients_with_adjusted_inputs()` https://github.com/nanohanno/captum/blob/feature/layer_wise_relevance_propagation/captum/attr/_core/lrp.py#L351-L354

I also added a class `Addition_Module(nn.Module)` https://github.com/nanohanno/captum/blob/feature/layer_wise_relevance_propagation/captum/attr/_core/lrp.py#L365-L374 to make skip connections work by exchanging the addition with the custom module. It seems to work but I still observe some strange behaviour and deviations to InputXGradient when applying to an adjusted Resnet model. I am working on that at the moment.

NarineK:
Hi @nanohanno!
Thank you very much  :) 
 	
Yes, it is not a gradient attribution method but since we use the formulations with gradients we can still extend from `GradientAttribution` similar to DeepLift. Yes, `compute_convergence_delta ` can be overwritten. 

Regarding: point 1 in #143 (comment) . I understand that the difference is the output value. My point was a bit different. What I meant is that in you original implementation before doing input adjustment (adjusted_inputs = tuple(input + 0 for input in inputs)) the backward hook on the input tensor wasn't called and since in that input hook you are multiplying grads with inputs that operation was somehow getting missed. Now, I see that you addressed it by removing the multiply by inputs here:
https://github.com/pytorch/captum/pull/342/commits/eb50b98ba29058a960eaaea1134e1407f6e645e3#diff-13ce98e740ae9bc5f643e9c73d2b9a02L170

Thank you for adding `compute_gradients_with_adjusted_inputs `. I was thinking that instead of creating a new `compute_gradients_with_adjusted_inputs ` we can potentially, define a new simple forward func such as:
```
def forward_fn_wrapper(inputs):
     adjusted_inputs = tuple(input + 0 for input in inputs) # of course it would be good to check and make sure that all inputs are tensors
     return self.model(adjusted_inputs)

self.compute_gradients(
            forward_fn_wrapper, inputs, target, additional_forward_args
        )
```

With respect to `Addition_Module `, I was thinking we could maybe move into somewhere in utils into a `custom_modules.py` ? We might add more in the future.

NarineK:
> on-Rule

Thank you for the update @nanohanno ! How large is the deviation ? Is it due to epsilon ? I assume that not because it's a small constant.

NarineK:
> Hey @NarineK ,
> I rebased again because I did not understand the failing test but I did not have in mind that it mixes up the PR. Sorry!
> However, I added the forward function wrapper, the module for custom modules, and extended from GradientAttribution. I also worked on the comments you made in the code, that do not exist anymore due to the rebase _facepalm_. There was an unused `module` argument in some hooks left over from de-bugging and the `torch.sign(outputs)` defaulted to 0 when a output value was 0, leading to division by zero. I added a fix for that now.

Thank you :D

nanohanno:
> Thank you for the update @nanohanno ! How large is the deviation ? Is it due to epsilon ? I assume that not because it's a small constant.

After the recent changes LRP works for VGG and Resnet models. It seems to be equivalent to `InputXGradient` also for Resnet50, the deviations in the attribution output are negligible, on the order of 10-4. 
I guess the strange behaviour I mentioned before is more a problem of finding suitable sets of rules for Resnets, especially the BatchNorm layers (discussed in https://doi.org/10.1007/978-3-030-20518-8_24 , already added a `IdentityRule` to `lrp_rules.py`). I am working on that in the tutorial which I can add in a separate pull request in the future.

NarineK:
Sorry for the delay, @nanohanno ! I'll have another look this week :) Thank you :)

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,218 @@
...
+    model output score represents the initial relevance which is decomposed
+    into values for each neuron of the underlying layers. The decomposition
+    is defined by rules that are chosen for each layer, involving its weights
+    and activations. Details on the model can be found in the original paper
+    [https://doi.org/10.1371/journal.pone.0130140]. The implementation is
+    inspired by the tutorial of the same group
+    [https://doi.org/10.1016/j.dsp.2017.10.011] and the publication by
+    Ancona et al. [https://openreview.net/forum?id=Sy21R9JAW].
+
+    Warning: In-place relus lead to unexptected failures in layer LRP.
```
NarineK, line 28: In layerLRP were you setting any hooks on the ReLUs or accessing the input of the Relu after running forward pass on it ?

nit: `unexpected`

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,129 @@
...
+
+    STABILITY_FACTOR = 1e-9
+
+    def forward_hook(self, module, inputs, outputs):
+        """Register backward hooks on input and output
+        tensors of linear layers in the model."""
+        self._handle_input_hooks = []
+        for input in inputs:
+            if not hasattr(input, "hook_registered"):
+                input_hook = self._create_backward_hook_input(input.data)
```
NarineK, line 22: nit: could we actually check
```
input._backward_hooks
```
instead?
https://github.com/pytorch/pytorch/blob/master/torch/tensor.py#L223

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,129 @@
...
+                self._handle_input_hooks.append(input.register_hook(input_hook))
+                input.hook_registered = True
+        output_hook = self._create_backward_hook_output(outputs.data)
```
NarineK, line 25: nit: usually it is not the case but the output could be a tuple of tensors as well ?

NarineK:
@nanohanno , I left some comments. Let me know if you have any questions. Thank you :)

nanohanno:
Hey @NarineK , sorry for responding so late, at the moment I am not working but will be back end of June. I can try to find some time to address your inline comments to not delay everything. The more general additions need to wait and it might be best to add it in separate PRs. What do you think?

NarineK:
> Hey @NarineK , sorry for responding so late, at the moment I am not working but will be back end of June. I can try to find some time to address your inline comments to not delay everything. The more general additions need to wait and it might be best to add it in separate PRs. What do you think?

Hi @nanohanno! No problem! Thank you for letting us know! Yes, I agree, we can address some of the more generic comments in a separate PR.

captum/attr/_core/layer/layer_lrp.py
nanohanno: Running the commented test case `test_layer_lrp.test_lrp_simple_inplace` and printing the input relevances and the output and input from the ReLUs gives strange outputs:

```
relevance: tensor([0.3333, 0.3333, 0.3333])
module:ReLU() output:(tensor([0.3333, 0.3333, 0.3333]),) input: (tensor([0.3333, 0.3333, 0.3333]),)
relevance: tensor([0.1667, 0.3333, 0.5000])
module:ReLU(inplace=True) output:(tensor([0.0278, 0.0278, 0.0278]),) input: (tensor([0.0278, 0.0278, 0.0278]),)
relevance: tensor([0.0278, 0.0278, 0.0278])
relevance: tensor([0.1667, 0.3333, 0.5000]
```

It appears as if the order in which the hooks are called is different if inplace ReLUs are used. I cannot understand where this is coming from.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+        for module in self.model.modules():
+            if hasattr(module, "rule"):
+                if (
+                    not isinstance(module.rule, PropagationRule)
+                    and module.rule is not None
+                ):
+                    raise TypeError(
+                        (
+                            "Please select propagation rules inherited from class "
+                            "PropagationRule"
```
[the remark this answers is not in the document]
nanohanno: I added the module information

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+                raise TypeError(
+                    (
+                        f"Module type {type(layer)} is not supported."
+                        "No default rule defined."
+                    )
+                )
+
+    def _check_rules(self):
+        for module in self.model.modules():
+            if hasattr(module, "rule"):
```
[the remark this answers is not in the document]
nanohanno: According to the documentation, the apply function uses the modules given by `.children()`. In complex models with subsubmodules the individual modules are not accessed if I remember correctly. Therefore, I used the recursive way in `_get_layers()` to get a list of all individual modules.

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,218 @@
...
+                        corresponding sized tensors is returned. The sum of attributions
+                        is one and not corresponding to the prediction score as in other
+                        implementations.
+            - **delta** (*tensor* or list of *tensors* returned if
+                return_convergence_delta=True):
+
+                        Delta is calculated per example, meaning that the number of
+                        elements in returned delta tensor is equal to the number of
+                        of examples in input.
+                        If attributions for all layers are returned (layer=None) a list
```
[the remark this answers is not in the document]
nanohanno: Good point, it actually belongs to both. If delta is calculated for all layers, the output is a list of attribution tensors and a list of delta values. I added the sentence to attributions too.

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,218 @@
...
+                    delta.append(self.compute_convergence_delta(relevance_layer))
+            else:
+                delta = self.compute_convergence_delta(relevances[0])
+            return _format_attributions(is_layer_tuple, relevances), delta
+        else:
+            return _format_attributions(is_layer_tuple, relevances)
+
+    def _is_layer_tuple(self, inputs, additional_forward_args):
+        if self.layer is None:
+            is_layer_tuple = isinstance(inputs, tuple)
```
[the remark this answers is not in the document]
nanohanno: That sounds tricky. I guess, here, the formatting in `_format_attributions()` is not suitable for the case of layer attributions? If `is_inputs_tuple=True` the attributions are returned as a tuple of the list of attributions from the different layers.
If the intermediate layers have tuple inputs or outputs this is not reflected.

I guess, if the correct format of complex models is supposed to be reflected correctly in the output format, a special format function or even changes in `lrp_rules.py` are necessary to register the layer attributions correctly in the beginning.

What do you think? Did I understand your point?

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,218 @@
...
+        relevances = self._get_output_relevance()
+
+        self._restore_model()
+        undo_gradient_requirements(inputs, gradient_mask)
+
+        if return_convergence_delta:
+            if self.layer is None:
+                delta = []
+                for relevance_layer in relevances[0]:
+                    delta.append(self.compute_convergence_delta(relevance_layer))
```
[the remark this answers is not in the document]
nanohanno: `relevances` is always a tuple of length 1, I think to keep this tuple convention. Then in loops over the list of relevances, which will be only 1 for single layers. So, similar to the point above, we might need a different concept for relevances of intermediate layers with tuple inputs or outputs.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+                        or a tuple of those. It is assumed that attribution
+                        tensor's dimension 0 corresponds to the number of
+                        examples, and if multiple input tensors are provided,
+                        the examples must be aligned appropriately.
+
+        Returns:
+            *tensor*:
+            - **delta** Difference of relevance in output layer and input layer.
+        """
+        remaining_dims = tuple(range(1, len(attributions.shape)))
```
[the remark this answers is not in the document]
nanohanno: True, I added a case for tuple of tensors.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+    """
+
+    def __init__(self, model):
+        """
+        Args:
+
+            model (callable): The forward function of the model or any modification of
+                it. Custom rules for a given layer need to be defined as attribute
+                `module.rule` and need to be of type PropagationRule. If no rule is
+                specified for a layer, a pre-defined default rule for the module type
```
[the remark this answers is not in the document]
nanohanno: Good idea, how can we best link to the documentation? It is built by sphinx, right?

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+        gradient_mask = apply_gradient_requirements(inputs)
+        # 1. Forward pass
+        self._change_weights(inputs, additional_forward_args)
+        self._register_forward_hooks()
+        # 2. Forward pass + backward pass
+        relevances = self.gradient_func(
+            self._forward_fn_wrapper, inputs, target, additional_forward_args
+        )
+
+        self._restore_model()
```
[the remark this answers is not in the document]
nanohanno: Good point, I added that for LRP and LayerLRP.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,345 @@
...
+            if layer.rule is not None:
+                forward_handle = layer.register_forward_pre_hook(
+                    layer.rule.forward_pre_hook_activations
+                )
+                self.forward_handles.append(forward_handle)
+
+    def _change_weights(self, inputs, additional_forward_args):
+        self._register_weight_hooks()
+        _ = _run_forward(self.model, inputs, None, additional_forward_args)
+        self._remove_forward_hooks()
```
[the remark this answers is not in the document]
nanohanno: Added it.

captum/attr/_utils/lrp_rules.py
nanohanno: I think the input tensor might have an output hook because it may be input of the next layer. Therefore only checking `input._backward_hooks` if it is None does not work. One would need to check if it has an input hook registered.

captum/attr/_utils/lrp_rules.py
nanohanno: So far I only observed that the `inputs` of the forward hook were always tuple even if it was of length 1 and outputs were always tensors. But I agree that it could also be tuples in general. Then, I may add a case for outputs of type tuple?

nanohanno:
Hey @NarineK ,
Have you seen the updates I have made? At some point I can also contribute the tutorial. Maybe even in this PR.

NarineK:
> Hey @NarineK ,
> Have you seen the updates I have made? At some point I can also contribute the tutorial. Maybe even in this PR.

Thank you very much for addressing the comments, @nanohanno ! I saw your changes. This week is a bit busy. I'll reply next week or later this week.
To keep the PR concise, I'd recommend to do create the Tutorial in a separate PR. Also, feel free in addition to creating a new tutorial, extend one of our CV tutorials and add an example of LRP as well.  Only, if you want to :) 

This one has IG and Occlusion: https://captum.ai/tutorials/Resnet_TorchVision_Interpret
We could add LRP example as well if you want to :) 

There is another one on CIFAR, if you want to extend that, it works too.
https://captum.ai/tutorials/CIFAR_TorchVision_Interpret

captum/attr/_core/lrp.py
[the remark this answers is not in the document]
NarineK: yes, we use sphinx for documentation. 

We don't have examples of cross referencing code. But I think that we can use
`py:const: `SUPPORTED_LAYERS_WITH_RULES`` for example. We need to experiment and see how it turns out.
https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

Or we can refer to a paper that has default rules described there.

captum/attr/_utils/lrp_rules.py
NarineK: yes, it is not common that the output of a layer is a tuple but it can sometimes happen. We support those cases for other attribution algorithms as well. It would be good to be consistent and support it in LRP as well.
We might have such a test model with a test layer that returns a tuple. Let me see if I can find it and I'll point it out to you.

heytitle:
@nanohanno please let me know if you have any hand with this PR or the LRP tutorials :)

captum/attr/_utils/lrp_rules.py
NarineK: I guess if we assume that the layers are always standard layers such as a linear layer then we can assume that the output is always a tensor but if it is a custom layer then it can potentially return a tuple.

captum/attr/_core/layer/layer_lrp.py
NarineK: I'll try to debug this as well. In-place Relus are pretty common in pre-trained models. e.g. in inception model.

NarineK: @nanohanno, to solve the problem with inplace relus we need to return  `return outputs.clone()` from `forward_hook` in lrp_rules.py.
The problem is that with inplace relus our inputs are being overwritten with the outputs and that's why the ordering also looks weird.

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @staticmethod
+    def backward_hook_activation(module, grad_input, grad_output):
+        """Backward hook to propagate relevance over non-linear activations."""
+        return grad_output
+
+    def _create_backward_hook_input(self, inputs):
+        def _backward_hook_input(grad):
+            relevance = grad * inputs
+            self.relevance_input = relevance.data
```
NarineK, line 36: Here we might be overwriting the relevance scores when the inputs is a tuple of multiple tensors ?

captum/attr/_core/layer/layer_lrp.py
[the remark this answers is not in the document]
NarineK: @nanohanno, Agree that `_is_layer_tuple` isn't suitable for the case when `self.layer is None`. I think that as you suggested the easiest way would be to format them in the `lrp_rules.py`. 
We might also need to make some adjustments here: https://github.com/pytorch/captum/pull/342/files#diff-ac7c4e84c72e959484504ba2f5467e49R36

captum/attr/_core/layer/layer_lrp.py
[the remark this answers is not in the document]
NarineK: Yeah, I think that returning relevances as `return (relevances,) ` in the `_get_output_relevance` is a bit artificial. I think that we can in case of `layer is None` we can always return a list or a tuple of layer relevances that are already correctly formatted in the `lrp_rules.py`. What do you think ?

captum/attr/_core/lrp.py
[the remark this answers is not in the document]
NarineK: I think `self.model.apply` recursively reaches all child nodes in all depths, doesn't it ?

NarineK:
@nanohanno, I think I replied to most of the comments. Let me know if you have any questions!

nanohanno:
Hey @NarineK , thanks for the response. I will take a look at it this week.

captum/attr/_core/lrp.py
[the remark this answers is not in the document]
nanohanno: It gave me trouble at some point as well, but as I understand there has been some discussion about the differences between `children()` and `modules()`. `children()` only returns the `immediate children modules` (https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.children), whereas `modules()` gives `an iterator over all modules in the network.`

captum/attr/_core/lrp.py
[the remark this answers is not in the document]
nanohanno: Okay, does `py:const:` directly go into the docstring? I don't have any experience using sphinx syntax.

captum/attr/_utils/lrp_rules.py
nanohanno: Wow, that is correct. Is it enough to have `self.relevance_input` as list and append to it? It is only used for output of layerLRP.

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,148 @@
...
+    STABILITY_FACTOR is used to assure that no zero divison occurs.
+    """
+
+    STABILITY_FACTOR = 1e-9
+
+    def forward_hook(self, module, inputs, outputs):
+        """Register backward hooks on input and output
+        tensors of linear layers in the model."""
+        self._handle_input_hooks = []
+        for input in inputs:
```
NarineK, line 20: A question: Do we assume that inputs are always a tuple here ?

captum/attr/_utils/lrp_rules.py
NarineK: Usually we keep inputs and outputs as tuples if they consist of more than one tensor. To be precise we can have tuple only if we know that  `_backward_hook_input` will be called only for one input tensor ? 
 https://github.com/pytorch/captum/blob/95cf31104a64eb5f14aa2147852dd9bff885d72b/captum/attr/_utils/lrp_rules.py#L20

captum/attr/_core/lrp.py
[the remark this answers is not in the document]
NarineK: I don't have much experience with it as well. I think we can give it a try that way and if it doesn't display correctly, we can always fix it later.

NarineK:
@nanohanno , we are planning a new release in October. Do you think you'll get chance to address the comments and make it ready by then ?

nanohanno:
Hey @NarineK ,
A new release in October sounds great. I will take that as an incentive to finish the pr. I was planning to work on it during the next two weeks anyways. Do you need type hints for the release?

NarineK:
> Hey @NarineK ,
> A new release in October sounds great. I will take that as an incentive to finish the pr. I was planning to work on it during the next two weeks anyways. Do you need type hints for the release?

Sounds great! Thank you, @nanohanno!
Yes, we need type hints for the release. It shouldn't be very time consuming. We can help with it if you have any questions.

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _manipulate_weights(self, module, inputs, outputs):
+        if hasattr(module, "weight"):
+            module.weight.data = module.weight.data.clamp(min=0)
+        if self.set_bias_to_zero and hasattr(module, "bias"):
+            if module.bias is not None:
+                module.bias.data = torch.zeros_like(module.bias.data)
+
+
+class IdentityRule(EpsilonRule):
```
rGure, line 135: I think this does not work as intended due to the order in which hooks are invoked on intermediate tensors, in particular if this is attached on layer l, the _backward_hook_input on any input tensor of layer l is executed after a putative _backward_hook_output of layer l-1 that is attached to the same intermediate tensor.
It clears the test_lrp_identity due to the relu between the layers as the relu gets a module backward hook attached and therefore does not attach hooks on the intermediate results.

captum/attr/_utils/lrp_rules.py
```diff
@@ -0,0 +1,148 @@
...
+        if hasattr(module, "weight"):
+            module.weight.data = (
+                module.weight.data + self.gamma * module.weight.data.clamp(min=0)
+            )
+        if self.set_bias_to_zero and hasattr(module, "bias"):
+            if module.bias is not None:
+                module.bias.data = torch.zeros_like(module.bias.data)
+
+
+class Alpha1_Beta0_Rule(PropagationRule):
```
rGure, line 112: This is actually not the alpha beta rule but rather the z^+ rule, cf. appendix 1c in https://arxiv.org/abs/1512.02479
It only coincides with the alpha beta rule for alpha=1 iff all inputs are positive.

captum/attr/_utils/lrp_rules.py
nanohanno: Yes, I think that is the assumption.

nanohanno: Yes, I think that is the assumption here.

nanohanno:
Hey @NarineK ,
I added some changes during the last weeks regarding your remarks. I think the remaining bigger topics were focusing on the output types of layer_lrp. I made changes now so that if `layer=None`, meaning relevance is computed for all layers, `lrp_rules` registers a list of tensors for every layer if layer has multiple inputs. This list is converted to a tuple in `_get_output_relevance()` for every layer and finally a list of relevances of all layers is returned. I think these adjustments should tackle your remarks from end of July. Do you think the PR is ready to be merged then?

I would suggest that we close the PR as quickly as possible so that @rGure can add his work. He agreed on supplying the fixes for LRP Rules and type hints, which he worked on already. I am going to add the tutorial in a separate PR. This way we can move on the fastest and have a chance for integration into the next release. Does that sound okay for you?

NarineK:
> Hey @NarineK ,
> I added some changes during the last weeks regarding your remarks. I think the remaining bigger topics were focusing on the output types of layer_lrp. I made changes now so that if `layer=None`, meaning relevance is computed for all layers, `lrp_rules` registers a list of tensors for every layer if layer has multiple inputs. This list is converted to a tuple in `_get_output_relevance()` for every layer and finally a list of relevances of all layers is returned. I think these adjustments should tackle your remarks from end of July. Do you think the PR is ready to be merged then?
> 
> I would suggest that we close the PR as quickly as possible so that @rGure can add his work. He agreed on supplying the fixes for LRP Rules and type hints, which he worked on already. I am going to add the tutorial in a separate PR. This way we can move on the fastest and have a chance for integration into the next release. Does that sound okay for you?

Thank you for addressing the comments @nanohanno ! I'll make another pass today.
Another thing that we have for all attribution algorithms is the support of DataParallels. Basically, if we wrap any model with DataParallel it should return the same result as without. The difference it that DataParallel perform forward and backward passes on different GPUs and aggregate them. 
In case of DataParallel, the hook is called for each GPU separately and we had to implement a mechanism to aggregate them when we wanted to access the input or output of a layer.
https://github.com/pytorch/captum/blob/master/captum/_utils/gradient.py#L262
We might want to look into this. Long term we need to add the support for DataParallels to bee consistent. Currently if it doesn't work with DataParallels, it would be good to document it.
Do you have GPU machines to try it out ? 
Trying DataParallels is very simple:
```
model = DataParallel(model)
```
Just wrap the model with `DataParallel`.
In terms of type hints, that's great that @rGure has it. Regarding new rules, I'll leave it for feature releases but fixes for current rules are welcome.

nanohanno:
> Do you have GPU machines to try it out ?
> Trying DataParallels is very simple:
> 
> ```
> model = DataParallel(model)
> ```
> 
> Just wrap the model with `DataParallel`.
> In terms of type hints, that's great that @rGure has it. Fixes for current rules would be great.

Unfortunately, I do not have GPUs at hand to try it out quickly. I have added the warning statement that data parallel is not yet supported some time ago, similar to what had been used for other algorithms.

captum/attr/_utils/lrp_rules.py
NarineK: @rGure, do you mean that this works in the use case described in `test_lrp_identity ` but will fail if we have two linear layers following each other ?

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+    def test_lrp_basic_attributions(self):
+        model, inputs = _get_basic_config()
+        logits = model(inputs)
+        _, classIndex = torch.max(logits, 1)
+        lrp = LRP(model)
+        relevance, delta = lrp.attribute(
+            inputs, classIndex.item(), return_convergence_delta=True
+        )
+        self.assertEqual(delta.item(), 0)
+        self.assertEqual(relevance.shape, inputs.shape)
```
NarineK, line 81: nit: Could we, please, also assert relevance values ?

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+    def test_lrp_simple_attributions_batch(self):
+        model, inputs = _get_simple_model()
+        model.eval()
+        model.linear.rule = EpsilonRule()
+        model.linear2.rule = EpsilonRule()
+        lrp = LRP(model)
+        inputs = torch.stack((inputs, 3 * inputs))
+        relevance, delta = lrp.attribute(inputs, return_convergence_delta=True)
+        self.assertEqual(relevance.shape, inputs.shape)
+        self.assertEqual(delta.shape[0], inputs.shape[0])
```
NarineK, line 103: nit: I think that it would be good to also assert the values instead of only the shapes.

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+        for model in [model_default, model_inplace]:
+            model.eval()
+            model.linear.rule = EpsilonRule()
+            model.linear2.rule = EpsilonRule()
+        lrp_default = LRP(model_default)
+        lrp_inplace = LRP(model_inplace)
+        relevance_default = lrp_default.attribute(inputs)
+        relevance_inplace = lrp_inplace.attribute(inputs)
+        assertTensorAlmostEqual(self, relevance_default, relevance_inplace)
+
```
NarineK, line 128: nit: Do we want to also add test cases when multiple linear models follow each other ?

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+    def test_lrp_simple_tanh(self):
+        class Model(nn.Module):
+            def __init__(self):
```
NarineK, line 131: nit: perhaps you can add this here and we can reuse it ?
https://github.com/pytorch/captum/blob/master/tests/helpers/basic_models.py#L405

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+                self.linear.weight.data.fill_(0.1)
+                self.tanh = torch.nn.Tanh()
+                self.linear2 = nn.Linear(3, 1, bias=False)
+                self.linear2.weight.data.fill_(0.1)
+
+            def forward(self, x):
+                return self.linear2(self.tanh(self.linear(x)))
+
+        model = Model()
+        _, inputs = _get_simple_model()
```
NarineK, line 143: it's a bit confusing that we have to call `_get_simple_model ` to get the inputs. Could wee, please, define the inputs here ?

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+        _ = model(inputs)
+        lrp = LRP(model)
+        relevance = lrp.attribute(inputs)
+        # assertTensorAlmostEqual(
+        #   self, relevance, torch.Tensor([[0.1186, 0.2372, 0.3558]])
```
NarineK, line 148: Let's remove this commenteed code if it is not needed.

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+            model.linear.weight.data[0][0] = -2
+        model.eval()
+        model.linear.rule = Alpha1_Beta0_Rule()
+        model.linear2.rule = Alpha1_Beta0_Rule()
+        lrp = LRP(model)
+        relevance = lrp.attribute(inputs)
+        assertTensorAlmostEqual(self, relevance, torch.tensor([0.1250, 0.3500, 0.5250]))
+
+    def test_lrp_Identity(self):
+        model, inputs = _get_simple_model()
```
NarineK, line 180: Could we, please, add another test case for testing the point mentioned by @rGure here: https://github.com/pytorch/captum/pull/342/files#diff-ac7c4e84c72e959484504ba2f5467e49R141

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,211 @@
...
+                          The size and dimensionality of the attributions
+                          corresponds to the size and dimensionality of the layer's
+                          input or output depending on whether we attribute to the
+                          inputs or outputs of the layer. If value is None, the
+                          relevance for all layers is returned in attribution.
+        """
+        LayerAttribution.__init__(self, model, layer)
+        LRP.__init__(self, model)
+
+        self._check_rules()
```
NarineK, line 44: Isn't `_check_rules` already being called here ? 
https://github.com/pytorch/captum/blob/1838324879c0bcbccebad71f7e95b9a6eb85550d/captum/attr/_core/lrp.py#L50

captum/attr/_core/layer/layer_lrp.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def __init__(self, model, layer):
+        """
+        Args:
+
+            model (callable): The forward function of the model or
+                        any modification of it. Custom rules for a given layer need to
+                        be defined as attribute
+                        `module.rule` and need to be of type PropagationRule.
+            layer (torch.nn.Module, None): Layer for which attributions are computed.
```
NarineK, line 34: @nanohanno, we have started to support lists of layers instead of just a single layer in some of the algorithms. I think that it would make sense to support here lists of layers as well. Here is an example:
https://github.com/pytorch/captum/blob/master/captum/attr/_core/layer/layer_activation.py#L32 

To be more consistent, I'd, actually, prefer to pass all layers in a list instead of having the convention of layer=None.
If we have that convention here, we have to adjust it in other algorithms as well and computing attributions for all layers isn't very common for other algorithms. Let me know what do you think.

tests/attr/layer/test_layer_lrp.py
```diff
@@ -0,0 +1,173 @@
...
+from ...helpers.basic import BaseTest, assertTensorAlmostEqual
+from ...helpers.basic_models import BasicModel_ConvNet_One_Conv, SimpleLRPModel
+
+
+def _get_basic_config():
+    input = torch.arange(16).view(1, 1, 4, 4).float()
+    return BasicModel_ConvNet_One_Conv(), input
+
+
+def _get_rule_config():
```
NarineK, line 18: nit: is this being used anywhere ?

tests/attr/layer/test_layer_lrp.py
```diff
@@ -0,0 +1,173 @@
...
+    def test_lrp_basic_attributions(self):
+        model, inputs = _get_basic_config()
+        logits = model(inputs)
+        score, classIndex = torch.max(logits, 1)
+        lrp = LayerLRP(model, model.conv1)
+        relevance, delta = lrp.attribute(
+            inputs, classIndex.item(), return_convergence_delta=True
+        )
+        self.assertEqual(delta[0].item(), 0)
+        self.assertEqual(relevance.shape, torch.Size([1, 2, 2, 2]))
```
NarineK, line 72: nit: Can we, please, also assert the values ?

tests/attr/layer/test_layer_lrp.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def test_lrp_basic_attributions(self):
+        model, inputs = _get_basic_config()
+        logits = model(inputs)
+        score, classIndex = torch.max(logits, 1)
+        lrp = LayerLRP(model, model.conv1)
+        relevance, delta = lrp.attribute(
+            inputs, classIndex.item(), return_convergence_delta=True
+        )
+        self.assertEqual(delta[0].item(), 0)
```
NarineK, line 71: nit: Why do we access the first element on delta? If there is only one layer delta should be single valued, not a list shouldn't it ?

```
self.assertEqual(delta.item(), 0)
```

tests/attr/layer/test_layer_lrp.py
```diff
@@ -0,0 +1,173 @@
...
+        model, inputs = _get_simple_model(inplace=False)
+        model.eval()
+        model.linear.rule = EpsilonRule()
+        model.linear2.rule = EpsilonRule()
+        lrp = LayerLRP(model, None)
+        inputs = torch.stack((inputs, 2 * inputs))
+        relevance, delta = lrp.attribute(
+            inputs, attribute_to_layer_input=True, return_convergence_delta=True
+        )
+        self.assertEqual(len(relevance), len(delta))
```
NarineK, line 173: nit: Can we, please, also assert the actual values instead of only the shapes ?

NarineK (commented):
Thank you for working on this @nanohanno! We are almost there 👍 

Overall looks great. I left small nit comments. It probably won't take much time to address them.
Let's get those small suggestions done and merge the code :)

**Regarding the bug fixes and changes that @rGure will make:**
Can we list them here so that we know what will be added in the follow up PR ?


**Regarding DataParallel:**
I understand. If you don't have GPU servers then I'll try to add the support for data parallel.
This can lead to some modification of the existing code.
For example:
When we use DataParallels, PT modules are being replicated and copied on different GPU devices, therefore the following line will breaks:
https://github.com/pytorch/captum/pull/342/files#diff-ac7c4e84c72e959484504ba2f5467e49R68
That is because the module.activations no longer exists.
This line was called for a module replica. That replica was alive for one forward pass only:
https://github.com/pytorch/captum/pull/342/files#diff-ac7c4e84c72e959484504ba2f5467e49R59 

Instead of calling module.activations we might need to get access to original layers and set the value for that layer then scatter those values across all devices and access the correct chunk of activations.

I assume that there will be some more changes required especially for LayerLRP. I can work on those after we merge your changes.

NarineK:
Thank you for doing great job and working on this PR, @nanohanno! Above, I've provided comments about minor fixes. Will you have to address those ?

nanohanno:
Thanks for the feedback. I am already working on it but found some unexpected discrepancies that I have to look into. I will try to address all points until the end of the week.

NarineK:
> Thanks for the feedback. I am already working on it but found some unexpected discrepancies that I have to look into. I will try to address all points until the end of the week.

Awesome! Thank you so much for working on it!

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,358 @@
...
+
+        Returns:
+            *tensor* or tuple of *tensors* of **attributions**
+            or 2-element tuple of **attributions**, **delta**::
+            - **attributions** (*tensor* or tuple of *tensors*):
+                        The propagated relevance values with respect to each
+                        input feature. The values are normalized by the output score
+                        value (sum(relevance)=1). To obtain values comparable to other
+                        methods or implementations these values need to be multiplied
+                        by the output score. Attributions will always
```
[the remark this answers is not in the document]
nanohanno: I changed the relevances to the absolute values now by multiplying by output. It should give comparable results as other methods now.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,358 @@
...
+                        providing the attribution of the corresponding input index.
+                        If a single tensor is provided as inputs, a single tensor is
+                        returned. If a tuple is provided for inputs, a tuple of
+                        corresponding sized tensors is returned. The sum of attributions
+                        is one and not corresponding to the prediction score as in other
+                        implementations.
+            - **delta** (*tensor*, returned if return_convergence_delta=True):
+                        Delta is calculated per example, meaning that the number of
+                        elements in returned delta tensor is equal to the number of
+                        of examples in input.
```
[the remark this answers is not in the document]
nanohanno: I changed the wording

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,358 @@
...
+        return delta
+
+    def _get_layers(self, model):
+        for layer in model.children():
+            if len(list(layer.children())) == 0:
+                self.layers.append(layer)
+            else:
+                self._get_layers(layer)
+
+    def _get_rules(self):
```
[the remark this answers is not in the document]
nanohanno: Changed to `_check_and_attach_rules `

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,358 @@
...
+                self.forward_handles.append(forward_handle)
+
+    def _change_weights(self, inputs, additional_forward_args):
+        try:
+            self._register_weight_hooks()
+            _ = _run_forward(self.model, inputs, None, additional_forward_args)
+        finally:
+            self._remove_forward_hooks()
+        # pre_hooks for 2nd pass
+        self._register_pre_hooks()
```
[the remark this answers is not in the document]
nanohanno: I added more documentation to the block.

captum/attr/_core/lrp.py
```diff
@@ -0,0 +1,358 @@
...
+                        the examples must be aligned appropriately.
+
+        Returns:
+            *tensor*:
+            - **delta** Difference of relevance in output layer and input layer.
+        """
+
+        def _single_attribution_delta(attributions):
+            remaining_dims = tuple(range(1, len(attributions.shape)))
+            delta = 1 - torch.sum(attributions, remaining_dims)
```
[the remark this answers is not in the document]
nanohanno: I adjusted the delta computation to return absolute values and work with tuple outputs. I need to do final checks with tuple and batch outputs.

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+        # (https://github.com/pytorch/pytorch/issues/35802)
+        self.assertTrue(torch.equal(attributions, attributions_add_input))
+
+    def test_lrp_multi_inputs(self):
+        model = BasicModel_MultiLayer()
+        input = torch.Tensor([1, 2, 3])
+        input = (input, 3 * input)
+        lrp = LRP(model)
+        attributions, delta = lrp.attribute(input, return_convergence_delta=True)
+        self.assertEqual(len(input), len(delta))
```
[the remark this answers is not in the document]
nanohanno: I added the value assertion

tests/attr/test_lrp.py
```diff
@@ -0,0 +1,314 @@
...
+        output = model(input)
+        output_add = model(input, add_input=add_input)
+        self.assertTrue(torch.equal(output, output_add))
+        lrp = LRP(model)
+        attributions = lrp.attribute(input)
+        attributions_add_input = lrp.attribute(
+            input, additional_forward_args=(add_input,)
+        )
+        # due to problem with grad() function the results do not match
+        # (https://github.com/pytorch/pytorch/issues/35802)
```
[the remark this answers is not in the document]
nanohanno: True, I removed the comment. The issue is still referenced in the code.

tests/attr/test_lrp.py
nanohanno: Added the value assertion

tests/attr/test_lrp.py
nanohanno: Added the assertion

tests/attr/test_lrp.py
nanohanno: Yes, I added the inputs here

tests/attr/test_lrp.py
nanohanno: Yes, I removed the comments

captum/attr/_core/layer/layer_lrp.py
nanohanno: True, I removed the line.

tests/attr/layer/test_layer_lrp.py
nanohanno: True, it is not used anymore. I removed it now.

tests/attr/layer/test_layer_lrp.py
nanohanno: I made some changes in the block.

rGure:
@NarineK 

For fixes as far as I am aware there are two issues with the current implementation (and both are due to the workaround of splitting the "module_backward_hook" on to the input and output tensors):
1.) hooks on intermediate tensors are attached s.t. the output hook of layer l is attached before the input hook of layer l+1. this means that for the backward pass on this intermediate tensor the output hook of the lower layer is called before the input hook of the higher layer. In the current implementation this only affects the identity rule, as for all other rules both input and output hook are hadamard products which commute...
cf. also https://discuss.pytorch.org/t/altering-tensor-hook-invocation-order/95771/2?u=rgure
I have a workaround for this, which is what @nanohanno meant
2.) There are some issues when the computational graph is not purely sequential, e.g. skip connections.
x -> f(x) + g(x)
then x is the input of both f and g however as it is now there can be only one input hook on x, however if one dropped this one would have to make sure that in hooks of f and g either commute or always are in the correct order. Moreover the gradients are accumulated (the Leibnitz rule) before any hook is called. The most convenient workaround for us was to add no ops into the graph, i.e. x -> f(x+0) + g(x+0), which kind of adds two nodes in between. These nodes however should not end up in the _modules attribute of the torch.nn.Module instance and therefore no hooks are attached ...
This would actually also solve issue 1.
However, I did not find the time yet to check whether I can override the forward function in the LRP code yet to do this automatically.

Other than the above, there is a minor change of the output hook of the identity rule that I would suggest as it inherits the method from Propagation rule but all the arithmetic in that rule are unnecessary, or @nanohanno does this in this pr :)

tests/attr/layer/test_layer_lrp.py
nanohanno: The values of the first relevance tensor are checked now.

captum/attr/_utils/lrp_rules.py
nanohanno: I have added `_format_tensor_into_tuples()` now

captum/attr/_core/layer/layer_lrp.py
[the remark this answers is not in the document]
nanohanno: This was change in a previous commit already. `lrp_rules.py` returns a list of relevances for each layer, which is converted to a tuple in `lrp.py`.

captum/attr/_core/layer/layer_lrp.py
nanohanno: That sounds useful. I changed to lists instead of None, also in the test cases.

captum/attr/_utils/lrp_rules.py
nanohanno: I added the function call now.

tests/attr/layer/test_layer_lrp.py
nanohanno: The delta and relevance values are checked now.

tests/attr/layer/test_layer_lrp.py
nanohanno: The relevance values are checked now.

nanohanno:
Hey @NarineK ,
I have worked on your suggestions and am finished with most points. I added comments where I have made changes.
Open points comprise
- Data Parallel implementation, which you wanted to take a look at
- The topics that @rGure mentioned, which I suggest he may add after merging the PR
- Additional testcases that you suggested in the last review
- Support for tuple output
The last two points seemed to involve some more work and not highest priority. I will see if I have time to work on it in the next days.

NarineK:
Thank you for making the changes @nanohanno ! We released 0.4 and are ready to merge this. The only thing is the new conflict that I see in `__init__.py`. Do you mind fixing it ?

nanohanno:
Congrats to the new release :rocket:, I am looking forward to seeing LRP in v0.4. I updated the branch and resolved the conflict.

NarineK:
> Congrats to the new release 🚀, I am looking forward to seeing LRP in v0.4. I updated the branch and resolved the conflict.

Thank you very much for working on this PR, @nanohanno !

PR #363, opened 2020-04-29
Fixing test_input_x_gradient mypy error

vivekmig:
Fixes mypy error on master in test_input_x_gradient

### DIFF
--- a/tests/attr/test_input_x_gradient.py
+++ b/tests/attr/test_input_x_gradient.py
@@ -55,6 +55,7 @@ class Test(BaseTest):
         nt_type: str = "vanilla",
     ) -> None:
         input_x_grad = InputXGradient(model)
+        attributions: TensorOrTupleOfTensorsGeneric
         if nt_type == "vanilla":
             attributions = input_x_grad.attribute(
                 inputs, additional_forward_args=additional_forward_args


tests/attr/test_input_x_gradient.py
```diff
@@ -55,6 +55,7 @@ def _input_x_gradient_base_assert(
         nt_type: str = "vanilla",
     ) -> None:
         input_x_grad = InputXGradient(model)
+        attributions: TensorOrTupleOfTensorsGeneric
         if nt_type == "vanilla":
             attributions = input_x_grad.attribute(
```
NarineK, line 60: nit: We have attributions defined like this in many test cases but only this one is failing because we didn't specify typehints ?

vivekmig: The case is a little messy here, since generally the type of returned attributions is inferred from the return type of the attribution method. This works in the if case, but the else uses NoiseTunnel, in which we don't actually explicitly type the return. This is because since NoiseTunnel can be used on a LayerAttribution method (and with or without convergence delta), we can't guarantee that the output type (tensor or tuple) doesn't match the input type.
Usually, this doesn't cause a problem, the type is simply inferred as Any. But the if conditions below check the attribution type and rely on the behavior that the input and expected attribution types also match. Since the attributions are typed as Any without any relationship to inputs, the type checker doesn't encode that the inputs and expected_attr should also match attributions, since the conditionals below are on the attribution type and not the input / expected_attr.

I'm not completely sure why the issue surfaced now, but I think it's likely related to PyTorch 1.5 being released, which probably has some more explicit hints, which caused stricter checking of those operations assuming tensor types of inputs and expected_attr.

NarineK: Oh, I see. This can be a little confusing for someone to understand the code if they do not have context. It would probably be better to document that behavior and if we will be able to reproduce it locally.

Issue #358, opened 2020-04-23
Add IBA to captum


berleon:
Hi everyone,

first I want to thank you for this library! It is great to have a place with the most commonly used interpretability methods. We have developed an attribution method that can quantify the importance of areas in bits / pixel. Here is the paper and the reviews: https://openreview.net/forum?id=S1xWh1rYwB  It was accepted for ICLR 2020.  

We developed an easy-to-use library for pytorch (& tensorflow): https://github.com/BioroboticsLab/IBA 

We think it would be a great addition to the many methods already included in captum.
If you are interested in including our method, I could offer to integrate our pytorch code into captum. 

Best,
Leon

cc my co-author: @karl-schulz

NarineK:
Thank you very much for the kind feedback. We definitely would love to support the community and are planning to start with an `awesome list` where people can add their algorithms. Apart from awesome list the next step would be to add an implementation to the contrib package which we will make ready for contributors and as a last step once the algorithms get more popularity we can graduate it and move to the core.

As a starting point we can start from adding it to the awesome list. Then we will review the approach and suggest to add it to contrib package if you agree :)

cc: @caraya10

berleon:
yes, adding it to a `awesome list` would be good first step. Let me know if I can be of any help with the review or with adding it to the contrib package.

PR #367, opened 2020-04-30
Adding initial awesome list

caraya10:
Feel free to add other resources!  Eventually we may want to organize things more but with only 3 things in the list it seems like overkill.

### DIFF
--- a/AWESOME_LIST.md
+++ b/AWESOME_LIST.md
@@ -1,0 +1,19 @@
+# Awesome List
+
+There is a lot of awesome research and development happening out in the interpretability community that we would like to share.  Here we will maintain a curated list of research, implementations and resources.  We would love to learn about more!  Please feel free to make a pull request to contribute to the list.
+
+
+#### TorchRay: Visualization methods for deep CNNs
+TorchRay focuses on attribution, namely the problem of determining which part of the input, usually an image, is responsible for the value computed by a neural network.
+  - [https://github.com/facebookresearch/TorchRay](https://github.com/facebookresearch/TorchRay)
+
+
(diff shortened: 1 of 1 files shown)


NarineK:
> Thanks for adding this! Do we want to have this on the website in addition to GitHub? If so, I think we'd need to move this file to docs and add it to the navigation sidebar.

I think that we decided to do it similar to PyTorch and leave it on github for now. We could put a link to the github page from the web site.

AWESOME_LIST.md
```diff
@@ -0,0 +1,19 @@
+# Awesome List
```
NarineK, line 1: Thank you, @caraya10 ! Do you mind also adding IBA to the list ? https://github.com/pytorch/captum/issues/358

AWESOME_LIST.md
```diff
@@ -0,0 +1,19 @@
...
+There is a lot of awesome research and development happening out in the interpretability community that we would like to share.  Here we will maintain a curated list of research, implementations and resources.  We would love to learn about more!  Please feel free to make a pull request to contribute to the list.
+
+
+#### TorchRay: Visualization methods for deep CNNs
+TorchRay focuses on attribution, namely the problem of determining which part of the input, usually an image, is responsible for the value computed by a neural network.
+  - [https://github.com/facebookresearch/TorchRay](https://github.com/facebookresearch/TorchRay)
+
+
+#### Score Cam: A gradient-free CAM extension
+  - [Paper](https://arxiv.org/abs/1910.01279)
```
NarineK, line 12: nit: Maybe a short description for this ?

PR #368, opened 2020-05-04
add algorithms_comparison_matrix.png

denstern:
Algorithms comparison matrix on site doesn't fit into the screen, so I decided to add the link this table in *.png format on the page's bottom

Hope it will be helpful

### DIFF
--- a/docs/algorithms_comparison_matrix.md
+++ b/docs/algorithms_comparison_matrix.md
@@ -214,4 +214,4 @@ Please, scroll to the right for more details.
 
 **˚ Including Neuron Variant**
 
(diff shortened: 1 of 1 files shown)


denstern:
Link to image
image

This image will be viewed if you follow the link
algorithms_comparison_matrix

NarineK:
Hi @denstern, thank you for adding an image of the algorithm comparison matrix.

Do you mind changing the path to the relative path of the image: `website/static/img/algorithms_comparison_matrix.png`  instead of the absolute path: `http: ... `?

Thank you,
Narine

denstern:
Thanks for your reply @NarineK
Changed to the similar path `/img/algorithms_comparison_matrix.png` as used in the file **captum_insights.md**

PR #369, opened 2020-05-05
Add Occlusion to Insights

edward-io:
image

### DIFF
--- a/captum/insights/config.py
+++ b/captum/insights/config.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from typing import Dict, List, NamedTuple, Optional, Tuple
+from typing import Dict, List, NamedTuple, Optional, Tuple, Callable, Any, Union
 
 from captum.attr import (
     Deconvolution,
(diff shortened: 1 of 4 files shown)


captum/insights/config.py
```diff
@@ -33,20 +43,47 @@ class StrEnumConfig(NamedTuple):
...
+    ),
+    FeatureAblation.get_name(): ConfigParameters(
+        params={"perturbations_per_eval": NumberConfig(value=1, limit=(1, 100)),}
+    ),
+    Occlusion.get_name(): ConfigParameters(
+        params={
+            "sliding_window_shapes": StrConfig(value=""),
+            "strides": StrConfig(value=""),
+        },
+        post_process={
```
J0Nreynolds, line 84: Why are these post-processing steps necessary?

edward-io: Occlusion takes in a tuple or an int e.g. `(3, 5, 5)`, so we need a way to convert a string representation into the appropriate format.

edward-io:
Add fix to UI config parsing

edward-io:
add perturbations_per_eval as a config option to occlusion

PR #380, opened 2020-05-12
Adding segmentation tutorial

vivekmig:
Adds a tutorial for applying Captum to a semantic segmentation model.


abhigoku10:
@vivekmig thanks for doing this feature was more interested, is the merging blocked due to certain reason ?? when iwll it be available

abhigoku10:
@bilalsal @vivekmig Hi when we use the source code where gpu is involved i am getting this following error 
"RuntimeError: forward hooks should never return any values, but '<function _forward_layer_distributed_eval.<locals>.forward_hook at 0x7f996e037e18>'didn't return None
"

vivekmig:
> @bilalsal @vivekmig Hi when we use the source code where gpu is involved i am getting this following error
> "RuntimeError: forward hooks should never return any values, but '<function _forward_layer_distributed_eval..forward_hook at 0x7f996e037e18>'didn't return None
> "

Hi @abhigoku10 , it seems this issue might be related to using an older version of PyTorch, which didn't support hooks. Can you add print(torch.__version__) to your code to check what version you are using? Captum requires at least PyTorch version 1.2, so you may need to upgrade it.

vivekmig:
Thanks for the review :) ! Addressed all comments. For 4, it was necessary to use out_max to sum based on the original labeling of pixels to classes (not after ablating the input), but I've added some comments to make it clearer.

PR #384, opened 2020-05-20
Updates division to be true_division

mruberry:
PyTorch 1.5 deprecated the torch.div() and the division operator performing integer division, and in PyTorch 1.6 this will throw a runtime error. Per discussion with @miguelmartin75, this case should always perform a true division, not a floor (integer) division.


mruberry:
@miguelmartin75 Looks like I finally satisfied lint. Please take a look.

Issue #382, opened 2020-05-14
LayerDeepLift fails when used on a MaxPooling layer?


Holt59:
I am trying to use `LayerDeepLift` on multiple layers of a VGG16 model from `torchvision.models`. It works for all layers except `MaxPooling2D` layers.

The following (layer `23` is a `MaxPool2d` layer):

```python
model = torchvision.models.vgg16(pretrained=True)
u = captum.attr.LayerDeepLift(
    model, list(model.features.children())[23]).attribute(
        torch_im[None, ...], target=156)[0]
```

Raises the following:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-66-668b5d33db17> in <module>
----> 1 u = captum.attr.LayerDeepLift(model, list(model.features.children())[23]).attribute(torch_im[None, ...], target=156)[0]

i:\languages\python\envs\deel-torch\lib\site-packages\captum\attr\_core\layer\layer_deep_lift.py in attribute(self, inputs, baselines, target, additional_forward_args, return_convergence_delta, attribute_to_layer_input, custom_attribution_func)
    306             inputs,
    307             attribute_to_layer_input=attribute_to_layer_input,
--> 308             output_fn=lambda out: chunk_output_fn(out),
    309         )
    310

i:\languages\python\envs\deel-torch\lib\site-packages\captum\attr\_utils\gradient.py in compute_layer_gradients_and_eval(forward_fn, layer, inputs, target_ind, additional_forward_args, gradient_neuron_index, device_ids, attribute_to_layer_input, output_fn)
    517             for layer_tensor in saved_layer[device_id]
    518         )
--> 519         saved_grads = torch.autograd.grad(torch.unbind(output), grad_inputs)
    520         saved_grads = [
    521             saved_grads[i : i + num_tensors]

i:\languages\python\envs\deel-torch\lib\site-packages\torch\autograd\__init__.py in grad(outputs, inputs, grad_outputs, retain_graph, create_graph, only_inputs, allow_unused)
    155     return Variable._execution_engine.run_backward(
    156         outputs, grad_outputs, retain_graph, create_graph,
--> 157         inputs, allow_unused)
    158
    159

i:\languages\python\envs\deel-torch\lib\site-packages\captum\attr\_core\deep_lift.py in _backward_hook(self, module, grad_input, grad_output, eps)
    461         multipliers = tuple(
    462             SUPPORTED_NON_LINEAR[type(module)](
[...]

PR #390, opened 2020-05-25
Assert grad_input and inputs in DeepLift and fix the layer attribution issue for MaxPool

NarineK:
Related to the issue: https://github.com/pytorch/captum/issues/382 asserting `grad_inputs` and `inputs` to have the same shape. More description about the workaround and why the issue happens can be found in the description of the assert. The error occurs when we attribute to the outputs of the layer because of the input or output tensor returned in the forward hook.

Added `forward_hook_with_return_excl_modules` that contains the list of modules for which we don't want to have a return in the forward_hook. This is only used in DeepLift and can be used for any algorithm that attributes to maxpool and at the same time has a backward hook set on it.

Added test cases for layer and neuron use cases.


NarineK:
Hi @vivekmig , that's a good point! In this PR I was focusing on fixing the issue with the Maxpool.
If we remove the backward hook for the relu, it works fine. Which means that we get incorrect behaviors because of the hook and as a solution the fix: 
```
forward_hook_with_return_excl_modules=list(self.SUPPORTED_NON_LINEAR.keys())
``` 
will cover all non-linearities that we hook. Let me know what do you think.

captum/_utils/gradient.py
```diff
@@ -155,6 +155,9 @@ def _forward_layer_distributed_eval(
     target_ind: TargetType = None,
     additional_forward_args: Any = None,
     attribute_to_layer_input: bool = False,
+    forward_hook_with_return_excl_modules: Union[
+        None, List[typing.Type[Module]]
```
vivekmig, line 159: nit: importing Type from typing above might make these expressions a little more concise

PR #406, opened 2020-06-16
Add EmptyFeature to Captum Insights to hide unneeded visualizations

edward-io:
Summary: Hide features that don't generate meaningful interpretable features, e.g. embeddings from other models.

Differential Revision: D22024855

### DIFF
--- a/captum/insights/api.py
+++ b/captum/insights/api.py
@@ -422,27 +422,27 @@ class AttributionVisualizer(object):
         net = self.models[0]  # TODO process multiple models
 
         # initialize baselines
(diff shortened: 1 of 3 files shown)


vivekmig:
Just curious, why do we need an EmptyFeature type? If we don't need attributions for that particular input, we could provide it as an additional arg, and just pass a wrapper function which reorders the inputs rather than providing the original model?

edward-io:
@vivekmig we can still calculate the attributions for the value to show the contribution to the total output, but the feature itself doesn't have a visualization. There's probably better terminology than EmptyFeature, open to suggestions.

PR #415, opened 2020-06-24
LayerDeepLift Hook Fix

vivekmig:
This avoids adding a hook to the target layer non-linearity when using LayerDeepLift if attributing with respect to layer output. This fixes an issue with in-place modules caused by skipping cloning and adds a corresponding test case.


captum/attr/_core/deep_lift.py
```diff
@@ -496,8 +496,10 @@ def _can_register_hook(self, module: Module) -> bool:
             or not self._is_non_linear(module)
         )
 
-    def _register_hooks(self, module: Module) -> None:
-        if not self._can_register_hook(module):
+    def _register_hooks(self, module: Module, skip_target_layer: bool = False) -> None:
```
NarineK, line 499: nit: target layer can be confusing because we also have target output. Perhaps, we can use the same naming for attributing to input or output. `attribute_to_layer_input=True`

PR #419, opened 2020-07-01
Insights reorganization

edward-io:
Move the attribution visualizer into its own folder, `attr_vis`. The public APIs stay the same.

importing from `captum.insights.features` or executing `captum.insights.example` will result in a DeprecationWarning.

Tested by running the example, and verify that unit tests pass with the new module.


vivekmig:
Also, seems like the feature import path will also need to be updated in tutorials?

edward-io:
@vivekmig good catch! Updated the README and Insights related tutorials.

vivekmig:
> @vivekmig good catch! Updated the README and Insights related tutorials.

Great, thanks! Actually, just realized, might want to still include the old path as well in the README until the next release since most users looking at it will be using 0.2.0 rather than master?

Issue #421, opened 2020-07-03
non-leaf tensor warning in NoiseTunnel usage


arnoldjulian:
I get the following warning when using the saliency attribution method in combination with NoiseTunnel:
```
C:\Users\julia\Anaconda3\envs\torch\lib\site-packages\torch\tensor.py:746: UserWarning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the gradient for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations.
  warnings.warn("The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad "
```
A MWE:
```python 
import torch
import torch.nn as nn

from captum.attr import Saliency
from captum.attr import NoiseTunnel


class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc = nn.Linear(5, 2)

    def forward(self, x):
        x = self.fc(x)
        return x


def sal(model, X, node_index):
    sal = Saliency(model)
    grads = sal.attribute(X, node_index).squeeze().detach().numpy()
    return grads


def sal_smooth(model, X, node_index):
    sal = Saliency(model)
    nt = NoiseTunnel(sal)
    grads = nt.attribute(X, target=node_index, nt_type='smoothgrad_sq',
                         n_samples=2, stdevs=0.2).squeeze().detach().numpy()
    return grads


net = Net()

X = torch.rand(1, 5)
X.requires_grad = True

attr = sal_smooth(net, X, 0)
```
When using the saliency attribution method as defined above without NoiseTunnel I do not get any warning.

Based on the API reference on NoiseTunnel and Saliency I would assume that ```requires_grad``` should be set to ```True``` for the input. And indeed, when removing the ```requires_grad``` statement I get the following warning:
```
UserWarning: Input Tensor 0 did not already require gradients, required_grads has been set automatically.
  warnings.warn(
```
The same warning also appears when using the saliency attribution method without NoiseTunnel.
When following the same approach as in the MWE using IntegratedGradients instead of Saliency no warnings are thrown.

Is there a way to use the saliency attribution method in combination with NoiseTunnel without getting any warnings?

NarineK:
Thank you for brining this up, @arnoldjulian !
[...]

PR #426, opened 2020-07-09
Fix require grad warning for non-leaf tensor in noise tunnel

NarineK:
This will fix the warning error specifically related to NoiseTunnel in #421.
In addition to that I moved almost everything under no_grad in the attribute method. This will hopefully also help with runtime performance.
In the `_forward_layer_eval ` I had to add `grad_enabled ` flag in order to allow to enable the gradients externally. As it is also needed in `test_neuron_gradient.py` test case.

### DIFF
--- a/captum/_utils/gradient.py
+++ b/captum/_utils/gradient.py
@@ -135,6 +135,7 @@ def _forward_layer_eval(
     additional_forward_args: Any = None,
     device_ids: Union[None, List[int]] = None,
     attribute_to_layer_input: bool = False,
+    grad_enabled: bool = False,
 ) -> Tuple[Tuple[Tensor, ...], Literal[True, False]]:
     return _forward_layer_eval_with_neuron_grads(
         forward_fn,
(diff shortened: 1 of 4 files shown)


captum/_utils/gradient.py
```diff
@@ -357,13 +362,16 @@ def _forward_layer_eval_with_neuron_grads(
...
-    saved_layer, is_layer_tuple = _forward_layer_distributed_eval(
-        forward_fn,
-        inputs,
-        layer,
-        additional_forward_args=additional_forward_args,
-        attribute_to_layer_input=attribute_to_layer_input,
-    )
+    grad_enabled = True if gradient_neuron_index is not None or grad_enabled else False
+
+    with torch.autograd.set_grad_enabled(grad_enabled):
```
NarineK, line 367: Another option is to leave here `set_grad_enabled(True)`

NarineK:
> Looks good, thanks for fixing this :) !
> 
> One question related to #421 , for NoiseTunnel with Saliency (or other methods that compute gradients directly wrt the provided input), the user will now always get this warning, since the input to the method will not require grad?
> 
> ```
> UserWarning: Input Tensor 0 did not already require gradients, required_grads has been set automatically.
>   warnings.warn(
> ```
> 
> To avoid this, we can internally require grad on inputs_with_noise for a gradient based method. What do you think?

Yeah, that's a good point! Added require_grads for GradientAttribution methods.

PR #428, opened 2020-07-11
Update the resnet tutorial 

wangyongjie-ntu:
I think there are errors in the current tutorial. Without specifying the baseline, the IG and Occulude will use the default all-zero vectors as baseline internally.  By IG and DeepLIFT, the baseline should be a black image (all-zero vector in image space, not the input space).  Therefore, we should first create an all-zero vector,  then normalize the all-zero vectors with the mean and std before feeding into the interpreted methods.


NarineK:
Hi @wangyongjie-ntu , thank you for the PR! I think the assumption is that all-zero tensor is in the normalized input space. Do you see significant difference in the attribution ? Have you tried to compare the attribution heatmaps ?
Do you mind posting them side-by-side ?

PR #436, opened 2020-08-06
Supporting non C contigous tensors

csabapol:
Replaced the view function with reshape. Reshape() can operate on both contiguous and non-contiguous tensors while view() can only operate on contiguous tensors. According to the [documentation](https://pytorch.org/docs/master/generated/torch.reshape.html#torch.reshape) the reshape() returns a view of the input when possible. 
Error:
"RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead."

### DIFF
--- a/captum/attr/_utils/attribution.py
+++ b/captum/attr/_utils/attribution.py
@@ -269,7 +269,9 @@ class GradientAttribution(Attribution):
         _validate_target(num_samples, target)
 
         def _sum_rows(input: Tensor) -> Tensor:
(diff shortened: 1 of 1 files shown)


NarineK:
Thank you for the PR, @csabapol ! Is it possible to add a test case for this. Perhaps when the `compute_convergence_delta` is called externally for any of the attribution algorithms.

NarineK:
@csabapo, do you mind rebasing this PR ? We will merge it. Thank you!

Issue #437, opened 2020-08-10
The documentation example for visualize_image_attr_multiple seems puzzling 


mikeaalv:
It seems to me the input in this example for visualize_image_attr_multiple is not the same as the argument format for the function `captum.attr.visualization.visualize_image_attr_multiple(attr, original_image, methods, signs, titles=None, fig_size=8, 6, use_pyplot=True, **kwargs)`. It seems to me the part of `methods` and `signs` are in front in this example. I haven't test whether the example can run as it need modification (input data) to run. 

```
>>> # ImageClassifier takes a single input tensor of images Nx3x32x32,
>>> # and returns an Nx10 tensor of class probabilities.
>>> net = ImageClassifier()
>>> ig = IntegratedGradients(net)
>>> # Computes integrated gradients for class 3 for a given image .
>>> attribution, delta = ig.attribute(orig_image, target=3)
>>> # Displays original image and heat map visualization of
>>> # computed attributions side by side.
>>> _ = visualize_mutliple_image_attr(["original_image", "heat_map"],
>>>                     ["all", "positive"], attribution, orig_image)
```

vivekmig:
Hi @mikeaalv , thanks for catching this! You are right, the input ordering is incorrect in this example, attribution and orig_image should be provided first. Would you be able to create a PR fixing this? Thanks!

mikeaalv:
@vivekmig  Definitely glad to do this.

vivekmig:
Resolved in #438 , thanks for fixing this issue @mikeaalv !

PR #438, opened 2020-08-10
[MRG] update the document example for function visualize_mutliple_image_attr

mikeaalv:
Make the changes to ensure the argument order is the same as stated in function description for visualize_mutliple_image_attr

1. I didn't pass `pytest -ra` but I figured it should be fine as I didn't change the code (it's only documentation modification). 
2. I try to check the documents locally by `./scripts/build_docs.sh` which didn't give me the page for API descriptions.

Otherwise, the format checking has all been run. Do let me know if the above two error is crucial and I need to make some changes. 

This is related to the issue #437


mikeaalv:
There seems also a block on merging but I assumed the reviewer will have the permission

PR #443, opened 2020-08-14
GPU CI Fixes

vivekmig:
This fixes CircleCI GPU tests as well as fixing test failures in DataParallel and Distributed Data Parallel tests due to torch changes in 1.5.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,10 +121,11 @@ commands:
       - run:
           name: "Install CUDA"
           command: |
-            wget http://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-ubuntu1604-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
-            sudo apt-get update
-            export OS_RELEASE=$(uname -r)
-            sudo dpkg -i cuda-repo-ubuntu1604-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
+            wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/cuda-ubuntu1604.pin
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -121,10 +121,11 @@ commands:
...
           command: |
-            wget http://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-ubuntu1604-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
-            sudo apt-get update
-            export OS_RELEASE=$(uname -r)
-            sudo dpkg -i cuda-repo-ubuntu1604-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
+            wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/cuda-ubuntu1604.pin
+            sudo mv cuda-ubuntu1604.pin /etc/apt/preferences.d/cuda-repository-pin-600
+            wget http://developer.download.nvidia.com/compute/cuda/10.2/Prod/local_installers/cuda-repo-ubuntu1604-10-2-local-10.2.89-440.33.01_1.0-1_amd64.deb
+            sudo dpkg -i cuda-repo-ubuntu1604-10-2-local-10.2.89-440.33.01_1.0-1_amd64.deb
+            sudo apt-key add /var/cuda-repo-10-2-local-10.2.89-440.33.01/7fa2af80.pub
```
NarineK, line 128: It looks like we don't need to do this two steps ?
```
sudo apt-get update
export OS_RELEASE=$(uname -r)
```

vivekmig: Yes, I think they're not needed, this is from the official install [instructions](https://developer.nvidia.com/cuda-10.2-download-archive?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=1604&target_type=deblocal), and seems like it installed CUDA appropriately and the DataParallel tests all ran.

PR #456, opened 2020-08-25
Multi Layer Support

vivekmig:
This adds support to obtain attributions for multiple layer for LayerActivation and LayerGradientXActivation with corresponding changes to helper methods to support multiple layers. Also adds appropriate DataParallel tests and updates documentation.


captum/attr/_core/layer/layer_conductance.py
```diff
@@ -356,11 +356,7 @@ def _attribute(
 
         # Conductance Gradients - Returns gradient of output with respect to
         # hidden layer and hidden layer evaluated at each input.
```
NarineK, line 358: Do we want to update the documentation for the return type ? Currently we have: 
```
                       Attributions are returned in a tuple based on whether
                        the layer inputs / outputs are contained in a tuple
                        from a forward hook. For standard modules, inputs of
                        a single tensor are usually wrapped in a tuple, while
                        outputs of a single tensor are not.
```
https://github.com/pytorch/captum/blob/3b9bb227ad6889a2438d25ab085d9e014a1b15ad/captum/attr/_core/layer/layer_conductance.py#L245
This might apply to other methods too.

NarineK (approved):
Looks good! Thank you for working on this PR.
Minor comments related to `torch.nn.Sequential` and documentation.

captum/attr/_core/layer/layer_conductance.py
vivekmig: Good point! Will update all the docs.

PR #459, opened 2020-08-26
Updating Formatting

vivekmig:
CircleCI is currently failing due to a new version of black, which seems to check for no leading white-space in docstrings. Runs updated black to resolve formatting issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,7 +183,6 @@ jobs:
       - conda_install:
           args: "-n"
       - lint_flake8
-      - lint_black
       - isort
       - unit_tests
       - sphinx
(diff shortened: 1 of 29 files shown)


NarineK:
Thank you for fixing it, @vivekmig ! Which version of black did it upgrade to ?

vivekmig:
> Thank you for fixing it, @vivekmig ! Which version of black did it upgrade to ?

pip now has version 20.8b1 (see [here](https://pypi.org/project/black/#description) for more details), seems like it was released today, which started causing the failures. [Anaconda](https://anaconda.org/conda-forge/black) still seems to have the previous version 19.10b0 (which is now failing tests on this PR with the updates made my 20.8b1), I assume they will release the same version to anaconda shortly. Otherwise, can temporarily disable black from conda if the versions are still not matching.

PR #461, opened 2020-09-01
switch to new google analytics id

edward-io:
Summary: we were using botorch's google analytics ID, whoops

Differential Revision: D23459973

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -28,7 +28,7 @@ const siteConfig = {
   projectName: "captum",
 
   // Google analytics
-  gaTrackingId: "UA-139570076-2",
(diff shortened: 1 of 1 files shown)


website/siteConfig.js
```diff
@@ -28,7 +28,7 @@ const siteConfig = {
   projectName: "captum",
 
   // Google analytics
-  gaTrackingId: "UA-139570076-2",
+  gaTrackingId: "UA-177018179-1",
```
NarineK, line 31: @edward-io, this is the old id, isn't it ?

NarineK:
@edward-io , I guess we can close it without merging.

PR #474, opened 2020-09-21
Updating Citation

vivekmig:
Updates citation file to reference Captum paper on arXiv


luyang-ai4med:
Do you have a peer-reviewed citation version that we could use (workshops, conference)? For some journals, preprints are not allowed. Thank you.

Issue #469, opened 2020-09-17
Error when running the script 'Multimodal_VQA_Captum_Insights.ipynb' using GPU


allenguo92:
## 🐛 Bug

When I run the notebook file Multimodal_VQA_Captum_Insights.ipynb(under 'tutorials') via jupyter notebook, for the last step, it shows the error message "AttributeError: 'NoneType' object has no attribute 'norm'".
## To Reproduce

1.Download and import relevant project 'pytorch-vqa' and 'pytorch-resnet' and pretrained VQA model '2017-08-04_00.55.19.pth' under my own work folder.
2.Change the project and pretrained model path for relevant variables.
3.Run the notebook step by step and finally click the button 'Fetch' after running the command 'visualizer.render()'.

Error messages:
Captum Insights widget created.
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/anaconda3/lib/python3.7/site-packages/captum/insights/widget/widget.py in _fetch_data(self, change)
     39         with self.out:
     40             self.visualizer._update_config(self.config)
---> 41             self.output = namedtuple_to_dict(self.visualizer.visualize())
     42             self.config = dict()
     43 

~/anaconda3/lib/python3.7/site-packages/captum/insights/api.py in visualize(self)
    526         while len(self._outputs) < self._config.num_examples:
    527             try:
--> 528                 self._outputs.extend(self._get_outputs())
    529             except StopIteration:
    530                 break

~/anaconda3/lib/python3.7/site-packages/captum/insights/api.py in _get_outputs(self)
    515             internal_batch_size=1,  # should be 1 until we have batch label support
    516         ):
--> 517             output = self._calculate_vis_output(inputs, additional_forward_args, label)
    518             if output is not None:
    519                 cache = SampleCache(inputs, additional_forward_args, label)

~/anaconda3/lib/python3.7/site-packages/captum/insights/api.py in _calculate_vis_output(self, inputs, additional_forward_args, label, target)
    434             net,
    435             tuple(transformed_inputs),
--> 436             additional_forward_args=additional_forward_args,
    437         )
    438 

~/anaconda3/lib/python3.7/site-packages/captum/attr/_utils/common.py in _run_forward(forward_func, inputs, target, additional_forward_args)
    501         *(*inputs, *additional_forward_args)
    502         if additional_forward_args is not None
--> 503         else inputs
    504     )
    505     return _select_targets(output, target)

[...]

PR #477, opened 2020-09-23
Avoid forward_hook under DataParallel

bilalsal:
Address same problem in the VQA Insights tutorial reported in issue #469


NarineK:
Thank you for the PR @bilalsal ! I originally tested the notebook for VQA in 8 GPU environment and it worked fine back then. Did you test this in a GPU environment ? Maybe you can send me the details in a local thread.

vivekmig:
@bilalsal @NarineK - Are we still planning to make this change to the tutorial or should we just close this PR? Alternatively, we can revert the other tutorial to the hook approach so that they are consistent.

NarineK:
Since based on this change: https://github.com/pytorch/captum/commit/174ecbdf926a312d07c0ef728540455b1c2c1edb
VQA in the notebook already has some differences from the insights one I think we can close this PR and revisit if there will be any issues. What do you think, @bilalsal ?

NarineK:
Closing it, since the issue got addressed

PR #495, opened 2020-10-15
Neuron Aggregation

vivekmig:
This adds support for neuron aggregation, neuron_selector can be a function which returns a custom aggregate of a layer's neurons for all neuron methods other than neuron conductance, which has dependence on output gradients. The neuron_index argument was renamed, and a deprecation decorator was added to provide a warning for usage of the old parameter as a keyword argument. This decorator can be removed prior to the 0.4.0 release.

Documentation of the new callable functionality has been added to NeuronDeepLift, this documentation will be propagated to other relevant methods after review.

### DIFF
--- a/captum/_utils/common.py
+++ b/captum/_utils/common.py
@@ -417,6 +417,15 @@ def _select_targets(output: Tensor, target: TargetType) -> Tensor:
         raise AssertionError("Target type %r is not valid." % target)
 
 
+def _contains_slice(target: Union[int, Tuple[Union[int, slice], ...]]) -> bool:
+    if isinstance(target, tuple):
+        for index in target:
+            if isinstance(index, slice):
+                return True
+        return False
+    return isinstance(target, slice)
+
+
 def _verify_select_column(
     output: Tensor, target: Union[int, Tuple[Union[int, slice], ...]]
 ) -> Tensor:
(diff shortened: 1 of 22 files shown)


captum/attr/_core/neuron/neuron_gradient_shap.py
```diff
@@ -116,7 +118,7 @@ def attribute(
                         that for all given input tensors, dimension 0 corresponds
                         to the number of examples, and if multiple input tensors
                         are provided, the examples must be aligned appropriately.
-            neuron_index (int or tuple): Index of neuron or neurons in output of
+            neuron_selector (int or tuple): Index of neuron or neurons in output of
```
edward-io, line 121: do these docstrings need to be updated as well?

captum/_utils/common.py
```diff
@@ -427,6 +436,24 @@ def _verify_select_column(
...
 
 
+def _verify_select_neuron(
+    layer_output: Tuple[Tensor, ...],
+    selector: Union[int, Tuple[Union[int, slice], ...], Callable],
+) -> Tensor:
+    if callable(selector):
+        return selector(layer_output if len(layer_output) > 1 else layer_output[0])
+
+    assert len(layer_output) == 1, (
```
NarineK, line 446: since we are allowing slice and summing after we do this check, will this work as expected ?
I think we probably need to add some test cases when the output per example is not a scalar.

NarineK (approved):
Looks great! Thank you for working on this, @vivekmig!
A question: Are we allowing slice dt for neuron_selector and summing per example across all neurons ? According to the documentation Callable can return scalar per example but it can also return slice, can't it ?

captum/_utils/gradient.py
```diff
@@ -423,7 +423,9 @@ def _forward_layer_eval_with_neuron_grads(
     evals in a dictionary protected by a lock, analogous to the gather implementation
     for the core PyTorch DataParallel implementation.
     """
-    grad_enabled = True if gradient_neuron_index is not None or grad_enabled else False
+    grad_enabled = (
+        True if gradient_neuron_selector is not None or grad_enabled else False
```
miguelmartin75, line 427: nit: redundant if-else, `grad_enabled = gradient_neuron_selector is not None or grad_enabled`

miguelmartin75 (approved):

captum/_utils/gradient.py
```diff
@@ -511,7 +513,7 @@ def compute_layer_gradients_and_eval(
     inputs: Union[Tensor, Tuple[Tensor, ...]],
     target_ind: TargetType = None,
     additional_forward_args: Any = None,
-    gradient_neuron_index: Union[None, int, Tuple[int, ...]] = None,
+    gradient_neuron_selector: Union[None, int, Tuple[int, ...], Callable] = None,
```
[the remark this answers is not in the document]
vivekmig: Good point, will update these type hints.

Issue #498, opened 2020-10-20
Type / Casting Error in ShapleyValueSampling


rbtsbg:
## 🐛 Bug

In the attribute method of ShapleyValueSampling, the variable total_attrib is derived from the input tensor and receives the input tensor's type: 

https://github.com/pytorch/captum/blob/b9d88a9f0b65c25f55337ab8b6617abffe87acc7/captum/attr/_core/shapley_value.py#L296 

This raises a RuntimeError here: 
https://github.com/pytorch/captum/blob/b9d88a9f0b65c25f55337ab8b6617abffe87acc7/captum/attr/_core/shapley_value.py#L355 
when a tensor of type Long is passed as input and Python tries to cast a Float into a Long. 

## To Reproduce

1. Pass to ShapleyValueSampling a forward_func which accepts tensors of type Long, i.e. token indices to a BERT model. 
1. Call `attribute` with input tensors of type Long.  
1. A RuntimeError is thrown: 
```
...
File "/lib/python3.8/site-packages/captum/attr/_core/shapley_value.py", line 350, in attribute
    total_attrib[j] += (
RuntimeError: result type Float can't be cast to the desired output type Long
```
## Expected behavior

ShapleyValueSampling should be able to handle input tensors of type Long, such as token indices to NLP models.

## Fix 

add `dtype=torch.float` to  the `zeros_like` constructor here: 

https://github.com/pytorch/captum/blob/b9d88a9f0b65c25f55337ab8b6617abffe87acc7/captum/attr/_core/shapley_value.py#L296 

Edit: This turned out to be insufficient. Maybe casting right away solves the issue. 

## Environment

```

 - Captum / PyTorch Version: 0.2.0 / 1.6.0
 - OS: Linux
 - How you installed Captum / PyTorch: pip
 - Python version: 3.8.5
 - CUDA/cuDNN version: 9.1.85

bilalsal:
Thank you for proposing this fix, Robert!

NarineK:
It looks like this issue got fixed with: https://github.com/pytorch/captum/pull/515
Closing ...

PR #499, opened 2020-10-20
Fixed type error in ShapleyValueSampling

rbtsbg:
Bug and fix are described in issue #498

All tests passed on my local machine.

### DIFF
--- a/captum/attr/_core/shapley_value.py
+++ b/captum/attr/_core/shapley_value.py
@@ -294,7 +294,9 @@ class ShapleyValueSampling(PerturbationAttribution):
 
             # Initialize attribution totals and counts
             total_attrib = [
-                torch.zeros_like(input[0:1] if agg_output_mode else input, dtype=torch.float)
+                torch.zeros_like(
+                    input[0:1] if agg_output_mode else input, dtype=torch.float
+                )
                 for input in inputs
             ]
 
--- a/captum/attr/_core/shapley_value.py
+++ b/captum/attr/_core/shapley_value.py
@@ -265,6 +265,14 @@ class ShapleyValueSampling(PerturbationAttribution):
         # Keeps track whether original input is a tuple or not before
         # converting it into a tuple.
         is_inputs_tuple = _is_tuple(inputs)
+
+        # Inputs that are subtypes of float will cause errors further down.
+        # To allow such inputs, (i.e. torch.int64 token indices for NLP models), we convert immediately.
+        if is_inputs_tuple:
+            inputs = tuple([t.float() for t in inputs])
+        else:
+            inputs = inputs.float()
+
         inputs, baselines = _format_input_baseline(inputs, baselines)
         additional_forward_args = _format_additional_forward_args(
             additional_forward_args
--- a/captum/attr/_core/shapley_value.py
+++ b/captum/attr/_core/shapley_value.py
@@ -267,7 +267,7 @@ class ShapleyValueSampling(PerturbationAttribution):
         is_inputs_tuple = _is_tuple(inputs)
 
         # Inputs that are subtypes of float will cause errors further down.
-        # To allow such inputs, (i.e. torch.int64 token indices for NLP models), we convert immediately.
+        # To allow such inputs we convert immediately.
         if is_inputs_tuple:
             inputs = tuple([t.float() for t in inputs])
         else:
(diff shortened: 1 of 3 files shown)


vivekmig (approved):
Thanks for finding and fixing this issue, looks great 👍 ! There seems to be a formatting failure due to the line length in the CircleCI results, would you be able to run black and flake8?

It would also be great to add a test case with integer output to ensure this case gets tested in the future, examples of these tests for feature ablation are in #207 . If you can add a test or two for Shapley Value Sampling as well, that would be great! Otherwise we can also add that separately, either way works. Thanks again for the contribution!

rbtsbg:
No problem! Sure, I can do that in a couple of days

rbtsbg:
It was a good idea to add and run a few tests. Turned out, the fix I proposed in issue #498 wasn't enough. 

The checks failed again, even though I ran flake8 and black. I followed the checks listed here: 

https://github.com/pytorch/captum/blob/master/.circleci/config.yml

and it appears that 

https://github.com/pytorch/captum/blob/b9d88a9f0b65c25f55337ab8b6617abffe87acc7/.circleci/config.yml#L67

fails. I do not think that this has anything to do with the changes I made, at least it doesn't appear so at first glance.

captum/attr/_core/shapley_value.py
```diff
@@ -265,6 +265,14 @@ def attribute(
         # Keeps track whether original input is a tuple or not before
         # converting it into a tuple.
         is_inputs_tuple = _is_tuple(inputs)
+
+        # Inputs that are subtypes of float will cause errors further down.
+        # To allow such inputs we convert immediately.
+        if is_inputs_tuple:
+            inputs = tuple([t.float() for t in inputs])
+        else:
+            inputs = inputs.float()
```
vivekmig, line 274: I'm not sure if we want to convert the inputs here, this assumes that the model will support taking a float tensor as input, which may not be the case for embedding inputs etc. The input used to run the model's forward pass should ideally maintain the same type as the input provided for attribution.

tests/attr/test_shapley.py
```diff
@@ -27,6 +27,18 @@ def test_simple_shapley_sampling(self) -> None:
             n_samples=250,
         )
 
+    # Test above, but with inputs of type long.
+    def test_sss_with_long_tensor_inputs(self) -> None:
+        net = BasicModel_MultiLayer()
+        inp = torch.tensor([[20, 50, 30]], dtype=torch.int64)
```
vivekmig, line 33: I think the float conversion added above is needed for this test to pass since the linear layer in BasicModel_MultiLayer expects a float tensor input. It would be better to update these tests with a model that accepts a Long input, which would be representative of cases where the user provides a Long input for the model.  One option might be BasicModel, it returns an output with the same type as the provided input and is used for the int tests in #207 as well. Just 1-2 cases should be sufficient, ideally covering both long input with long output and long input with float output.

vivekmig (commented):
Thanks for adding the test cases 👍 ! Just a few comments on the added float conversion.

captum/attr/_core/shapley_value.py
rbtsbg: Fair point; the type assumptions are a bit too strong. I'll find a workaround.

vivekmig: Sounds good, thanks!

vivekmig:
Hi @rbtsbg , we are planning to do our next release for Captum 0.3.0 early next week. Would you be able to update this fix before then? If not, it's no problem, we can fix the type separately to include it in the release, and we can merge your additional test cases whenever you have time.

rbtsbg:
Hi @vivekmig, I apologize for the delay. I have a paper deadline Nov 23 and planned to propose the fix shortly afterwards.

I am afraid, I won't be able to fix it until the deadline, so please go ahead and include your own changes in the release.

Thanks a lot!

vivekmig:
> Hi @vivekmig, I apologize for the delay. I have a paper deadline Nov 23 and planned to propose the fix shortly afterwards.
> 
> I am afraid, I won't be able to fix it until the deadline, so please go ahead and include your own changes in the release.
> 
> Thanks a lot!

Sounds good, no problem, good luck with the paper deadline!

vivekmig:
Closing this PR since the fix and tests were merged here in #515 . Thanks again for identifying this issue @rbtsbg !

PR #501, opened 2020-10-23
GradCAM Test Fix

vivekmig:
Latest PyTorch is causing test failures in GradCam, since in-place modification of a tensor requiring grad now fails (https://github.com/pytorch/pytorch/pull/46204). Updates to set requires grad after in-place modification.


NarineK (approved):
Looks great!
@greentfrapp caught this on optim-wip branch yesterday too.

Issue #509, opened 2020-11-04
Running Insight's inside Docker container


ahtonen:
I would like to run the nice Insights Flask app in Docker since my whole Deep Learning stack is built in Dockerfile. But for some reason running `CIFAR_TorchVision_Captum_Insights.ipynb` gave strange "model not found" error although it printed nicely that the widget was created. First I though the problem was in notebook widget installation, but after double checking that I did it exactly according instructions, I started to look Docker related things.

Running exactly same example works when I run it directly in my host, Ubuntu 18.04 LTS. Next I went to bash shell, inside the running container where Captum was also installed, launched example by `python -m captum.insights.example`, took another bash inside Docker and used `netcat` to connect given port and voila. The connection works.

I'm not expert with Flask, but I suspect that once the Flask server thread is spawned it by default accepts incoming connections only from host it runs on. In a Docker container's case that is by default some IP from `172.17.0.0/16`. That would explain why running it in 8008, for example, in Docker container, exposing 8008 to host and still the connection does not work. My host's network is `192.168.16.0/24`.

Has anyone had similar problems so that I could verify the root cause? And also hints how to overcome it would be greatly appreciated.

btw. I didn't find anything related running in Docker in FAQs, community forums, etc. That's why opening this question.

orionr:
cc @edward-io for Captum Insights (and potentially @seemethere for Docker)

bilalsal:
Hi @ahtonen,

yes, mapping ports in Docker is tricky, I hope you find hints [here](https://docs.docker.com/config/containers/container-networking/).

ahtonen:
The problem is not with mapping ports in Docker, cause I have already expose the port I run Insights in Docker to localhost. The root cause that I found out yesteday is that  `server.py` Flask app line does not include `host='0.0.0.0'` flag. Changing it to:
```
def run_app(debug: bool = True):
    app.run(port=port, use_reloader=False, debug=debug, host='0.0.0.0')
``` 

Fixes the problem. Then Flask app allow connections from other IPs than the Docker container's one.

Shall I make a pull request?

ahtonen:
... continuing this I think that the flag could be optional and propagated from `AttributionVisualizer.serve` and `AttributionVisualizer.render`. I.e. like
```
visualiser.serve(debug=True, bind_all=True)
```

PR #513, opened 2020-11-06
Fix issue with Flask app not accepting connections outside localhost.

ahtonen:
_Proposal_ for fixing #509. This allows running Insights inside Docker container and connecting to Flask app via port that is exposed to localhost (e.g. 6006).

Note: I couldn't get Insights widget to work, however, since I'm not expert with widgets. Probably when running `render` port is assigned automatically, since I don't see `port` argument there. To make connection to work, port would need to be set manually and binding flag passed in like in `server` method. Maybe somebody more knowledgeable with widgets could make that change?

### DIFF
--- a/captum/insights/attr_vis/app.py
+++ b/captum/insights/attr_vis/app.py
@@ -229,17 +229,21 @@ class AttributionVisualizer(object):
             display(widget.out)
 
     @log_usage()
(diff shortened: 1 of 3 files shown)


orionr:
Thank you @ahtonen! @edward-io thoughts on this?

edward-io (approved):
Looks great! Thanks for the fix.

PR #514, opened 2020-11-09
Add docs for metrics

NarineK:
Adding docs for metrics on the website and README file.


docs/algorithms.md
```diff
@@ -3,15 +3,18 @@ id: algorithms
...
 
-The algorithms in Captum are separated into three groups, primary attribution, layer attribution and neuron attribution, which are defined as follows:
+The attribution algorithms in Captum are separated into three groups, primary attribution, layer attribution and neuron attribution, which are defined as follows:
 * Primary Attribution: Evaluates contribution of each input feature to the output of a model.
 * Layer Attribution: Evaluates contribution of each neuron in a given layer to the output of the model.
 * Neuron Attribution: Evaluates contribution of each input feature on the activation of a particular hidden neuron.
 
 Below is a short summary of the various methods currently implemented for primary, layer, and neuron attribution within Captum, as well as noise tunnel, which can be used to smooth the results of any attribution method.
 
+Beside attribution algorithms Captum also offers metrics to estimate the trustworthiness of the explanations and model predictions.
```
vivekmig, line 15: nit: Can we just say trustworthiness of explanations here? This seems to suggest that they can be used to estimate trustworthiness of model predictions as well, although it's primarily the explanations?

vivekmig (approved):
Looks great, thanks for adding this! We'll also need to add these to rst files to show up in the API documentation, do you want to do that in a separate diff or add that here?

sphinx/source/metrics.rst
```diff
@@ -0,0 +1,15 @@
+Metrics
+======
```
NarineK, line 2: @vivekmig, updated RSTs as well. I guess this should be sufficient ?

vivekmig: I think we need to also add metrics to index.rst so that this is included in the main API list. You can also run ./scripts/build_docs locally to preview the website and double-check the docs look ok if you want.

PR #515, opened 2020-11-09
Fix Shapley Value for Non-Float Input

vivekmig:
Addresses #498 by adding torch.float when initializing attr and adding corresponding test cases for integer input.


NarineK (approved):
LGTM! Thank you!

PR #517, opened 2020-11-09
Lime and KernelSHAP Documentation

vivekmig:
This diff exposes Lime and KernelSHAP from captum.attr and adds documentation for these methods to algorithms.md as well as to appropriate rst files for API generation.


NarineK (approved):
Looks great! Thank you for adding the docs!

PR #518, opened 2020-11-10
Optim wip - Fix PixelImage

ProGamerGov:
Just a quick fix so that PixelImage works now. I also fixed a spelling mistake in the main optim README.


ProGamerGov:
@NarineK It should be ready for you to merge as I only changed where a single line of code was.

NarineK (approved):
Awesome! Thank you!

PR #528, opened 2020-11-16
Ablation Device Fix

vivekmig:
Summary: Allows output to be on a different device than input by moving output difference to input device.

Differential Revision: D25001273


captum/attr/_core/feature_ablation.py
```diff
@@ -348,7 +348,7 @@ def attribute(
                         ).reshape((-1, num_outputs) + (len(inputs[i].shape) - 1) * (1,))
                     if self.use_weights:
                         weights[i] += current_mask.float().sum(dim=0)
-                    total_attrib[i] += (eval_diff * current_mask.to(attrib_type)).sum(
+                    total_attrib[i] += (eval_diff.to(total_attrib[i].device) * current_mask.to(attrib_type)).sum(
```
NarineK, line 351: looks like there is a flake8 issue.

NarineK (approved):
LGTM!
It looks like there is a flake8 issue.

PR #531, opened 2020-11-19
Add dlrm tutorial

NarineK:
Adding DLRM tutorial and the KDD presentation slides related to it.
The actual model is 2.1GB, which it pretty big. Git allows 100MB maximum. Either we need to override max size or put the model elsewhere.

Update
............
Storing the model on aws S3


vivekmig (approved):
Looks great :) ! This is a really nice addition for a recommender system tutorial! Just had some questions a few minor suggestions.

•Feature Importance: 
- "the primary signal for Non-Clicked is coming from the dense features." -> Related to this, I'm curious if the baseline sparse feature (empty ID list / 0 embedding output) is itself a strong signal for Non-Click. If so, this effect might not be hidden in attributions when baseline matches the input. Might be interesting to look at the output click score for just the baseline input, whether it's closer to 0.5 click / no-click or as a baseline a no-click? 

• Interaction Layer Importance
- nit: Interaction Layer Importance - > 27 26 0.5 = 351 - Multiplication seems to be used in formatting italics
- nit: "those interactions have no affect on" -> no effect on
- Since the attributions are being summed, is it possible that either positive / negative attributions cancel for some features or that some features have effects on only a smaller percentage of examples? Am curious whether the interaction neurons with 0 attribution have no effect for any examples or a larger effect on a small percentage of examples

• Neuron Importance
- For this section could we potentially just use LayerConductance on the last FC layer instead? It seems the analysis of 0 and negative attributions focuses on the sum of attributions for all features for each neuron, which seems to align better with layer attribution? Do we need the extra information of each input feature contribution to the neuron?
- nit: The Neuron Indices - Accumulated Attribution plot is pretty large, could maybe shrink it in some way (or show only a subset?) or group neurons to make it easier to read?

NarineK:
**Feature Importance**

- That's an interesting point. Since sparse features are anonymizes we don't know exactly what they mean but we assume that their presence is more likely to be connected to the `Clicked` Ads.  When I set both sparse and dense features to 0s prediction score is ~0.2. We would ideally want baseline to be neutral and in this case be somewhere close to ~0.5. Perhaps it would be good to try out other baselines as well. When I zero out the sparse embeddings only for some samples predictions score goes down in different magnitudes and for a few it goes up.  When I zero out dense features I also see that the prediction score goes down, for some special cases it might also go up.  For the dense feature all-zero values might also itself be strong signal for Non-Click - It's hard to tell. I'll mention about it in the tutorial

I tried another experiment for probs_above_0999.  If I zero out dense features for the half of the samples the prediction scores go up to 1.0.  If I zero out sparse features the prediction scores go down for all samples below 0.5.

**Interaction Layer Importance**
- That's a good question. Since the computation of attributions is time consuming I haven't run it for larger samples. I made the observations only for a small sample of ads that were predicted as `Clicked` with relatively large probability score.  It is worth it to note that it would be good to run it for larger samples to confirm the hypothesis that we make.  Regarding the summation of attribution scores: Yes, it is possible that positives and negatives get cancelled out. e.g. I found 2 indices such as 5 and 6 for which attributions are 0 across all 497 examples. I'll add a brief description about it in the tutorial.

Neuron Importance
* That's a good point. I originally Used NeuronIntegrated Gradients in the KDD presentation and later for this tutorial switched to NeuronConductance. We can use LayerConductance instead since we are looking into all neurons in the layer.
* I'll visualize for 10 samples only. It is visually more clear.

PR #542, opened 2020-11-30
Fix grammatical mistakes in README

ProGamerGov:
I pasted the README into a Google Doc to see if the spelling and grammar check would find anything. It was able to find a few mistakes that I've now corrected.


NarineK (approved):
Looks great! Thank you, @ProGamerGov!

Issue #544, opened 2020-12-01
Ability to return HTML object from visualize_text()


cdpierse:
## 🚀 Feature

The visualization function `visualize_text()` currently displays an an IPython `HTML` object if the user is inside a notebook. I'd like to propose a feature that allows this function to return either the `HTML` object constructed from the fields in `dom` or even just to return the contents of `dom` itself. This would allow for outputs from the function to be saved easily as right now I finding it difficult to save the visualization if not inside a notebook. 

## Pitch

I think it could be implemented as something as simple as the example below, essentially just an additional flag parameter that returns the HTML object if true. 
```python
from captum.attr import visualization as viz
.... # get attributions and data record


html_obj = viz.visualize_text([score_viz], return_html=True)

```

If you think this would make sense I'd be happy to put a PR together for it.

bilalsal:
Hi Charles,

that is a good proposal.
Yes, it would be great if you could draft a PR - Thank you!

cdpierse:
No problem at all, put in a PR (#548) for this earlier.

bilalsal:
PR merged into master

heyCoke:
Returning an HTML object, how do I save this object?

shanto-Rahman:
I am getting the error when adding return_html=True. It says, 
html_data = viz.visualize_text(vis_data_records_ig, return_html=True)
TypeError: visualize_text() got an unexpected keyword argument 'return_html'

PR #548, opened 2020-12-03
Add option to return HMTL object from visualize_text()

cdpierse:
Related issue: #544 

Adds flag `return_html`to `visualize_text()` that allows for the IPython HTML to be returned if the flag is set to `True`.

This is useful for cases where users may want to save the output from outside of a notebook etc. 

Usage looks like:

```python
from captum.attr import visualization as viz
.... # get attributions and data record


html_obj = viz.visualize_text([score_viz], return_html=True)
```


bilalsal:
Thanks @cdpierse for putting this together!

@miguelmartin75 and @vivekmig: do you prefer we add return_html as a parameter? Or to always return the HTML content instead? (the routine currently does not return any value(.

miguelmartin75:
> Thanks @cdpierse for putting this together!
> 
> @miguelmartin75 and @vivekmig: do you prefer we add return_html as a parameter? Or to always return the HTML content instead? (the routine currently does not return any value(.

I don't see any reason for why we couldn't always return HTML, since this method doesn't return anything and the HTML content is always being constructed.

NarineK:
Feel free to return HTML. To avoid additional input argument, we can always return HTML and the user can chose to use it or not.

cdpierse:
I've updated the PR so that HTML is now just always returned by default. Thanks.

PR #551, opened 2020-12-07
Add ability to compare multiple models in Captum Insights

Reubend:
This PR adds the ability to compare multiple models in Captum Insights.

Screenshot of model comparison

In order to test this, I went through two scenarios. First, I made sure there are no regressions to single model workflows like this:
1. Start the Insights example with `python3 -m captum.insights.example`
2. Ensure that the original functionality is still working and there are no changes, other than the visual changes for column headers

Then, I tested comparing multiple models by duplicating the existing example one:

1. Go to `example.py`
2. Duplicate the example model, by changing `models=[model]` to `models=[model, model, model]`
3. Check to make sure that it renders properly, and that selecting different target classes works to properly update the data for each visualization


edward-io (approved):
Great, thanks!

PR #556, opened 2020-12-11
disable js map files to reduce package size

edward-io:
- Clean up some JS warnings
- remove unpkg
- switch to plotlyjs-basic-dist-min


NarineK (approved):
Thank you, @edward-io ! It looks like internal tests are failing. Do you mind investigating it ?

Issue #549, opened 2020-12-04
n_perturb_samples on captum.metrics.sensitivity_max


gkakogeorgiou:
The n_perturb_samples variable is also used in Lime definition. This means that you can not use n_perturb_samples to **kwargs on metrics.sensitivity_max, in case you want to use Sensitivity algorithm on Lime for specific n_perturb_samples. I propose to change the name of n_perturb_samples on sensitivity_max algorithm.

NarineK:
Thank you for bringing this up, @gkakogeorgiou !
`n_perturb_samples` is used in both `sensitivity_max ` and `infidelity` metrics.
For Lime and KernelSHAP we can probably rename that variable to `n_surrogate_samples` ?
What do you think, @vivekmig ?

vivekmig:
Yeah, this is a good point! We should avoid argument name overlaps with metrics. Could we possibly just use `n_samples` in Lime and KernelSHAP, which would match other attribution methods?

NarineK:
`n_samples` sounds good. One thing only - `n_samples` is also used in noise tunnel. It is less likely that someone will use Lime and KernelSHAP w/noise tunnel but just in case ?

NarineK:
This got resolved with #558 , @gkakogeorgiou !
We will create a patch release to reflect these changes. Thank you for bringing it up.

PR #558, opened 2020-12-11
Deprecate and rename conflicting argument names

NarineK:
As also mentioned in #549, we had conflicting argument names in the API. In this PR we deprecate and rename those argument names. More specifically:
1. `n_samples` in NoiseTunnel is being renamed to `nt_samples`
2. `n_perturbed_samples` in Lime and KernelSHAP are being renamed to `n_samples` in order to remain consistent with Shapely Values (Sampling)


captum/attr/_utils/common.py
```diff
@@ -367,23 +367,68 @@ def _construct_default_feature_mask(
...
+
+
+def noise_tunnel_n_samples_deprecation_decorator(func: Callable) -> Callable:
+    r"""
+    Decorator to depricate n_samples parameter for NoiseTunnel's `attribute` method.
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args: Any, **kwargs: Any):
+        if "n_samples" in kwargs:
```
vivekmig, line 399: I think this condition can also check that nt_samples is not already in kwargs? Otherwise, it wouldn't be possible to pass both for Lime and KernelShap until this is deprecated right?

vivekmig (approved):
Looks great, thanks for updating this :) ! I think we might also need to update the generated tests here https://github.com/pytorch/captum/blob/master/tests/attr/helpers/test_config.py with the updated names, but can also leave it for now and update when we deprecate to continue testing the older names, either way works.

captum/attr/_utils/common.py
NarineK: yeah, that's a good point!

PR #575, opened 2020-12-28
Remove redundant 'object' inheritance reference from classes

ProGamerGov:
As I understand, Captum uses Python 3, which means that classes don't need to inherit from object as it done implicitly already.


NarineK (approved):
LGTM! Thank you, @ProGamerGov !

PR #552, opened 2020-12-08
optim-wip: Fix objectives.py, images.py, RedirectedReLU etc.

greentfrapp:
- `InputOptimization` no longer requires target_modules (inferred from `loss_function`)
- Consolidate `ImageTensor` and `CudaImageTensor`
- Update `RedirectedRELU` to only allow "wrong" gradients when grad tensor is zero
- Make losses composable
- Move optim tutorials to separate folder under `tutorials/optim`

PR #579, opened 2021-01-06
Optim-wip: Add Activation Atlas tutorial & functions

ProGamerGov:
This PR implements support for Activation Atlases and Class Activation Atlases, based on the Activation Atlas research paper: https://distill.pub/2019/activation-atlas/

* Added atlas functions and tests for them. 

* Added full documentation for all the new atlas functions.

* Added new Activation Atlas tutorial. The corresponding Lucid tutorial notebook can be found [here](https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-simple.ipynb).

* Added new Class Activation Atlas tutorial. The corresponding Lucid tutorial notebook can be found [here](https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/class-activation-atlas.ipynb).

* Both atlas tutorials share some of their text cells and code cells, so keep that in mind when reviewing them.

* I re-added the `RandomRotation` transform as the Torchvision one does not accept lists of degree values. Also added tests for the new `RandomRotation`  transform.

* Vectorized the `weights_to_heatmap_2d` heatmap function so that it's faster and more efficient. Also improved the tests for the function.

* Fixed `nchannels_to_rgb` so that it works properly with CUDA inputs, and also improved the tests for the function. This function isn't used by the atlas related code and tutorials, so it's a bit out of scope for this PR.

* Added `umap-learn` package to the tutorial requirements in `setup.py`. See: https://umap-learn.readthedocs.io/en/latest/ & https://arxiv.org/abs/1802.03426 for more information about Uniform Manifold Approximation and Projection for Dimension Reduction (UMAP). UMAP is used for calculating the atlas structure.

* Added a relaxed version of the MaxPool2d class for calculating neuron class attributions.

* Added new Activation Atlas sample collection tutorial. The corresponding Lucid tutorial notebook can be found [here](https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb). It might be better if we had an easy to download dataset for demonstration purposes. Also, should we have the sample collection functions be inside Captum, or should they only be in the sample collection tutorial?

Atlases might also make a good banner image for Captum as well.


ProGamerGov:
@NarineK Do you have any ideas on how to generate irregular grids of (x,y) coordinates for tests?

ProGamerGov:
This PR doesn't overlap with https://github.com/pytorch/captum/pull/552, so it can be reviewed at any time.

ProGamerGov:
@Narine So, Lucid calculates class attributions for neurons using this function:



Source: https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb

https://github.com/pytorch/pytorch/issues/10223, https://github.com/pytorch/rfcs/pull/11

ProGamerGov:
@Narine So, Lucid calculates class attributions for neurons using this function:

```
def fwd_gradients(ys, xs, d_xs):
  
  """Forward-mode pushforward analogous to the pullback defined by tf.gradients.
  With tf.gradients, grad_ys is the vector being pulled back, and here d_xs is
  the vector being pushed forward.
  
  By mattjj@google.com from
  https://github.com/renmengye/tensorflow-forward-ad/issues/2
  """
  
  v = tf.zeros_like(ys)
  g = tf.gradients(ys, xs, grad_ys=v)
  return tf.gradients(g, v, grad_ys=d_xs)
```

Source: https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb

When I looking looking up how to recreate the same basic algorithm, I found that the above may not be currently possible in PyTorch as I think it's Forward AD: https://pytorch.org/docs/stable/autograd.html#torch.autograd.functional.jvp

> we don’t have support for forward mode AD in PyTorch at the moment.

Looking at PyTorch's development of the required feature makes it seem like it isn't coming anytime soon: https://github.com/pytorch/pytorch/issues/10223, https://github.com/pytorch/rfcs/pull/11


So, is there a way I can use Captum to calculate how much a neuron contributes to the different labels / classes of a model?

ProGamerGov:
@NarineK So, Lucid calculates class attributions for neurons using this function:

```
def fwd_gradients(ys, xs, d_xs):
  
  """Forward-mode pushforward analogous to the pullback defined by tf.gradients.
  With tf.gradients, grad_ys is the vector being pulled back, and here d_xs is
  the vector being pushed forward.
  
  By mattjj@google.com from
  https://github.com/renmengye/tensorflow-forward-ad/issues/2
  """
  
  v = tf.zeros_like(ys)
  g = tf.gradients(ys, xs, grad_ys=v)
  return tf.gradients(g, v, grad_ys=d_xs)
```

Source: https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb

When I looking looking up how to recreate the same basic algorithm, I found that the above may not be currently possible in PyTorch as I think it's Forward AD: https://pytorch.org/docs/stable/autograd.html#torch.autograd.functional.jvp

> we don’t have support for forward mode AD in PyTorch at the moment.

Looking at PyTorch's development of the required feature makes it seem like it isn't coming anytime soon: https://github.com/pytorch/pytorch/issues/10223, https://github.com/pytorch/rfcs/pull/11


So, is there a way I can use Captum to calculate how much a neuron contributes to the different labels / classes of a model?

ProGamerGov:
I've made progress on getting the attribution stuff working:

```
def fwd_gradients(ys, xs, d_xs):
    v = torch.zeros_like(ys)
    v.requires_grad = True
    g = torch.autograd.grad(
            outputs=[ys],
            inputs=xs,
            grad_outputs=[v],
            create_graph=True,        
        )[0]

    return torch.autograd.grad(
            outputs=[g],
            inputs=v,
            grad_outputs=d_xs,
            create_graph=True,   
        )[0]
```

NarineK:
Thank you @ProGamerGov! This week has been a bit busy. I will look into your PRs next week or on the weekend.

ProGamerGov:
@NarineK No worries! Atlas attributions may have be added a future PR. I also don't really have a way to display the atlas attribution information either right now (Lucid leaves that to Distill.pub's interactive HTML stuff).

I also used the heatmap function to help visualize the sample counts for each atlas visualization. I'm not sure if there's a better way to display this information right now as it's also left to Distill.pub in Lucid.

ProGamerGov:
Another potential issue is that downloading the ImageNet 2012 training dataset & collecting samples from it for the tutorial notebook could be difficult and time consuming for users. So, it may be wise to host a few pre-collected layer samples somewhere for users to download and use, as the samples are only around 100mb for about 100k samples. Though if we did that, someone might want to collect the full 1 million samples for the sake of atlas visualization quality.

ProGamerGov:
After some testing, I think I can make sample collection faster and more memory efficient. Saving every sample batch using `torch.save` and then loading all the files at once & concatenating them together seems to be a better way of going about things.

ProGamerGov:
Update: 

I resized every image in the ImageNet2012 dataset to 256 and replaced the Pillow library with pillow-simd to speed things up now that memory usage is no longer an issue.

According to tqdm, collecting samples for the entire dataset of 1281167 images for all of the main layers at once took just under 4 and half hours, and averaged 80 images a second  (`[4:23:59<00:00, 80.88 images/s]`). 

The issue now is that I am running out of memory when using UMAP on the resulting samples tensors.

ProGamerGov:
@NarineK As the activation atlas tutorial requires activation samples from the model's training data, I think that we should provide pre-collected activation samples and their UMAP reduced versions for download. 


I'm not sure if we want to host files for multiple layers or just the one though.

For just the Mixed4c layer activation samples, the main file is 2.44GB and the UMAP reduced file is 9.77MB. 
For all the Mixed4.x & Mixed5.x layers, the size of all the sample files is 22.7GB.

ProGamerGov:
@NarineK In order to make the activation atlas match the 2d scatter plot, I use `torch.flip` to flip the horizontal axis. I tried to avoid using this fix, but I was unable to find the right combo of grid parameters to make it happen.

ProGamerGov:
@NarineK I've figured out the first part of the atlas spatial attribution function, but I'm having trouble with the second part. Would you happen to have any ideas on how to get it working?

```
def fwd_gradients(logit_activ, layer_activ, x: int , y: int):
    zeros = torch.nn.Parameter(torch.zeros_like(logit_activ))
    grad_one = torch.autograd.grad(
            outputs=[logit_activ],
            inputs=[layer_activ],
            grad_outputs=[zeros],
            create_graph=True,      
        )[0]
    grad_two = torch.autograd.grad(
            outputs=[grad_one],
            inputs=zeros,
            grad_outputs=[torch.ones_like(grad_one)],
            create_graph=True,
        )[0]
```

```
    # Issue occurs somewhere after here
    # Equivalent tensorflow code is: 
    # logit_attr = grad_two.eval({layer_name: layer_activ, d_previous: layer_zeros})

    layer_zeros = nn.Parameter(torch.zeros_like(target_activ))
    layer_zeros[..., x, y] = layer_activ[..., x, y]

    # Calc logit attribution
    logit_attr = torch.autograd.grad(
            outputs=[layer_activ],
            inputs=grad_two,
            grad_outputs=[layer_zeros],
            create_graph=True,     
        )[0]
    return logit_attr


from captum.optim._utils.models import collect_activations
activation_dict = collect_activations(model, [model.mixed4d, model.fc])
target_activ = activation_dict[model.mixed4d]
logit_activ = activation_dict[model.fc]


attr_tensor = fwd_gradients(logit_activ, target_activ, 5, 5)
```

And the resulting error message:


```
<ipython-input-4-0367a6753650> in fwd_gradients(logit_activ, layer_activ, x, y)
     26             inputs=grad_two,
     27             grad_outputs=[layer_zeros],
---> 28             create_graph=True,
     29         )[0]
     30     return logit_attr

/usr/local/lib/python3.6/dist-packages/torch/autograd/__init__.py in grad(outputs, inputs, grad_outputs, retain_graph, create_graph, only_inputs, allow_unused)
    202     return Variable._execution_engine.run_backward(
    203         outputs, grad_outputs_, retain_graph, create_graph,
--> 204         inputs, allow_unused)
    205 
    206 

RuntimeError: One of the differentiated Tensors appears to not have been used in the graph. Set allow_unused=True if this is the desired behavior.
```

ProGamerGov:
I finally figured out how to do the attributions! The code below will find the class attributions for a particular position in a layer:

```
def find_pos_attr(
    logit_activ: torch.Tensor, layer_activ: torch.Tensor, y: int, x: int
) -> torch.Tensor:
    zeros = torch.nn.Parameter(torch.zeros_like(logit_activ))
    target_zeros = torch.zeros_like(target_activ)
    target_zeros[..., y, x] = target_activ[..., y, x]
    grad_one = torch.autograd.grad(
        outputs=[logit_activ],
        inputs=[layer_activ],
        grad_outputs=[zeros],
        create_graph=True,
    )[0]
    logit_attr = torch.autograd.grad(
        outputs=[grad_one],
        inputs=zeros,
        grad_outputs=[target_zeros],
        create_graph=True,
    )[0]
    return torch.argsort(-logit_attr[0])
```

NarineK:
@ProGamerGov, sorry I haven't had time to look into this PR. I will finish reviewing other optim related PRs and will get to this one as next.

ProGamerGov:
@NarineK No worries! I've just been keeping you (and others) updated on some of the stuff that I've been working on adding to the PR (like attributions for class activation atlases).

I have a quick question though, do you want any attribution specific functions that I create for the Optim module to be added to Captum's attribution module? Or should I leave them in the Optim module, possibly in it's own attribution specific file?

NarineK:
@ProGamerGov , in terms of attribution I think that we should be able to use attribution algorithms from captum.attr package. I haven't looked into your implementation but if you need the grads w.r.t. inputs you can use saliency and other support other sophisticated techniques from captum.attr. package.

ProGamerGov:
@NarineK The attribution code that I wrote for the Optim module is the `find_pos_attr` function:

```
def find_pos_attr(
    logit_activ: torch.Tensor,
    target_activ: torch.Tensor,
    y: Optional[Union[int, torch.Tensor]] = None,
    x: Optional[Union[int, torch.Tensor]] = None,
    position_mask: Optional[torch.Tensor] = None,
) -> torch.Tensor:
    """
    Args:
        logit_activ: Captured activations from the FC / logit layer.
        target_activ: Captured activations from the target layer.
        x (int or torch.Tensor): The width position of the target position.
        y (int or torch.Tensor): The height position of the target position.
        position_mask (torch.Tensor): If using a batch size greater than one, a mask
            is used to zero out all the non-target positions.
    Returns:
        logit_attr (torch.Tensor): A sorted list of class attributions for the target
            spatial positions.
    """

    assert x is not None and y is not None or x is None and y is None
    assert x is None and position_mask is not None or x is not None and position_mask is None

    zeros = torch.nn.Parameter(torch.zeros_like(logit_activ))
    target_zeros = torch.zeros_like(target_activ)

    if position_mask is None:
        if x is not None and y is not None:
            target_zeros[..., y, x] = target_activ[..., y, x]
        else:
            target_zeros = target_activ
    else:
        target_zeros = target_activ * position_mask

    grad_one = torch.autograd.grad(
        outputs=[logit_activ],
        inputs=[target_activ],
        grad_outputs=[zeros],
        create_graph=True,
    )[0]
    logit_attr = torch.autograd.grad(
        outputs=[grad_one],
        inputs=zeros,
        grad_outputs=[target_zeros],
        create_graph=True,
    )[0]

    # Sort attributions - could be moved out of the function
    return torch.argsort(-logit_attr)
```

The function is based on the code used in the companion code for the activation atlas research paper: https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb#scrollTo=sY7TB--AF1Iz

The Lucid code describes what is happening as follows:

>   Forward-mode pushforward analogous to the pullback defined by tf.gradients.
>   With tf.gradients, grad_ys is the vector being pulled back, and here d_xs is
>   the vector being pushed forward.
>   
>   By mattjj@google.com from
>   https://github.com/renmengye/tensorflow-forward-ad/issues/2

  
It's just a small position for getting the class attributions of particular spatial position in the target layer. I tried looking into the attribution code that Captum already has, and I didn't see anything that would let me accomplish this?

NarineK:
@ProGamerGov, I need to look into it -  I think that we can still use existing attribution methods for that.

ProGamerGov:
So this comment is completely unrelated to this PR because I didn't want to make a new Issue for it. It is not related to the activation atlas attributions needed for class activation atlases.

In [the original Optim PR](https://github.com/pytorch/captum/pull/412) @ludwigschubert mentioned using Captum's attr module for Optim related attribution tasks, and I have been trying to figure out how to recreate them in Capum. I've been looking through the attr module's documentation & code, and I am having trouble figuring out which attr module classes / functions will perform the right calculations.

The code below is basically how most of Lucid attribution related code works (except for class activation atlases, and path integrated channel attribution):

```
def get_attributions(
    target_layer_activations: torch.Tensor,  # activations from target layer
    logit_activations: torch.Tensor,  # activations from logit / FC layer
    class1: int = 176,
    class2: Optional[int] = 55,
) -> Union[torch.Tensor, List[torch.Tensor]]:

    # Sometimes the exact way that score is calculated is different than this:
    score = target_layer_activations[:, class1]
    if class2 is not None:
        score = score - logit_activations[:, class2]

    grad = torch.autograd.grad(
        outputs=[score],
        inputs=target_layer_activations,
        grad_outputs=torch.ones_like(score),
    )[0]

    attr = (grad * activations)[0] # channel attribution, we reduce down to channels like so:  attr.sum(1).sum(1)
    # attr = torch.sum(grad * activations, 1)[0] # This is for spatial attributions
    # attr = [torch.sum(group_vec * grad) for group_vec in group_vecs] # This is for neuron group attributions
    return attr
```

I don't have any code for a PR for this code yet, but it's based on [The Building Blocks of Interpretability](https://distill.pub/2018/building-blocks/) research paper that both I and @greentfrapp have had PyTorch users ask us about. The paper describes a lot of the core features of the Lucid library as they were in 2018, so there may be better methods for attribution calculations that I am not aware of.

ProGamerGov:
@NarineK So this comment is completely unrelated to this PR because I didn't want to make a new Issue for it. It is not related to the activation atlas attributions needed for class activation atlases.

In [the original Optim PR](https://github.com/pytorch/captum/pull/412) @ludwigschubert mentioned using Captum's attr module for Optim related attribution tasks, and I have been trying to figure out how to recreate them in Capum. I've been looking through the attr module's documentation & code, and I am having trouble figuring out which attr module classes / functions will perform the right calculations.

The code below is basically how most of Lucid attribution related code works (except for class activation atlases, and path integrated channel attribution):

```
def get_attributions(
    target_layer_activations: torch.Tensor,  # activations from target layer
    logit_activations: torch.Tensor,  # activations from logit / FC layer
    class1: int = 176,
    class2: Optional[int] = 55,
) -> Union[torch.Tensor, List[torch.Tensor]]:

    # Sometimes the exact way that score is calculated is different than this:
    score = target_layer_activations[:, class1]
    if class2 is not None:
        score = score - logit_activations[:, class2]

    grad = torch.autograd.grad(
        outputs=[score],
        inputs=target_layer_activations,
        grad_outputs=torch.ones_like(score),
    )[0]

    attr = (grad * activations)[0] # channel attribution, we reduce down to channels like so:  attr.sum(1).sum(1)
    # attr = torch.sum(grad * activations, 1)[0] # This is for spatial attributions
    # attr = [torch.sum(group_vec * grad) for group_vec in group_vecs] # This is for neuron group attributions
    return attr
```

I don't have any code for a PR for this code yet, but it's based on [The Building Blocks of Interpretability](https://distill.pub/2018/building-blocks/) research paper that both I and @greentfrapp have had PyTorch users ask us about. The paper describes a lot of the core features of the Lucid library as they were in 2018, so there may be better methods for attribution calculations that I am not aware of.

ProGamerGov:
@NarineK So, I figured out that reason why I was struggling to get the non-atlas related attribution stuff to work with Captum's attr module! For some reason moving the final ReLU layers in each InceptionModule to after the `torch.cat` operation allows the model to give the correct attribution results.

I created a small PR here that should fix the issue with the model: https://github.com/pytorch/captum/pull/621

ProGamerGov:
@NarineK I don't think that my attribution code is working correctly. I've checked over all my other class atlas code far too many times to count, and I haven't found an issue yet. So my next best guess is that the attribution calculates may not be 100% correct.

The code below filters out the activation samples based on their attribution strengths ([exactly like Lucid does](https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/class-activation-atlas.ipynb)):

```
def get_top_class_indices(attributions: torch.Tensor, class_idx: int) -> torch.Tensor:
    """
    Args:
      attributions (torch.Tensor): A set of attributions that correspond to a set of
           activation samples.
      class_idx (int): The target class id.
    Returns:
        indices (torch.Tensor): A set of indices where the top idx is the target class.
    """
    assert attributions.dim() == 2
    return torch.tensor(
        [
            i
            for i, attr in enumerate(attributions)
            if class_idx in torch.argsort(-attr)[0:1]
        ],
        device=attributions.device,
    )

class_indices = get_top_class_indices(sample_attributions, 76)  # 982 scuba diver, 76 tiger
filtered_samples = activation_samples[class_indices]
```

And for the tiger class I get atlases that look like this:

tiger_class_atlas

While Lucid get something like this:

lucid_tiger_class_atlas

My class atlas seems to have a little bit of orange and black coloring which makes me think that the issue is with getting the correct attributions for each sample.


The current iteration of my sample attribution code looks like this: 


```
def find_pos_attr(
    target_activ: torch.Tensor,
    logit_activ: torch.Tensor,
    position_mask: torch.Tensor,
) -> torch.Tensor:
    """
    Calculate class attributions for each target xy spatial position (height, width) 
    for each image in the batch.

    Args:
        logit_activ: Captured activations from the FC / logit layer.
        target_activ: Captured activations from the target layer.
        position_mask (torch.Tensor, optional): If using a batch size greater than
        one, a mask is used to zero out all the non-target spatial positions.
    Returns:
        logit_attr (torch.Tensor): A list of class attributions for the target
            spatial positions.
    """

    assert target_activ.dim() == 4
    assert logit_activ.dim() == 2
    assert position_mask.size(0) == target_activ.size(0)

    zeros = torch.nn.Parameter(torch.zeros_like(logit_activ))
    target_zeros = target_activ * position_mask

    grad_one = torch.autograd.grad(
        outputs=[logit_activ],
        inputs=[target_activ],
        grad_outputs=[zeros],
        create_graph=True,
    )
    logit_attr = torch.autograd.grad(
        outputs=grad_one,
        inputs=zeros,
        grad_outputs=[target_zeros],
        create_graph=True,
    )[0]
    return logit_attr


# Example function for basic testing
def attribute_pos(
    tensor: torch.Tensor, model: nn.Module, target: nn.Module, pos: Tuple[int, int]
) -> torch.Tensor:
    activations_dict = optimviz.models.collect_activations(
        model, [target, model.fc], tensor
    )
    target_activ = activations_dict[target]
    logit_activ = activations_dict[model.fc]
    position_mask = torch.zeros_like(target_activ)
    position_mask[:, :, pos[1], pos[0]] = 1
    return find_pos_attr(target_activ, logit_activ, position_mask)

test_image = torch.zeros(1, 3, 224, 224)
xy_position = (5, 7)  # random spatial position
attr_tensor = attribute_pos(test_image, attr_model, attr_model.mixed4d, xy_position)
```

I've looked through all the Captum attribution algorithms and non of them seem similar enough to be used. The closest Captum class looks like `captum.attr.NeuronGradient`, but that gives an output with a shape of the input rather the fc layer. I think that `NeuronGradient` also only uses one `torch.grad.autograd` so it seems unlikely that it could give the output I need. My / Lucid code is supposed to return a tensor of attributions with a shape of the final fc layer ([1, 1008] for Inceptionv1) for the target spatial position.

NarineK:
@ProGamerGov, it looks like you want to compute second order derivative of the output w.r.t. inputs for two intermediate layers?
In captum we don't have an option to compute the grads between two intermediate layers and currently we support first order derivatives only. I will think how to extend it without loss of generality so that we can support those cases too.
Do you mind pointing me out to the place where lucid is computing those derivatives ?

I'm not quite sure about the second gradients:
```
logit_attr = torch.autograd.grad(
        outputs=grad_one,
        inputs=zeros,
        grad_outputs=[target_zeros],
        create_graph=True,
    )[0]
```
Why is the input zeros here ?

ProGamerGov:
@NarineK The Lucid code can be found in their tutorial notebook here (under the activation atlas tutorial list on Lucid's Github page): https://colab.research.google.com/github/tensorflow/lucid/blob/master/notebooks/activation-atlas/activation-atlas-collect.ipynb

The calculations are also mentioned in the research paper under "Aggregating Multiple Images" subheading here: https://distill.pub/2019/activation-atlas/


I am not 100% sure why Lucid uses the zeros input, but when I replicated it I got the correct results. I can ask Chris about it though if you need me to!

I would also check out this Github issue that the Lucid code links to (and refers to what's happening as a trick for making forward-mode JVPs work): https://github.com/renmengye/tensorflow-forward-ad/issues/2

NarineK:
> https://distill.pub/2019/activation-atlas/

Thank you for the reference, @ProGamerGov! According to the definition I see that it is vector-jacobean-product (VJP). 
Screen Shot 2021-03-01 at 1 40 06 PM
Then it should be equivalent to Layer Gradient X Activation?

https://github.com/pytorch/captum/blob/master/captum/attr/_core/layer/layer_gradient_x_activation.py#L22

Do you mind trying it out and seeing if you get the same results ?

Thank you!

ProGamerGov:
@NarineK How do I set Layer Gradient X Activation to target all classes? Because it's given me an error if don't specify a class.

NarineK:
> @NarineK How do I set Layer Gradient X Activation to target all classes? Because it's given me an error if don't specify a class.

No it is for a specific class. If you want for all classes you can e.g. summarize across all class in a custom forward function that you'll pass to attribution.

ProGamerGov:
@NarineK Ah okay! It also looks like the output has the same shape as the target layer and not the final layer / FC / logits layer when I try the model: `torch.Size([1, 1024, 7, 7])`? The output I'm looking for should be `torch.Size([1, 1008])` (or be able to be reduced to that shape), where `1008` is the number of channels in the final layer that computes class values.

ProGamerGov:
@NarineK The attribution code for the activation samples should be calculating the attributions for all classes and not just a single class. Target classes are then filtered out afterwards in various ways when we use the attributions to filter our sample activations.

NarineK:
> final layer / FC / logits layer

> @NarineK The attribution code for the activation samples should be calculating the attributions for all classes and not just a single class. Target classes are then filtered out afterwards in various ways when we use the attributions to filter our sample activations.

@ProGamerGov, if I understand correctly, you want that the attributions have the shape: [1, 1008] where 1 is the batch size and 1008 is the number of classes ?
In that case you can try to compute the attribution for each class separately, summarize it into a scalar and stack them into final attribution tensor for each class. This might not be computationally efficient but it is interesting if we will get the same result.

ProGamerGov:
@NarineK Yes that's correct!

I'll test out doing it separately. We really need this code to be efficient though because it's going to be used on entire image training datasets (through batching).

ProGamerGov:
@NarineK So I tested it by doing each attribution individually and got mostly the same results. The top 25 strongest values were the exact same, but the results differed somewhere after that.


```
layer_ga = attr.LayerGradientXActivation(attr_model, attr_model.mixed5b)

A=[]
for c in range(0, 1008):
    print(c)
    c_attr_tensor = layer_ga.attribute(catdog_image, c).detach()[...,5,5].sum()
    A.append(c_attr_tensor)


attr_tensor_test = torch.as_tensor(A)
print(torch.argsort(-attr_tensor) == torch.argsort(-attr_tensor_test))
```

NarineK:
> @NarineK Yes that's correct!
> 
> I'll test out doing it separately! We really need this code to be efficient though because it's going to be used on entire image training datasets (through batching).

I see, yeah, another way is to expand the input image to [1008 x channel x  height x width] shape and adjust the target vector accordingly but that is going to be time consuming

> @NarineK So I tested it by doing each attribution individually and got mostly the same results. The top 25 strongest values were the exact same, but the results differed somewhere after that.
> 
> ```
> attr_model = googlenet(pretrained=True, replace_relus_with_redirectedrelu=False).to(device).eval()
> 
> layer_ga = attr.LayerGradientXActivation(attr_model, attr_model.mixed5b)
> 
> A=[]
> for c in range(0, 1008):
>     print(c)
>     c_attr_tensor = layer_ga.attribute(catdog_image, c).detach()[...,5,5].sum()
>     A.append(c_attr_tensor)
> 
> 
> attr_tensor_test = torch.as_tensor(A)
> 
> # Using my attribution code that I posted in a comment above:
> attr_tensor = attribute_pos(catdog_image, attr_model, attr_model.mixed5b, (5, 5))
> 
> print(torch.argsort(-attr_tensor )[0:25])
> print(torch.argsort(-attr_tensor_test )[0:25])
> 
> print(torch.argsort(-attr_tensor) == torch.argsort(-attr_tensor_test))
> ```

Thank you for testing it, @ProGamerGov ! yes, that solution is more performant. We could potentially do one forward pass with an expanded input of [1008 x channel x height x width] then aggregate but I think that it is going to increase space complexity.

ProGamerGov:
@NarineK It's interesting that an trick designed for the original TensorFlow 1.x (in 2017) is still seemingly more efficient than what we can do 4 years later. Though eventually it looks like PyTorch will support a far faster and easier way of doing it, but at the moment it's still WIP: https://github.com/pytorch/pytorch/issues/10223

Maybe it might make for faster and more efficient code if we can find a way to implement the forward-mode JVP algorithm in Captum?

NarineK:
I've been thinking about it. I've just noticed that that trick can be very expensive since PT currently doesn't support forward-mode AD.
https://github.com/pytorch/pytorch/issues/10223#issuecomment-547504596
https://github.com/pytorch/pytorch/issues/24004#issuecomment-519940913
It would be interesting to compare the performances and see if the trick is more performance because it says the second backward is taking substantially longer than the first one.

ProGamerGov:
@NarineK So, I asked Chris if he know of any faster & more efficient methods, and he was not aware of any that we could use. He suggests that we stick with the trick for now until PyTorch adds official support for the forward ad feature.

NarineK:
Thanks for checking with Chris! I meant, comparing the performance of the trick with 2 backward passes with the performance of the class level attribution with LayerActivation x Layer Gradients for each class in the loop.

ProGamerGov:
@NarineK Okay, I've run some speed tests for the different methods:

```
def time_test(test_func):
    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)

    start.record()
    test_func()
    end.record()

    torch.cuda.synchronize()
    print(start.elapsed_time(end))
```

Testing the double backward trick:

```
def attribute_pos_test():
    attr_tensor = attribute_pos(catdog_image.clone(), attr_model, attr_model.mixed5b, (5, 5))

for i in range(5):
    time_test(attribute_pos_test)
```

And the timings for the above test:

```
10.443360328674316
10.459136009216309
9.350687980651855
10.356767654418945
10.608672142028809
```


Testing the LayerActivation x Layer Gradients for each class in the loop:

```
def layer_x_gradient():
    attr_tensor= attr.LayerGradientXActivation(attr_model, attr_model.mixed5b)
    c_attr_tensor = torch.tensor([layer_ga.attribute(catdog_image, c).detach()[...,5,5].sum() for c in range(1008)], device=catdog_image.device)


for i in range(5):
    time_test(layer_x_gradient)
```

And the timings for the above test:

```
9482.23828125
9191.9150390625
9105.234375
9142.1015625
9532.21875
```

I also tried the 'batch_size = num_classes' idea and that failed due to out of memory errors.

NarineK:
> @NarineK Okay, I've run some speed tests for the different methods (using a Colab GPU instance):
> 
> ```
> def time_test(test_func):
>     start = torch.cuda.Event(enable_timing=True)
>     end = torch.cuda.Event(enable_timing=True)
> 
>     start.record()
>     test_func()
>     end.record()
> 
>     torch.cuda.synchronize()
>     print(start.elapsed_time(end))
> ```
> 
> Testing the double backward trick:
> 
> ```
> def attribute_pos_test():
>     attr_tensor = attribute_pos(catdog_image.clone(), attr_model, attr_model.mixed5b, (5, 5))
> 
> for i in range(5):
>     time_test(attribute_pos_test)
> ```
> 
> And the timings for the above test:
> 
> ```
> 10.443360328674316
> 10.459136009216309
> 9.350687980651855
> 10.356767654418945
> 10.608672142028809
> ```
> 
> Testing the LayerActivation x Layer Gradients for each class in the loop:
> 
> ```
> def layer_x_gradient():
>     attr_tensor= attr.LayerGradientXActivation(attr_model, attr_model.mixed5b)
>     c_attr_tensor = torch.tensor([layer_ga.attribute(catdog_image, c).detach()[...,5,5].sum() for c in range(1008)], device=catdog_image.device)
> 
> 
> for i in range(5):
>     time_test(layer_x_gradient)
> ```
> 
> And the timings for the above test:
> 
> ```
> 9482.23828125
> 9191.9150390625
> 9105.234375
> 9142.1015625
> 9532.21875
> ```
> 
> The time for a single class was similar to the first test but often slightly higher.
> 
> I also tried the 'batch_size = num_classes' idea and that failed due to out of memory errors.

Thank you for testing it out, @ProGamerGov! It looks like the trick is faster.

ProGamerGov:
@NarineK How can implement this gradient modification in PyTorch?

```
def forward(input: torch.Tensor) -> torch.Tensor:
    x = max_pool(input)
    x.grad = avg_pool_grad(max_pool(x.grad)-max_pool(-x.grad)).grad
```

ProGamerGov:
The algorithm I'm trying to create is an alternative method that Chris suggested for creating a relaxed gradient for the max pooling layers 

In order to calculate the right attributions, I need to replace the max pooling layers with a special max pooling layer that creates a relaxed gradient like so:


```
class MaxPool2dSmoothed(torch.nn.Module):
    """
    Pooling layer where, if we backprop through it,
    gradients get allocated proportional to the input activation.
    Then we backprop through that instead.

    #https://github.com/tensorflow/lucid/blob/master/lucid/optvis/overrides/smoothed_maxpool_grad.py

    Args:
        kernel_size (int or tuple of int): The size of the window to
            perform average pooling with.
        stride (int or tuple of int, optional): The stride window size
            to use.
        padding (int or tuple of int): The amount of
            zero padding to add to both sides in the nn.AvgPool2d module.
        ceil_mode (bool, optional): Whether to use ceil or floor for
            creating the output shape.
    """

    def __init__(
        self,
        kernel_size: Union[int, Tuple[int, ...]],
        stride: Optional[Union[int, Tuple[int, ...]]] = None,
        padding: Union[int, Tuple[int, ...]] = 0,
        ceil_mode: bool = False,
    ) -> None:
        super().__init__()
        self.avgpool = torch.nn.AvgPool2d(
            kernel_size=kernel_size,
            stride=stride,
            padding=padding,
            ceil_mode=ceil_mode,
        )

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x_avg = self.avgpool(x ** 2) / (1e-2 + self.avgpool(torch.abs(x)))

        x_smoothed = torch.autograd.grad(
            outputs=[x_avg],
            inputs=[x],
            grad_outputs=[torch.ones_like(x_avg)],
            create_graph=True,
        )[0]
        return x_smoothed
```

The problem is however that the `torch.autograd.grad` line in the above code causes out of memory errors on GPU devices. So, I'm trying to get an alternative solution working.

ProGamerGov:
@NarineK So I went back I tried Chris' other suggestion of using `stop_grad(maxpool(x)) + avgpool(x) - stop_grad(avgpool(x))`, and I was able to get the attributions working correctly!


This layer is used to replace the existing MaxPool2d layers in the Inception V1 model:

```
class MaxPool2dRelaxed(torch.nn.Module):
    """
    A relaxed pooling layer, that's useful for calculating attributions of spatial
    positions. This layer reduces Noise in the gradient through the use of a
    continuous relaxation of the gradient.

    Args:
        kernel_size (int or tuple of int): The size of the window to perform max &
            average pooling with.
        stride (int or tuple of int, optional): The stride window size to use.
        padding (int or tuple of int): The amount of zero padding to add to both sides
            in the nn.MaxPool2d & nn.AvgPool2d modules.
        ceil_mode (bool, optional): Whether to use ceil or floor for creating the
            output shape.
    """

    def __init__(
        self,
        kernel_size: Union[int, Tuple[int, ...]],
        stride: Optional[Union[int, Tuple[int, ...]]] = None,
        padding: Union[int, Tuple[int, ...]] = 0,
        ceil_mode: bool = False,
    ) -> None:
        super().__init__()
        self.maxpool = torch.nn.MaxPool2d(
            kernel_size=kernel_size, stride=stride, padding=padding, ceil_mode=ceil_mode
        )
        self.avgpool = torch.nn.AvgPool2d(
            kernel_size=kernel_size, stride=stride, padding=padding, ceil_mode=ceil_mode
        )

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        return self.maxpool(x.detach()) + self.avgpool(x) - self.avgpool(x.detach())
```

I tested the resulting samples &attributions with my PyTorch activation atlas code, and I got poor results:

class_atlas_882_top_rgb_attr


Then I went and plugged my samples & attributions into Lucid, and it gave me the correct visualizations:

lucid_pytorch_samples2


So, I think that in addition to the attributions not being completely correct (that's now fixed!), there must some difference between my implementation and Lucid or some sort of bug.

ProGamerGov:
@NarineK I figured out the final remaining issue, and now class activation atlases work correctly!

NarineK:
> @NarineK I figured out the final remaining issue, and now class activation atlases work correctly!

That's awesome! Thank you very much @ProGamerGov ! I haven't had time to look into this. I'll check it out.

ProGamerGov:
@NarineK Awesome! I haven't added the `MaxPool2dRelaxed` code yet as I am still trying to figure out how to properly write tests for it. 

If you want to run the activation atlas tutorials, check the `experimental` directory inside of the directory I linked to on the Google Doc.

Then you can load the appropriate files like so for the Activation Atlas tutorial notebook:

```
# Activation Atlas Tutorial
activation_samples = torch.load("mixed4c_samples_activations_100k.pt").to(device)
print(activation_samples.shape)

sample_attributions = torch.load("mixed4c_samples_attributions_100k.pt").to(device)
print(sample_attributions.shape)
```

And for the Class Activation Atlas tutorial notebook:

```
activation_samples = torch.load("mixed5b_samples_activations_400k.pt").to(device)
print(activation_samples.shape)

sample_attributions = torch.load("mixed5b_samples_attributions_400k.pt").to(device)
print(sample_attributions.shape)
```

There is one more outstanding issue where the `capture_activation_samples` function seems to inexplicably freeze after a random amount of collected samples. There's no apparent memory leak, and I've tried everything that I could think of to solve the problem with no luck. I also just put the attribution code in the same file as the `capture_activation_samples` function until we have the attr module solution ready.

ProGamerGov:
Currently both atlas notebooks crash when this cell is run:

```
from tqdm import tqdm

A = []
num_attempts = 1

for i in tqdm(range(vecs.size(0)), position=0, leave=True):
    attempts, loss_list = [], []
    for a in range(num_attempts):
        img, loss = vis_neuron_direction(
            model, model.mixed5b_relu, vec=vecs[i], vec_whitened=whitend_activation_samples
        )
        attempts.append(img.detach())
        loss_list.append(loss)

    A.append(attempts[torch.argmax(torch.cat(loss_list))])
```

The crash can seemingly be averted by removing the final line and the appends. I have no idea why this issue is happening.

ProGamerGov:
@NarineK So, I just discovered that using the newly implemented batch targeting makes it possible to render an atlas with hundreds of images in a fraction of the time it takes to render each individually. While this is totally awesome, it presents a slight issue with comparing multiple atlas rendering attempts.  Previously I would compare the losses for each image and then take the images with lowest loss value to create the final atlas, but this is not possible if I render each atlas all at once.

Do you have any suggestions for other ways that I could automatically judge the best image for each atlas cell?

NarineK:
> @NarineK So, I just discovered that using the newly implemented batch targeting makes it possible to render an atlas with hundreds of images in a fraction of the time it takes to render each individually! While this is totally awesome, it presents a slight issue with comparing multiple atlas rendering attempts. Previously I would compare the losses for each image and then take the images with lowest loss value to create the final atlas, but this is not possible if I render each atlas all at once.
> 
> Do you have any suggestions for other ways that I could automatically judge the best image for each atlas cell?

@ProGamerGov, do you mean the new batch targeting merged with SK's PR ?  I have to look deeper into the code. I haven't had time to look into this PR. Can you maybe point me out to the code ?

ProGamerGov:
@NarineK Yes, I'm referring to the batch targeting added by SK's PR.

The code I'm referring to exists in both atlas tutorial notebooks, and it renders a stack of direction vectors.

Currently every atlas cell image is rendered individually like this, and thus I can take the final loss value from each rendering to use for comparing multiple attempts:

```
def vis_neuron_direction(
    model: torch.nn.Module,
    target: torch.nn.Module,
    vec: torch.Tensor,
    vec_whitened: Optional[torch.tensor] = None,
) -> torch.Tensor:
    image = opt.images.NaturalImage((80, 80)).to(device)
    loss_fn = opt.loss.AngledNeuronDirection(
        target, vec, vec_whitened, cossim_pow=4.0
    )
    transforms = torch.nn.Sequential(
        torch.nn.ConstantPad2d(2, value=1.0),
        opt.transforms.RandomSpatialJitter(4),
        opt.transforms.RandomSpatialJitter(4),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomScale(
            scale=[0.995 ** n for n in range(-5, 80)]
            + [0.998 ** n for n in 2 * list(range(20, 40))]
        ),
        opt.transforms.RandomRotation(
            degrees=list(range(-20, 20))
            + list(range(-10, 10))
            + list(range(-5, 5))
            + 5 * [0]
        ),
        opt.transforms.RandomSpatialJitter(2),
    )
    obj = opt.InputOptimization(model, loss_fn, image, transforms)
    history = obj.optimize(opt.optimization.n_steps(256, False), lr=0.020)
    return image(), history[-1]


from tqdm import tqdm

A = []
num_attempts = 2

for i in tqdm(range(vecs.size(0)), position=0, leave=True):
    attempts, loss_list = [], []
    for a in range(num_attempts):
        img, loss = vis_neuron_direction(
            model, model.mixed5b_relu, vec=vecs[i], vec_whitened=whitend_activation_samples
        )
        attempts.append(img.detach().cpu())
        loss_list.append(loss)

    # Keep best attempt for each atlas grid cell
    A.append(attempts[torch.argmin(torch.stack(loss_list))])
```

The above code is extremely slow though when you compare it to rendering everything at once with batch targeting:

```
def vis_neuron_direction(
    model: torch.nn.Module,
    target: torch.nn.Module,
    vecs: torch.Tensor,
    vec_whitened: Optional[torch.Tensor] = None,
) -> torch.Tensor:
    image = opt.images.NaturalImage((80, 80), batch=vecs.shape[0]).to(device)
    loss_fn = sum(
        [
            opt.loss.AngledNeuronDirection(
                target, vec, vec_whitened, cossim_pow=4.0, batch_index=i
            )
            for i, vec in enumerate(vecs)
        ]
    )
    transforms = torch.nn.Sequential(
        torch.nn.ConstantPad2d(2, value=1.0),
        opt.transforms.RandomSpatialJitter(4),
        opt.transforms.RandomSpatialJitter(4),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomSpatialJitter(8),
        opt.transforms.RandomScale(
            scale=[0.995 ** n for n in range(-5, 80)]
            + [0.998 ** n for n in 2 * list(range(20, 40))]
        ),
        opt.transforms.RandomRotation(
            degrees=list(range(-20, 20))
            + list(range(-10, 10))
            + list(range(-5, 5))
            + 5 * [0]
        ),
        opt.transforms.RandomSpatialJitter(2),
    )
    obj = opt.InputOptimization(model, loss_fn, image, transforms)
    history = obj.optimize(opt.optimization.n_steps(256, False), lr=0.021)
    return image(), history[-1]


num_attempts = 2
attempts, loss_list = [], []
for a in range(num_attempts):
    imgs, losses = vis_neuron_direction(
        model, model.mixed5b_relu, vec=vecs, vec_whitened=whitend_activation_samples
    )
    attempts.append(imgs.detach().cpu())
    loss_list.append(losses)

best_imgs = # Somehow find the best img for each atlas grid cell.
```

However, I can't determine which image for each cell is the better image based on loss values as the losses for each batch element are merged together.

ProGamerGov:
@NarineK I think that I may have figured out a solution to the problem! 

After some testing I found that if I saved each loss function objective as a list first, then I could use that list to run the output images through the model and objectives again. This essentially recreates the final loss step (I think), but without the transforms.

```
def vis_neuron_direction(
    model: torch.nn.Module,
    target: torch.nn.Module,
    vecs: torch.Tensor,
    vec_whitened: Optional[torch.Tensor] = None,
) -> Tuple[torch.Tensor, torch.Tensor]:
    image = opt.images.NaturalImage((80, 80), batch=vecs.shape[0]).to(device)
    loss_fn_list = [
        opt.loss.AngledNeuronDirection(
            target, vec, vec_whitened, cossim_pow=4.0, batch_index=i
        )
        for i, vec in enumerate(vecs)
    ]

    # This is the objective we pass to InputOptimization
    # But we can still access each individual objective from the list
    loss_fn = sum(loss_fn_list)

    transforms = torch.nn.Sequential(
    # Transforms removed as they aren't important for this example
    )

    obj = opt.InputOptimization(model, loss_fn, image, transforms)
    history = obj.optimize(opt.optimization.n_steps(128, True))

    # Collect the final loss values separately for each batch element
    final_activations = opt.models.collect_activations(model, target, image().detach())
    final_losses = torch.stack([l(final_activations).mean() for l in loss_fn_list])

    return image(), final_losses
```

I can then use the function above to render the full atlas like this:

```
num_attempts = 2

attempts, attempt_losses = [], []
for a in range(num_attempts):
    imgs, losses = vis_neuron_direction(
        model, model.mixed5b_relu, vecs=vecs, vec_whitened=whitend_activation_samples
    )
    attempts.append(imgs.detach().cpu())
    attempt_losses.append(losses)
```

And then sorting the results like this seems work!

```
A = []
attempt_losses = torch.stack(attempt_losses)
for i in range(vecs.shape[0]):
    idx = torch.argmax(attempt_losses[:,i])
    A.append(attempts[idx][i].unsqueeze(0))

#  A is now filled with the best atlas image for each cell
#  and can be used to create the full atlas image
```

ProGamerGov:
To better showcase why multiple rendering attempts can be necessary, I made 21 attempts at rendering the first atlas in the class activation atlas tutorial.  The `torch.argmax` and `torch.argmin` functions were used to determine how close the loss values are to `0`. The closer to zero, the better the visualizations should be. For visualization tasks that have mostly negative loss values, we use `torch.argmax` to get the closest values to zero and `torch.argmin` for the values farthest from zero.

Theses are the best of 21 attempts for each atlas cell (using `torch.argmax`):

class_atlas_982_min_21_attempts

And these ones are the worst of 21 attempts for each atlas cell (using `torch.argmin`):
class_atlas_982_max_21_attempts
 
It's pretty obvious why the first atlas is better for research as it better illustrates the scuba diver class than the second atlas. Hopefully this also clears up why I am doing multiple rendering attempts in the atlas notebooks in the first place!

ProGamerGov:
The `lint_test_py36_pip` and `lint_test_py37_conda` tests are failing because they use nightly builds of PyTorch, and there is a bug that accidentally got introduced a couple of days ago. I posted the issue on the PyTorch GitHub repo here: https://github.com/pytorch/pytorch/issues/57421, but I don't think that there is anything that I can do to resolve the issue in the meantime.

ProGamerGov:
There's a small section at the end of the Class Activation Atlas tutorial, which helps demonstrate one of the ways that you can use the information obtained from an activation atlas. For this section I was using two images from the Lucid servers, but they stopped working yesterday and Chris has no idea when it'll be fixed (he doesn't have access to the server anymore). Luckily I had both images saved on my PC just in case something like this happened, and so I uploaded them to a temporary GitHub link (via https://user-images.githubusercontent.com) so that the tutorial will continue to work as normal.

ProGamerGov:
This PR now implements the main activation atlas tutorial and it's supporting core code. 

Activation atlas sample collection / advanced setup was moved to: https://github.com/pytorch/captum/pull/729

Class Activation Atlases were moved to: https://github.com/pytorch/captum/pull/730

ProGamerGov:
I may close this PR and the class atlas one in favor of newer PRs where I can compress all 95 commits into a single commit, or a small number of commits.

ProGamerGov:
@NarineK I have created a cleaned up version of this PR, and Github has it listed as changing 953 lines (including empty space and formatting, so in practice fewer lines were added). Do you want me to include the main tutorial in the same PR, or should I leave it for the PR with the second tutorial? I think the tutorial adds around 444 lines (including empty / formatting lines, and excluding .ipynb formatting related code).

NarineK:
> @NarineK I have created a cleaned up version of this PR, and Github has it listed as changing 953 lines (including empty space and formatting, so in practice fewer lines were added). Do you want me to include the main tutorial in the same PR, or should I leave it for the PR with the second tutorial? I think the tutorial adds around 444 lines (including empty / formatting lines, and excluding .ipynb formatting related code).

@ProGamerGov, thank you for the update. I think if the main is 953 and the tutorial adds 444, it's fine to have the tutorial in that same PR.  So the total LOC will be 953 + 444 for that PR ? What's the id of that new PR ?

ProGamerGov:
@NarineK The new PR can be found here: https://github.com/pytorch/captum/pull/782

Issue #491, opened 2020-10-13
non-leaf tensor warning in some attribution algorithms


hal-314:
## 🐛 Bug

I get the following warning when using the Saliency or InputXGradient attribution method but not with IntegratedGradients or GradientShap:
```
UserWarning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the gradient for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations.
  warnings.warn("The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad "
```
## To Reproduce

Execute this:
```python
import warnings
import torch
import torch.nn as nn

from captum.attr import Saliency, IntegratedGradients, InputXGradient
from captum.attr import configure_interpretable_embedding_layer

class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.embedding = nn.Embedding(10,3)
        self.fc = nn.Linear(5+3, 2)

    def forward(self, x_cat, x):
        x_cat = self.embedding(x_cat)
        x = torch.cat([x, x_cat], dim=1)
        x = self.fc(x)
        return x

def sal(model, X, node_index):
    sal = Saliency(model)
    #sal = IntegratedGradients(model)
    #sal = InputXGradient(model)
    grads = sal.attribute(X, target=node_index)
    return grads


net = Net()

X_cont = torch.rand(1, 5)
X_cat = torch.randint(0,9, (1,))

X_cont.requires_grad = True

net(X_cat, X_cont)

#with torch.no_grad(): # <- Uncomment to remove warnings. 
X_cat_emb = net.embedding(X_cat)

X_cat_emb.requires_grad_()

with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    emb_net = configure_interpretable_embedding_layer(net)

# It's the same if we compute embeddings from InterpretableEmbeddingBase like Bert tutorial
#X_cat_emb2 = emb_net.indices_to_embeddings(X_cat)
#X_cat_emb2.requires_grad_()


attr = sal(net, (X_cat_emb, X_cont), 0)
#attr = sal(net, (X_cat_emb2, X_cont), 0)

```

## Expected behavior

[...]

PR #597, opened 2021-01-25
Leaf Warning Fix

vivekmig:
This removes the resetting of grad attribute to zero, which is causing warnings as mentioned in #491 and #421 . Based on torch [documentation](https://pytorch.org/docs/stable/autograd.html#torch.autograd.grad), resetting of grad is only needed when using torch.autograd.backward, which accumulates results into the grad attribute for leaf nodes. Since we only utilize torch.autograd.grad (with only_inputs always set to True), the gradients obtained in Captum are never actually accumulated into grad attributes, so resetting the attribute is not actually necessary.

This also adds a test to confirm that the grad attribute is not altered when gradients are utilized through Saliency.

### DIFF
--- a/tests/attr/test_saliency.py
+++ b/tests/attr/test_saliency.py
@@ -136,6 +136,13 @@ class Test(BaseTest):
     def test_saliency_classification_vargrad(self) -> None:
         self._saliency_classification_assert(nt_type="vargrad")
 
+    def test_saliency_grad_unchanged(self) -> None:
+        model, inp, grads, add_args = _get_basic_config()
+        inp.grad = torch.randn_like(inp)
+        grad = inp.grad.detach().clone()
+        self._saliency_base_assert(model, inp, grads, add_args)
+        assertTensorTuplesAlmostEqual(self, inp.grad, grad)
+
     def _saliency_base_assert(
         self,
         model: Module,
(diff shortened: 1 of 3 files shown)


tests/attr/test_saliency.py
```diff
@@ -136,6 +136,13 @@ def test_saliency_classification_smoothgrad(self) -> None:
     def test_saliency_classification_vargrad(self) -> None:
         self._saliency_classification_assert(nt_type="vargrad")
 
+    def test_saliency_grad_unchanged(self) -> None:
+        model, inp, grads, add_args = _get_basic_config()
+        inp.grad = torch.randn_like(inp)
+        grad = inp.grad.detach().clone()
+        self._saliency_base_assert(model, inp, grads, add_args)
+        assertTensorTuplesAlmostEqual(self, inp.grad, grad)
```
NarineK, line 144: nit: Can we do exact equal here to be sure that the grads didn't change ?
It would be interesting to use backward and compare with the grads computed with the backward to contrast the difference.

vivekmig: Sure, updated with delta=0.0 for exact equality. From the docs, I think backward should always compute the same gradients just also updating the grad attribute for leaf nodes, but can double-check this.

NarineK: Thank you! I was thinking that if we use backward instead autograd.grad then we will see the difference, right ?

vivekmig: Makes sense, yup, if we used backward instead, this test should fail! Although backward doesn't take inputs as parameters and just requires accessing grad for the gradients, so would be a different structure from what we currently use and might fail for non-leaf inputs.

Issue #595, opened 2021-01-25
An error about captum.metric


bachml:
Hi all,

When I run the code from the API Reference  (https://captum.ai/api/metrics.html#infidelity):

net = model. # model is a resnet18 trained on CIFAR10
saliency = Saliency(net)
input = torch.randn(2, 3, 32, 32, requires_grad=True)
sens = sensitivity_max(saliency.attribute, input)

Then,  I got this error

AssertionError                            Traceback (most recent call last)
<ipython-input-7-be7018aa3b15> in <module>
      2 saliency = Saliency(net)
      3 input = torch.randn(2, 3, 32, 32, requires_grad=True)
----> 4 sens = sensitivity_max(saliency.attribute, input)

~/.local/lib/python3.6/site-packages/captum/metrics/_core/sensitivity.py in sensitivity_max(explanation_func, inputs, perturb_func, perturb_radius, n_perturb_samples, norm_ord, max_examples_per_batch, **kwargs)
    306 
    307     with torch.no_grad():
--> 308         expl_inputs = explanation_func(inputs, **kwargs)
    309         metrics_max = _divide_and_aggregate_metrics(
    310             cast(Tuple[Tensor, ...], inputs),

~/.local/lib/python3.6/site-packages/captum/log/__init__.py in wrapper(*args, **kwargs)
     33             @wraps(func)
     34             def wrapper(*args, **kwargs):
---> 35                 return func(*args, **kwargs)
     36 
     37             return wrapper

~/.local/lib/python3.6/site-packages/captum/attr/_core/saliency.py in attribute(self, inputs, target, abs, additional_forward_args)
    130         # They are being formated in the `_run_forward` function in `common.py`
    131         gradients = self.gradient_func(
--> 132             self.forward_func, inputs, target, additional_forward_args
    133         )
    134         if abs:

~/.local/lib/python3.6/site-packages/captum/_utils/gradient.py in compute_gradients(forward_fn, inputs, target_ind, additional_forward_args)
    118         outputs = _run_forward(forward_fn, inputs, target_ind, additional_forward_args)
    119         assert outputs[0].numel() == 1, (
--> 120             "Target not provided when necessary, cannot"
    121             " take gradient with respect to multiple outputs."
    122         )

AssertionError: Target not provided when necessary, cannot take gradient with respect to multiple outputs.

===

Seems like the official code doesn't work for me. Anyone meet similar issues? Thanks

vivekmig:
[...]

PR #598, opened 2021-01-25
Update Sensitivity Documentation

vivekmig:
Addresses #595 , adding target to documentation for sensitivity.


NarineK (approved):
Thank you for the fix! It looks like there is an issue with GPU tests. It popped up since yesterday,

bilalsal (approved):

Issue #547, opened 2020-12-02
Questions about the IMDB sentiment analysis tutorial


Jethro85:
## ❓ Questions and Help
Hello,

I'm new to captum. 
I download the IMDB sentiment analysis tutorial and run it in jupyter notebook. I download the model and the Glove according to the tutorial and run the codes step by step. However, I get different results:

![Screenshot from 2020-12-01 18-55-52](https://user-images.githubusercontent.com/52518717/100811023-b02bad80-3407-11eb-89a9-580418995e01.png)

As you can see, the prediction results for the fourth and sixth sentences are totally different from the results shown by the turtorial. 
Are there any errors in my operation?

NarineK:
Hi @Jethro85, thank you for the question! We had a similar question asked here:
https://github.com/pytorch/captum/issues/399
It could be different versions of IMDB dataset, perhaps also related to different versions of torchtext. What is the vocabulary size in your example ?

Jethro85:
> Hi @Jethro85, thank you for the question! We had a similar question asked here:
> #399
> It could be different versions of IMDB dataset, perhaps also related to different versions of torchtext. What is the vocabulary size in your example ? What version of torchtext do you use ?

I see. My torchtext version is 0.6.0, and the vocabulary size is 101513. Maybe I have to train a new model.

NarineK:
Thank you @Jethro85, I tested it locally and I also see vocabulary Size:  101513 and the same result as you see. However as you can see in the github version the vocabulary Size is  101982:
https://github.com/pytorch/captum/blob/master/tutorials/IMDB_TorchText_Interpret.ipynb
I tried lower versions of torchtext, such as 0.5.0 and 0.4.0 and I see the same results as you do.
It can be that the version of spacy has changed that caused different number of vocabulary size.

Jethro85:
> Thank you @Jethro85, I tested it locally and I also see vocabulary Size: 101513 and the same result as you see. However as you can see in the github version the vocabulary Size is 101982:
> https://github.com/pytorch/captum/blob/master/tutorials/IMDB_TorchText_Interpret.ipynb
> I tried lower versions of torchtext, such as 0.5.0 and 0.4.0 and I see the same results as you do.
> It can be that the version of spacy has changed that caused different number of vocabulary size.
> 
> I unfortunately do not remember what version of spacy I used for training the model.

Ok. Thanks for your reply. I'll train a new model.

NarineK:
[...]

Issue #399, opened 2020-06-08
Unproduce the results of text classification tutorial


wangyongjie-ntu:
I download the Jupiter file of https://captum.ai/tutorials/IMDB_TorchText_Interpret and run step by step.  I use the same training set, the model officially provided. But gives different results.  Can you help explain why? What should I do to fix it? Thanks. By the way, the prediction is also different? Is there any random seed in building vocab? 

![image](https://user-images.githubusercontent.com/12019443/83987824-12818500-a974-11ea-854f-bc5157563197.png)

wangyongjie-ntu:
There seems a bug in this tutorial. As the train-test split is random, the training set used to build vocab is inconsistent. says the indices of input sentences are changing.  You should keep the random seeds consistent between training and testing.

wangyongjie-ntu:
> Hi @wangyongjie-ntu ! I originally trained the model long time ago. Most probably it used the default random seed. If you use GPU then you might also need to fix the seed for GPU too.

Thanks for your reply. I find PyTorch currently supports the save/load of TEXT field object.  To make the tutorial consistent to yours, one option is the save your prebuilt vocab with 'torch.save', and load it in the tutorial with 'torch.load', like loading the pre-trained weights.

NarineK:
Hi @wangyongjie-ntu, I looked into this again and I see that the dataset that I downloaded `aclImdb` already contains train / test split.  IMDB dataset doesn't seem to be splitting randomly: https://github.com/pytorch/text/blob/4b32cf52768baa73a5d59e11451bb580a8dcf192/torchtext/data/dataset.py#L54

wangyongjie-ntu:
Hi @NarineK, It seems you are right. But I find the indices of words can change, I have not understand why this happens. And I save/load the TEXT and solve it.

NarineK:
It's interesting, the vocab size used to be: Vocabulary Size:  101982 and now I see Vocabulary Size:  101867.
Perhaps different versions of torchtext ?

PR #601, opened 2021-01-27
Retrain and update IMDB sentiment analysis model

jessijzhao:
* changed number of embedding dimensions from 50 to 100
* retrained model in models/imdb-model-cnn.pt
* ran IMDB_TorchText_Interpret.ipynb with new model and current library versions
* updated img/sentiment_analysis.png with image of results
* addresses issues #399 and #547


NarineK (approved):
Thank you for working on this PR, @jessijzhao! 
A question: It looks like the size of the model decreased by ~2x. Is that because we store 2x less parameters or did the old version store more than the parameters ?

jessijzhao:
Thanks for the comment @NarineK! The model I trained has 2,620,801 parameters (matching the original training script by bentrevett) and the old model has 5,159,701. I followed the original script by keeping 
MAX_VOCAB_SIZE = 25_000 whereas it looks like the old model was using all 101982 words - should I retrain the model with the full vocabulary?

NarineK:
> Thanks for the comment @NarineK! The model I trained has 2,620,801 parameters (matching the original training script by bentrevett) and the old model has 5,159,701. I followed the original script by keeping
> MAX_VOCAB_SIZE = 25_000 whereas it looks like the old model was using all 101982 words - should I retrain the model with the full vocabulary?

Thank you for verifying it, @jessijzhao! If it is not too time consuming, let's try full vocab size. We might retain better performance results.

PR #602, opened 2021-01-28
Neuron Conductance Tuple Support

vivekmig:
This adds support for Neuron Conductance to allow a callable for selector function, as long as this function is used for selection rather than aggregation. This functionality allows neuron conductance to work with layers which may have input or output as a tuple of tensors.

### DIFF
--- a/captum/attr/_core/neuron/neuron_conductance.py
+++ b/captum/attr/_core/neuron/neuron_conductance.py
@@ -133,7 +133,7 @@ class NeuronConductance(NeuronAttribution, GradientAttribution):
                         - a callable, which should
                           take the target layer as input (single tensor or tuple
                           if multiple tensors are in layer) and return a selected
-                          neuron (output shape should be 1D with length equal to
+                          neuron - output shape should be 1D with length equal to
                           batch_size (one scalar per input example)
 
                           NOTE: Callables applicable for neuron conductance are
--- a/captum/attr/_core/neuron/neuron_conductance.py
+++ b/captum/attr/_core/neuron/neuron_conductance.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import warnings
 from typing import Any, Callable, List, Tuple, Union
 
 import torch
@@ -11,7 +12,7 @@ from captum._utils.common import (
     _format_additional_forward_args,
     _format_output,
     _is_tuple,
-    _verify_select_column,
+    _verify_select_neuron,
 )
 from captum._utils.gradient import compute_layer_gradients_and_eval
 from captum._utils.typing import BaselineType, TargetType, TensorOrTupleOfTensorsGeneric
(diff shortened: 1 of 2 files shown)


tests/attr/neuron/test_neuron_conductance.py
```diff
@@ -93,6 +100,15 @@ def test_simple_conductance_multi_input_batch_relu(self) -> None:
...
         )
 
+    def test_layer_tuple_selector_fn(self) -> None:
+        net = BasicModel_MultiLayer(multi_input_module=True)
+        inp = torch.tensor([[0.0, 6.0, 0.0]])
+        self._conductance_input_test_assert(
+            net, net.multi_relu, inp, lambda x: x[0][:, 1], [0.0, 6.0, 0.0]
+        )
+        lc = LayerConductance(net, net.multi_relu)
+        print(lc.attribute(inp, target=0))
```
NarineK, line 110: do we need this printout ? Did you wanted to assert with the results from Layer Conductance ?

NarineK (approved):
Thank you for woking on this PR, @vivekmig!
I left a small nit.

tests/attr/neuron/test_neuron_conductance.py
```diff
@@ -43,6 +43,13 @@ def test_simple_conductance_input_linear1(self) -> None:
         inp = torch.tensor([[0.0, 100.0, 0.0]])
         self._conductance_input_test_assert(net, net.linear1, inp, 0, [0.0, 90.0, 0.0])
 
+    def test_simple_conductance_input_linear1_selector_fn(self) -> None:
+        net = BasicModel_MultiLayer()
+        inp = torch.tensor([[0.0, 100.0, 0.0]])
+        self._conductance_input_test_assert(
+            net, net.linear1, inp, lambda x: x[:, 0], [0.0, 90.0, 0.0]
+        )
+
```
NarineK, line 52: If there is a test model that has a custom layer that returns a tuple, could we, please, add a test for that too.

tests/attr/neuron/test_neuron_conductance.py
vivekmig: Whoops, good catch thanks! Was initially going to do that, but then made it a separate test, forgot to remove this.

tests/attr/neuron/test_neuron_conductance.py
vivekmig: Yes, I think both the other 2 added tests cover this case. Passing multi_input_module=True uses a custom layer MultiRelu, which returns a tuple.

captum/attr/_core/neuron/neuron_conductance.py
```diff
@@ -113,13 +114,38 @@ def attribute(
...
+                          tuple must be one less than the number of dimensions
+                          in the input / output of the given layer (since
+                          dimension 0 corresponds to number of examples).
+                          This can be used as long as the layer input / output
+                          is a single tensor.
+
+                        - a callable, which should
+                          take the target layer as input (single tensor or tuple
+                          if multiple tensors are in layer) and return a selected
+                          neuron (output shape should be 1D with length equal to
```
miguelmartin75, line 136: mismatched parens here

PR #606, opened 2021-01-30
Optim-wip: Fix _check_layer_in_model test

ProGamerGov:
The function now checks if the check is true then returns, as per: https://github.com/pytorch/captum/pull/574#issuecomment-767780288


NarineK (approved):
Thank you for the fix, @ProGamerGov !

PR #618, opened 2021-02-16
Fix line-break in docstring (visualize_image_attr)

TGabor:
Thanks a lot for providing captum!

Here is a small docstring fix.


NarineK (approved):
LGTM! Thank you!

Issue #581, opened 2021-01-08
Progressbar for time-intensive approaches


lukasfolle:
## 🚀 Feature

Progressbar that indicates the time left to progress given input image (could be tqdm).
## Motivation

When using 3D data or large images, methods like `Occlusion` take a considerable amount of time to finish. Knowing the time left would help impatient users like me 😄

bilalsal:
Hi @lukasfolle,

This is a thoughtful suggestion!
yes, we will consider possibilities to provide progress signals, either using `tqdm` or via an optional hook. 

@edward-io, @vivekmig, @miguelmartin75: what would you recommend?

aobo-y:
@lukasfolle the option for displaying the progress bar has been implemented. We will close this issue.
Please refer to the parameter `show_progress` in the API doc for details https://captum.ai/api/occlusion.html

lukasfolle:
Awesome! Thanks.

PR #630, opened 2021-03-10
add progress & apply to feature ablation, shapley, & lime based methods

aobo-y:
Issue https://github.com/pytorch/captum/issues/581
- implement `progress(...)`. By default, it try to load `tqdm`. If unavailable, fallback to a naive local implementation
- adopt it to feature ablation based methods, `FeatureAblation`, `FeaturePermutation`, `Occlution`
- add corresponding tests. Use `unitest.mock` to mock `stderr` to test the progress output.
- some unrelated auto formattings after running `black .` and `flake8 .`

Sample fallback progress output
Screen Shot 2021-03-10 at 3 58 14 PM

Sample tqdm progress output
Screen Shot 2021-03-10 at 3 58 02 PM


NarineK (commented):
Looks great! Thank you for working on this @aobo-y !
I left couple comments.

Do you mind running type hint checker and fix related issues: ./scripts/run_mypy.sh
Also, running `isort .` for sorting the imports - just to check that everything is okay.

cc: @vivekmig, @miguelmartin75 , @bilalsal

captum/attr/_core/feature_ablation.py
```diff
@@ -307,21 +315,36 @@ def attribute(
                 # Skip any empty input tensors
                 if torch.numel(inputs[i]) == 0:
                     continue
-                for (
-                    current_inputs,
-                    current_add_args,
-                    current_target,
-                    current_mask,
-                ) in self._ablation_generator(
+
```
vivekmig, line 318: One suggestion here, could we instead add an instance method as something like `get_total_features` or `get_total_evaluations`, which would provide the total number of features / evaluations necessary?
This would allow us to have a single progress bar, rather than separate ones for each input, which feels more natural (especially in cases where each feature is a separate input tensor) and would also be consistent with Shapley Value Sampling, Lime, KernelShap, etc., where there is a fixed number of total perturbations / samples and iteration is not dependent on by input index.

In feature ablation, this would just need to find the min / max of each mask (or number of elements if unavailable). This would require an overriding implementation of this method in Occlusion, but this should be pretty simple, e.g. something like:
```
shift_counts = kwargs["shift_counts"] 
total_evals = sum(math.ceil(np.prod(count) / perturbations_per_eval) for count in shift_counts)
```

We would just need to use tqdm with manual control rather than wrapping the generator, so calling update within the inner loop would be sufficient. This should also avoid needing any changes to the generator. What do you think of this approach @aobo-y and @NarineK ?

vivekmig (commented):
Thanks for the great work on this @aobo-y ! Just one suggestion on the progress bar per input.

captum/attr/_core/feature_ablation.py
aobo-y: @vivekmig Yes, this is a very reasonable suggestion and I also considered it before. 
It is definitely feasible, but such function `get_total_evaluations ` will duplicate some logic of the following sections:
https://github.com/pytorch/captum/blob/master/captum/attr/_core/feature_ablation.py#L382-L393
https://github.com/pytorch/captum/blob/master/captum/attr/_core/feature_ablation.py#L504-L516

Duplication may increase the cost of future maintenance that people has to realize there are two places to synchronize ( `get_total_evaluations` will be easy to ignore since it is a trivial util only for progress). For example, how to get the min/max features when mask is None must always have the same behaviors in `get_total_evaluations` and `ablation_generator`. When we update anything about the generator, we have to verify if the progress calculation is still correct.

But I know this concern is a little over-engineering, since after all, `get_total_evaluations` is basically as simple as
```py
input_mask.max() - input_mask.min() if input_mask is not None else input.size(-1)
```
Feel free to let me know your taste. I would love to change it to one progress.

vivekmig: Makes sense! Yes, I agree, we would need to ensure that this total evaluation counter remains in sync with the generator, but I think this trade-off is likely worthwhile to have a single progress-bar, which will be a consistent user experience over all perturbation-based methods. Like your example, duplicate code isn't a big issue here, since this new function shouldn't be more than 2-3 lines. We also likely don't need the kwarg splitting logic in Lines 382-390 and a single loop within the method should be sufficient.

For now, let's just add a comment that this should be verified if the generator is updated. To be safer, we could also add a test for this method that confirms that the total number of forward passes made by the function matches the result of get_total_evaluations, but I don't think that's really necessary for now.

vivekmig (approved):
Looks great, thanks for the awesome work on this :) ! 

As @NarineK mentioned, it would be great to also add this for option for other perturbation based methods (Shapley Value Sampling, Lime, etc.), but we can do that later in a separate diff as well.

aobo-y:
hi @vivekmig @NarineK 

two additional changes as suggested
- support `show_progress` in shapley and lime based algorithms
- count the initial eval of feature ablation & shapley into progress

PR #633, opened 2021-03-15
Fix for Layer Conductance OOM issues

vivekmig:
Layer Conductance currently leads to OOM issues despite using internal_batch_size, suggesting that memory used for intermediate batches is not deallocated appropriately.

It seems that these attribution results were still attached to the corresponding compute graph, which inhibits deallocation of the memory used. Explicitly detaching attribution results when aggregating internal batches resolves this issue.

Confirmed that provided example below led to OOM's prior to this fix and seems to be resolved with this fix.

```
class ToyModel(nn.Module):
    def __init__(self):
        N = 1000000
        super().__init__()
        self.lin1 = nn.Linear(3, N)
        self.relu = nn.ReLU()
        self.lin2 = nn.Linear(N, 2)

    def forward(self, input):
        return self.lin2(self.relu(self.lin1(input)))

input = torch.rand(2, 3)
toymodel = ToyModel()
toymodel.to('cuda:1')

# Runs fine for 500 steps.
att_lc = LayerConductance(toymodel.forward, toymodel.lin1).attribute(
    input.to('cuda:1'),
    target=0,
    n_steps=10000,
    internal_batch_size=100,
)

```

### DIFF
--- a/captum/attr/_utils/batching.py
+++ b/captum/attr/_utils/batching.py
@@ -84,10 +84,10 @@ def _batch_attribution(
             total_attr = current_attr
         else:
             if isinstance(total_attr, Tensor):
-                total_attr = total_attr + current_attr
+                total_attr = total_attr + current_attr.detach()
             else:
                 total_attr = tuple(
-                    current + prev_total
+                    current.detach() + prev_total
                     for current, prev_total in zip(current_attr, total_attr)
                 )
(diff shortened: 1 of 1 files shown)


bilalsal:
Great work, @vivekmig! Thank you for taking initiative.

PR #635, opened 2021-03-16
Remove tests from Captum distributions

vivekmig:
Update package exclusion to include tests. Previously tests were included in Captum distributions unnecessarily, outside of the captum package, causing conflicts with other packages.


NarineK (approved):
LGTM! Thank you for the fix!

Issue #613, opened 2021-02-08
A question about the captum.metrics.infidelity


bachml:
As shown in the example of API reference:

>>> net = ImageClassifier()
>>> saliency = Saliency(net)
>>> input = torch.randn(2, 3, 32, 32, requires_grad=True)
>>>  Computes saliency maps for class 3.
>>> attribution = saliency.attribute(input, target=3)
>>>  define a perturbation function for the input
>>> def perturb_fn(inputs):
>>>    noise = torch.tensor(np.random.normal(0, 0.003, inputs.shape)).float()
>>>    return noise, inputs - noise
>>>  Computes infidelity score for saliency maps
>>> infid = infidelity(net, perturb_fn, input, attribution)

As you can see, if we multiply attribution with a constant, such like attribution = saliency.attribute(input, target=3)*100, then we will have larger infidelity results.

This means the result of infidelity is sensitive to the norm of the attribution map.

But I retry the experiment by the author's implementation(the author who proposed this metric in NIPS19, https://github.com/chihkuanyeh/saliency_evaluation), seems like the author's infidelity implementation is robust to the norm of the attribution map. 

So captum's implementation is different from the author's implementation?

NarineK:
Thank you for the question, @bachml! I think the reason is because in their implementation the authors of the paper scale the explanation sum by a factor as you can see here:
https://github.com/chihkuanyeh/saliency_evaluation/blob/44a66e2531f30b803be3bf5b0786971b7e7f72a1/infid_sen_utils.py#L296 
That scaling factor isn't included in the original paper that's why we haven't included it in our implementation: https://arxiv.org/pdf/1901.09392.pdf
We could potentially perform that type of scaling based on an input flag as well.

NarineK:
Closing, since it got addressed.

PR #639, opened 2021-03-24
support normalize in metric infidelity

aobo-y:
https://github.com/pytorch/captum/issues/613

Support normalizing the infidelity like the author's implementation https://github.com/chihkuanyeh/saliency_evaluation/blob/44a66e2531f30b803be3bf5b0786971b7e7f72a1/infid_sen_utils.py#L295

### DIFF
--- a/captum/metrics/_core/infidelity.py
+++ b/captum/metrics/_core/infidelity.py
@@ -347,11 +347,16 @@ def infidelity(
 
                 Default: None
         normalize (bool, optional): Normalize the dot product of the input
-                perturbation and the attribution so the infedelity value is invariant
+                perturbation and the attribution so the infidelity value is invariant
                 to constant scaling of the attribution values. The normalization factor
-                is defined as the ratio of two mean values across all perturbations:
-                `mean(dot product * func value diff) / mean(dot product * dot product)`.
-                Details can be found in the original paper's implementation
+                beta is defined as the ratio of two mean values:
+                $$ \beta = \frac{
+                    \mathbb{E}_{I \sim \mu_I} [ I^T \Phi(f, x) (f(x) - f(x - I)) ]
+                }{
+                    \mathbb{E}_{I \sim \mu_I} [ (I^T \Phi(f, x))^2 ]
+                } $$.
+                Please refer the original paper for the meaning of the symbols. Same
+                normalization can be found in the paper's official implementation
                 https://github.com/chihkuanyeh/saliency_evaluation
 
                 Default: False
@@ -449,7 +454,7 @@ def infidelity(
 
     def _next_infidelity_tensors(
         current_n_perturb_samples: int,
-    ) -> Tuple[Tensor, Tensor]:
+    ) -> Tuple[Tensor, Tensor, Tensor]:
         perturbations, inputs_perturbed = _generate_perturbations(
             current_n_perturb_samples
         )
@@ -503,17 +508,21 @@ def infidelity(
         )
         attr_times_perturb_sums = cast(Tensor, attr_times_perturb_sums)
 
-        # in order to normalize, we have to keep all the results
-        # return as Tensor(bsz, current_n_perturb_samples)
+        # reshape as Tensor(bsz, current_n_perturb_samples)
+        attr_times_perturb_sums = attr_times_perturb_sums.view(bsz, -1)
+        perturbed_fwd_diffs = perturbed_fwd_diffs.view(bsz, -1)
+
+        # in order to normalize, we have to aggregate the following tensors
+        # to calculate MSE in its polynomial expansion:
+        # (a-b)^2 = a^2 - 2ab + b^2
         return (
-            attr_times_perturb_sums.view(bsz, -1),
-            perturbed_fwd_diffs.view(bsz, -1),
+            attr_times_perturb_sums.pow(2).sum(-1),
+            (attr_times_perturb_sums * perturbed_fwd_diffs).sum(-1),
+            perturbed_fwd_diffs.pow(2).sum(-1),
         )
 
-    def _concat_infidelity_tensors(agg_tensors, tensors):
-        return tuple(
-            torch.cat([agg_t, t], dim=1) for agg_t, t in zip(agg_tensors, tensors)
-        )
+    def _sum_infidelity_tensors(agg_tensors, tensors):
+        return tuple(agg_t + t for agg_t, t in zip(agg_tensors, tensors))
 
     # perform argument formattings
     inputs = _format_input(inputs)  # type: ignore
@@ -537,28 +546,28 @@ def infidelity(
 
     bsz = inputs[0].size(0)
     with torch.no_grad():
-        metrics_sum = _divide_and_aggregate_metrics(
+        # aggregated MSE's polynomial expansion tensors (a^2, ab, b^2)
+        agg_tensors = _divide_and_aggregate_metrics(
             cast(Tuple[Tensor, ...], inputs),
             n_perturb_samples,
             _next_infidelity_tensors,
-            agg_func=_concat_infidelity_tensors,
+            agg_func=_sum_infidelity_tensors,
             max_examples_per_batch=max_examples_per_batch,
         )
 
-    attr_times_perturb_sums, perturbed_fwd_diffs = metrics_sum
-
-    assert attr_times_perturb_sums.size(-1) == n_perturb_samples
-    assert perturbed_fwd_diffs.size(-1) == n_perturb_samples
-
     if normalize:
-        beta_num = (attr_times_perturb_sums * perturbed_fwd_diffs).mean(-1)
-        beta_denorm = attr_times_perturb_sums.pow(2).mean(-1)
+        beta_num = agg_tensors[1]
+        beta_denorm = agg_tensors[0]
 
         beta_denorm[beta_denorm == 0] += 1e-10  # safe divide
         beta = beta_num / beta_denorm
 
-        attr_times_perturb_sums *= beta.unsqueeze(-1)
+        infidelity_values = (
+            beta ** 2 * agg_tensors[0] - 2 * beta * agg_tensors[1] + agg_tensors[2]
+        )
+    else:
+        infidelity_values = agg_tensors[0] - 2 * agg_tensors[1] + agg_tensors[2]
 
-    infidelity_values = (attr_times_perturb_sums - perturbed_fwd_diffs).pow(2).mean(-1)
+    infidelity_values /= n_perturb_samples
 
     return infidelity_values
--- a/captum/metrics/_core/infidelity.py
+++ b/captum/metrics/_core/infidelity.py
@@ -498,10 +498,8 @@ def infidelity(
         )
 
         attr_times_perturb_sums = sum(
-            [
-                torch.sum(attribution_times_perturb, dim=1)
-                for attribution_times_perturb in attributions_times_perturb
-            ]
+            torch.sum(attribution_times_perturb, dim=1)
+            for attribution_times_perturb in attributions_times_perturb
         )
         attr_times_perturb_sums = cast(Tensor, attr_times_perturb_sums)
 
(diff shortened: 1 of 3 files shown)


captum/metrics/_core/infidelity.py
```diff
@@ -345,6 +346,15 @@ def infidelity(
                 `input batch size * n_perturb_samples`.
 
                 Default: None
+        normalize (bool, optional): Normalize the dot product of the input
+                perturbation and the attribution so the infedelity value is invariant
```
vivekmig, line 350: nit: infidelity

captum/metrics/_core/infidelity.py
```diff
@@ -345,6 +346,15 @@ def infidelity(
...
+                to constant scaling of the attribution values. The normalization factor
+                is defined as the ratio of two mean values across all perturbations:
+                `mean(dot product * func value diff) / mean(dot product * dot product)`.
```
vivekmig, line 353: nit: Would it be possible to make this line a little more detailed to explain this factor for new users? E.g. it may not be immediately clear that func value diff is the same as the difference between the predictor function at its input and perturbed input described above.

captum/metrics/_core/infidelity.py
```diff
@@ -523,7 +540,25 @@ def _next_infidelity(current_n_perturb_samples: int) -> Tensor:
...
             cast(Tuple[Tensor, ...], inputs),
             n_perturb_samples,
-            _next_infidelity,
+            _next_infidelity_tensors,
+            agg_func=_concat_infidelity_tensors,
             max_examples_per_batch=max_examples_per_batch,
         )
-    return metrics_sum * 1 / n_perturb_samples
+
+    attr_times_perturb_sums, perturbed_fwd_diffs = metrics_sum
```
vivekmig, line 548: nit: Can include this above as return of _divide_and_aggregate_metrics ? metrics_sum isn't really applicable anymore

vivekmig (approved):
Looks great, thanks for adding this @aobo-y ! Just some minor nits on documentation.

NarineK (commented):
Thank you for working on this PR @aobo-y!
 
If I understand this implementation correctly we aggregate all inputs and all perturbations in the memory which can lead to out of memory very quickly. The idea of using torch.add or torch.mean as an aggregate function was to help to scale the implementation and avoid out of memory as much as possible.

In `_next_infidelity_tensors`  you have access to normalize input argument (I've just verified it. In python you are still running in the same context and you have access to normalize argument in that context). You can keep track of np.mean(pdt_diff * exp_sum) and np.mean(exp_sum * exp_sum) in `_next_infidelity_tensors`  too and ultimately apply final beta per example in the end. Let me know what you think.

captum/metrics/_core/infidelity.py
aobo-y: sure, i can use latex here

aobo-y:
@NarineK if you mean I keep all the perturbation results in the memory, then yes, I am aware of the consumption of memory. If the perturbation number is too large, we will surely end with out of memory.

In python, I can access the context in the nested aggregation function and know if we need to normalize. Unfortunately, it is not the reason that I have to keep all perturbation result. If I understand correctly, the author's normalization applies the `beta` to each perturbation pair `(pdt_diff, exp_sum)`, not each example.

What you suggested does help calculate the two `mean` which will be used for calculating `beta`. I agree we can aggregate them in each forward. However, in the end, the author uses the `beta` to time each perturbation, then calculate the MSE. In other word, you will only know `beta` after all perturbations,  but then you still need each perturbation pair to calculate `beta * pdt_diff - exp_sum`. 
https://github.com/chihkuanyeh/saliency_evaluation/blob/44a66e2531f30b803be3bf5b0786971b7e7f72a1/infid_sen_utils.py#L298

Mathematically, what we need is `mean((beta * pdt_diff - exp_sum) ^ 2)`. I don't think we can get it with pre-aggregated `mean(pdf_diff)`, `mean(exp_sum)`, or `mean((pdt_diff - exp_sum) ^ 2)`. Let me know if it makes sense.

vivekmig:
> @NarineK if you mean I keep all the perturbation results in the memory, then yes, I am aware of the consumption of memory. If the perturbation number is too large, we will surely end with out of memory.
> 
> In python, I can access the context in the nested aggregation function and know if we need to normalize. Unfortunately, it is not the reason that I have to keep all perturbation result. If I understand correctly, the author's normalization applies the `beta` to each perturbation pair `(pdt_diff, exp_sum)`, not each example.
> 
> What you suggested does help calculate the two `mean` which will be used for calculating `beta`. I agree we can aggregate them in each forward if that's all we need. However, in the end, the author uses the `beta` to time each perturbation, then calculate the MSE. In other word, you will only know `beta` after all perturbations, but then you still need each perturbation pair to calculate `beta * pdt_diff - exp_sum`.
> https://github.com/chihkuanyeh/saliency_evaluation/blob/44a66e2531f30b803be3bf5b0786971b7e7f72a1/infid_sen_utils.py#L298
> 
> Mathematically, what we need is `mean((beta * pdt_diff - exp_sum) ^ 2)`. I don't think we can get it with pre-aggregated `mean(pdf_diff)`, `mean(exp_sum)`, or `mean((pdt_diff - exp_sum) ^ 2)`. Let me know if it makes sense.

I agree that maintaining just the means necessary for beta would not be sufficient to avoid maintaining results per perturbation sample. I think the approach to not store the sample-wise results would be to maintain `mean(attr_times_perturb_sums^2)`, `mean(attr_times_perturb_sums * perturbed_fwd_diffs)` and `mean(perturbed_fwd_diffs ^ 2)` . With these means, it should then be possible to compute both the beta and the final `mean[ (beta * attr_times_perturb_sums - perturbed_fwd_diffs)^2 ]` as:
`beta^2 *  mean(attr_times_perturb_sums^2) - 2 * beta * mean(attr_times_perturb_sums * perturbed_fwd_diffs) + mean(perturbed_fwd_diffs ^ 2)`

 based on the expansion of (a - b)^2

This approach would avoid the additional memory, but at the tradeoff of a potentially trickier formulation to follow. The additional memory used here should be on the order of batch_size * n_perturb_samples (full input perturbations wouldn't be maintained), so this shouldn't be a large issue / lead to OOMs with typical use cases. But to be on the safer side, if we expect potentially larger values, might be worth considering the alternative approach. What do you think @NarineK , @aobo-y ?

NarineK:
> @NarineK if you mean I keep all the perturbation results in the memory, then yes, I am aware of the consumption of memory. If the perturbation number is too large, we will surely end with out of memory.
> 
> In python, I can access the context in the nested aggregation function and know if we need to normalize. Unfortunately, it is not the reason that I have to keep all perturbation result. If I understand correctly, the author's normalization applies the `beta` to each perturbation pair `(pdt_diff, exp_sum)`, not each example.
> 
> What you suggested does help calculate the two `mean` which will be used for calculating `beta`. I agree we can aggregate them in each forward if that's all we need. However, in the end, the author uses the `beta` to time each perturbation, then calculate the MSE. In other word, you will only know `beta` after all perturbations, but then you still need each perturbation pair to calculate `beta * pdt_diff - exp_sum`.
> https://github.com/chihkuanyeh/saliency_evaluation/blob/44a66e2531f30b803be3bf5b0786971b7e7f72a1/infid_sen_utils.py#L298
> 
> Mathematically, what we need is `mean((beta * pdt_diff - exp_sum) ^ 2)`. I don't think we can get it with pre-aggregated `mean(pdf_diff)`, `mean(exp_sum)`, or `mean((pdt_diff - exp_sum) ^ 2)`. Let me know if it makes sense.

 


> ttr_times_perturb_sums * perturbed_fwd_diffs) + mean(perturbed_fwd_diffs ^ 2)

I saw that beta is a scalar and it looked to me that we should be able to do that sample based. Sorry for the confusion. 

sum(a - b) = sum (a) - sum(b), so if we have sum(a) and sum(b) we can compute sum(beta * a - b) = beta * sum(a) - sum(b)
In terms of squares as @vivekmig mentioned sum([beta * a - b]^2) = sum((beta*a)^2) - 2*beta * sum(ab) + sum(b^2). I think that it is important to be memory aware and we shouldn't sacrifice memory for a normalization factor.

captum/metrics/_core/infidelity.py
```diff
@@ -345,6 +346,20 @@ def infidelity(
...
                 Default: None
+        normalize (bool, optional): Normalize the dot product of the input
+                perturbation and the attribution so the infidelity value is invariant
+                to constant scaling of the attribution values. The normalization factor
+                beta is defined as the ratio of two mean values:
+                $$ \beta = \frac{
+                    \mathbb{E}_{I \sim \mu_I} [ I^T \Phi(f, x) (f(x) - f(x - I)) ]
+                }{
+                    \mathbb{E}_{I \sim \mu_I} [ (I^T \Phi(f, x))^2 ]
+                } $$.
```
aobo-y, line 357: compiled equation
Screen Shot 2021-04-05 at 4 43 53 PM

aobo-y:
@NarineK  @vivekmig The unittest `test_classification_infidelity_tpl_target_w_baseline` passed the latest pytorch but failed in v1.2 and v1.3, because of the precision violates the 0.05 tolerance in the those older version of pytorch.

I tried to study where the differences come from and found an inconsistent behavior of `IntegratedGradients`. The attribution is `dtype.float64` in latest pytorch but `dtype.float32` in older version. It is caused by the ambiguous call of `torch.tensor(step_sizes)` at https://github.com/pytorch/captum/blob/master/captum/attr/_core/integrated_gradients.py#L362

just in case this is unknown before.

For this PR, this means due to the precision in float32, the difference between `mean(a^-2ab+b^2)` and `mean((a-b)^2)` can be larger than 0.05.

aobo-y:
Here is the detailed example https://github.com/pytorch/captum/blob/master/tests/metrics/test_infidelity.py#L268
In pytorch-1.3, the infidelity score is `tensor([0.1068, 0.0208, 0.0000, 0.0833])` if we keep the default `float` attribution but if I convert it to double `attr = attr.double()`, the infidelity will become `tensor([1.0687e-01, 2.0955e-09, 8.3819e-09, 8.3819e-09], dtype=torch.float64)` and successfully pass the tests.

But in the latest pytorch, both `float` and `double` attribution lead to the same infidelity:`tensor([0.1069, 0.0000, 0.0000, 0.0000])` and `tensor([0.1069, 0.0000, 0.0000, 0.0000], dtype=torch.float64)`

captum/metrics/_core/infidelity.py
```diff
@@ -472,31 +489,44 @@ def _next_infidelity(current_n_perturb_samples: int) -> Tensor:
...
+            # to calculate MSE in its polynomial expansion:
+            # (a-b)^2 = a^2 - 2ab + b^2
+            return (
+                attr_times_perturb_sums.pow(2).sum(-1),
+                (attr_times_perturb_sums * perturbed_fwd_diffs).sum(-1),
+                perturbed_fwd_diffs.pow(2).sum(-1),
+            )
+        else:
+            # returns (a-b)^2 if no need to normalize
+            return ((attr_times_perturb_sums - perturbed_fwd_diffs).pow(2).sum(-1),)
```
aobo-y, line 526: due to the above issue, I keep two ways for aggregation based on `normalize` instead of always using `a^2-2ab+b^2`:
- `(a-b)^2` if not normalize
- `a^2`, `ab`, `b^2` if normalize

This change allows me to pass the tests.
But still worth noting that in older version of pytorch, when normalize, `a^2-2ab+b^2` will lose some precision compared with direct `(a-b)^2`

vivekmig: Makes sense, sounds good, this fix looks good to me!

NarineK (approved):
Thank you for working on this, @aobo-y!  Looks good to me! I left one small nit comment.

aobo-y:
Updated the PR with `safe_div` as suggested.
But opened another issue abt `safe_div` for discussion & future improvement.

PR #647, opened 2021-04-05
Fix gradient requirements for layer methods

vivekmig:
This updates gradient requirements to be set on layer inputs / outputs rather than original inputs, which ensures that gradient requirements are set when inputs are non-floating point (e.g. token indices). This also avoids unnecessarily requiring gradients between the input and target layer, when only layer gradients are required.

### DIFF
--- a/captum/_utils/gradient.py
+++ b/captum/_utils/gradient.py
@@ -24,7 +24,7 @@ from captum._utils.typing import (
 
 
 def apply_gradient_requirements(
-    inputs: Tuple[Tensor, ...], warn: bool = False
+    inputs: Tuple[Tensor, ...], warn: bool = True
 ) -> List[bool]:
     """
     Iterates through tuple on input tensors and sets requires_grad to be true on
--- a/captum/_utils/gradient.py
+++ b/captum/_utils/gradient.py
@@ -23,7 +23,9 @@ from captum._utils.typing import (
 )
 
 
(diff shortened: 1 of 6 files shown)


captum/_utils/gradient.py
```diff
@@ -23,7 +23,9 @@
 )
 
 
-def apply_gradient_requirements(inputs: Tuple[Tensor, ...]) -> List[bool]:
+def apply_gradient_requirements(
+    inputs: Tuple[Tensor, ...], warn: bool = False
```
NarineK, line 27: nit: In order to support original behavior don't we want to not warn only when we know that warning is not necessary in case of layer approaches ?

captum/_utils/gradient.py
```diff
@@ -251,6 +258,8 @@ def forward_hook(module, inp, out=None):
 
             if not is_eval_tuple:
                 eval_tsrs = (eval_tsrs,)
+            if require_layer_grads:
+                apply_gradient_requirements(eval_tsrs, warn=False)
```
NarineK, line 262: Default behavior now warn=False too, right ? Perhaps we don't want that warn=False is default behavior.

NarineK (approved):
LGTM! Thank you for the fix. Couple nits and questions.

captum/_utils/gradient.py
vivekmig: Great catch, thanks! Yes, you're definitely right, meant to set the default to True

PR #649, opened 2021-04-06
Fix KernelShap Overflow Issues

vivekmig:
This ensures that Lime / Kernel Shap interpretable model inputs, targets, and similarity scores are floats, which avoids sporadic int overflow issues occurring in Kernel Shap.

### DIFF
--- a/captum/attr/_core/lime.py
+++ b/captum/attr/_core/lime.py
@@ -511,17 +511,17 @@ class LimeBase(PerturbationAttribution):
             if show_progress:
                 attr_progress.close()
 
-            combined_interp_inps = torch.cat(interpretable_inps).float()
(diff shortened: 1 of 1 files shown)


captum/attr/_core/lime.py
```diff
@@ -511,17 +511,17 @@ def attribute(
             if show_progress:
                 attr_progress.close()
 
-            combined_interp_inps = torch.cat(interpretable_inps)
+            combined_interp_inps = torch.cat(interpretable_inps).float()
```
miguelmartin75, line 514: Wondering if .double() would be a good alternative in some cases?

miguelmartin75 (approved):
lgtm

PR #655, opened 2021-04-26
Optim-wip: Add the pre-trained InceptionV1 Places365 model

ProGamerGov:
This PR adds the InceptionV1 / GoogleNet Places365 model from MIT!

* Used by Lucid here: https://github.com/tensorflow/lucid/blob/master/lucid/modelzoo/caffe_models/InceptionV1.py#L89
* Originally from: https://github.com/CSAILVision/places365
* Associated research paper: https://arxiv.org/abs/1610.02055
* Website for the Places image datasets: http://places2.csail.mit.edu/

* Pictured below: An Activation Atlas for the the Mixed5b ReLU layer

inceptionv1_places365_mixed5b_relu_atlas

atlas_21_attempts_g20x20_mixed5b

The InceptionV1 Places365 is the second most popular model from Lucid's model zoo, and appears to have some unique neuron types like 'perspective neurons'. Examples of perceptive detectors can be found on OpenAI's Microscope [here](https://microscope.openai.com/models/inceptionv1_caffe_places365/inception_4a_5x5_inception_4a_5x5_0?models.op.feature_vis.type=neuron&models.op.technique=feature_vis), and in the model at layer `model.mixed4a.conv_5x5`.

The changes in this PR are as follows:

* Added the MIT InceptionV1 model trained on the Places365 Standard dataset. The mean normalization values in the model's `transform_input` function were calculated by me using the version of the Places365 Standard dataset where every image was resized to `256x256`. I then reduced the size down to `224x224` via a transform. This resized version of the dataset had a smaller file size than the normal version, but I think the mean values should be same.
* Added the list of class names for the InceptionV1 Places365 model.
* Added `__constants__` to both models in anticipation of upcoming JIT support for `Typing.Union`: https://github.com/pytorch/pytorch/pull/53180 The PyTorch docs suggest using `typing.Final` instead of `__constants__`, but earlier versions of Python don't have `typing.Final`.
* Auxiliary branch layer names were renamed for both models, based on Torchvision's [GoogLeNet model](https://github.com/pytorch/vision/blob/master/torchvision/models/googlenet.py).
* Removed `padding` option from `Conv2dSame` as `same` does not involve setting a padding value.
* Fixed the download link for the InceptionV1 model.
* Corrected all instances of `super()` in the model class files to `super().__init__()`.
* Improved existing model tests, added new CUDA tests, and added tests for the Places365 model.
* Replaced input range assertions in `_transform_input` with user warnings.
* JIT support for both InceptionV1 models

I also generated a KLT matrix for the Places365 Standard dataset, and it can be used as an input the `ToRGB` transform in `NaturalImage`:

```
# RGB format, use torch.flip([0, 1]) to convert to BGR
klt_mtx = torch.tensor(
    [[0.0175, -0.0713, -0.2234], [-0.0340, -0.0019, -0.2299], [0.0168, 0.0702, -0.2332]]
)
```
There's not really a large different compared to just using the default ImageNet KLT matrix, though I found that I had to increase the learning rate to `0.06` to get comparable results.


ProGamerGov:
@NarineK This PR is ready to be reviewed!

NarineK:
> @NarineK This PR is ready to be reviewed! The failed tests are caused by the update to `black` which is unrelated to the code introduced in this PR.

Thank you, @ProGamerGov! I'll take a look into it next week - This week is busy.

ProGamerGov:
All the tests should pass once https://github.com/pytorch/captum/pull/656 is merged!

NarineK:
@ProGamerGov, it looks like there is merge conflict related to some files. Do you mind checking that ? Do we still want to merge this PR?

ProGamerGov:
@NarineK I'll remake this PR to resolve the conflicts

ProGamerGov:
@NarineK I remade this PR here: https://github.com/pytorch/captum/pull/935

NarineK:
@ProGamerGov, if we remade this PR here: https://github.com/pytorch/captum/pull/935 can we then close this one ?

PR #656, opened 2021-04-26
Optim-wip: Fix failing tests, model download link, & more

ProGamerGov:
* Quick fix for the model file path as it's breaking all the notebooks & tests.

* It also looks like `black` was updated and now is reporting an issue with `captum/attr/_core/feature_ablation.py` &  `captum/metrics/_core/infidelity.py`, so I've resolved that as well. The files in the master branch may be a bit different, but we need to at least keep the tests passing for the other areas of Captum until we merge into master.

* Added new `ImageTensor` tests.

* Added `detach()` to `InputOptimization` to fix out of memory crashes.

* Made sure that `SkipLayer` can handle any addition arguments to it's init and forward functions. This adds to the usefulness of the layer and improves it beyond just copying `nn.Identity`.

* Fixed various bugs.

### DIFF
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -179,7 +179,6 @@ class FFTImage(ImageParameterization):
         return torch_rfft, torch_irfft, torch_fftfreq
 
     def forward(self) -> torch.Tensor:
-        h, w = self.size
         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
         output = self.torch_irfft(scaled_spectrum)
         return output.refine_names("B", "C", "H", "W")
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -138,9 +138,7 @@ class FFTImage(ImageParameterization):
     def rfft2d_freqs(self, height: int, width: int) -> torch.Tensor:
         """Computes 2D spectrum frequencies."""
         fy = self.torch_fftfreq(height)[:, None]
-        # on odd input dimensions we need to keep one additional frequency
-        wadd = 2 if width % 2 == 1 else 1
-        fx = self.torch_fftfreq(width)[: width // 2 + wadd]
+        fx = self.torch_fftfreq(width)[: width // 2 + 1]
         return torch.sqrt((fx * fx) + (fy * fy))
 
     def get_fft_funcs(self) -> Tuple[Callable, Callable, Callable]:
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -35,13 +35,13 @@ class ImageTensor(torch.Tensor):
             return super().__new__(cls, x, *args, **kwargs)
 
     @classmethod
-    def open(cls, path: str, scale: float = 255.0) -> "ImageTensor":
+    def open(cls, path: str, scale: float = 255.0, mode: str = "RGB") -> "ImageTensor":
         if path.startswith("https://") or path.startswith("http://"):
             response = requests.get(path, stream=True)
             img = Image.open(response.raw)
         else:
             img = Image.open(path)
-        img_np = np.array(img.convert("RGB")).astype(np.float32)
+        img_np = np.array(img.convert(mode)).astype(np.float32)
         return cls(img_np.transpose(2, 0, 1) / scale)
 
     def __repr__(self) -> str:
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -418,17 +418,17 @@ class NaturalImage(ImageParameterization):
 
     def __init__(
         self,
-        size: Tuple[int, int] = [224, 224],
+        size: Tuple[int, int] = (224, 224),
         channels: int = 3,
         batch: int = 1,
         init: Optional[torch.Tensor] = None,
         parameterization: ImageParameterization = FFTImage,
         squash_func: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
-        decorrelation_module: Optional[nn.Module] = ToRGB(transform="klt"),
-        decorrelate_init: bool = True,
+        decorrelation_module: Optional[nn.Module] = None,
+        decorrelate_init: bool = False,
     ) -> None:
         super().__init__()
-        self.decorrelate = decorrelation_module
+        self.decorrelate = decorrelation_module or ToRGB(transform="klt")
         if init is not None:
             assert init.dim() == 3 or init.dim() == 4
             if decorrelate_init:
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -421,15 +421,17 @@ class NaturalImage(ImageParameterization):
         init: Optional[torch.Tensor] = None,
         parameterization: ImageParameterization = FFTImage,
         squash_func: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
-        decorrelation_module: Optional[nn.Module] = None,
+        decorrelation_module: Optional[nn.Module] = ToRGB(transform="klt"),
         decorrelate_init: bool = True,
     ) -> None:
         super().__init__()
-        self.decorrelate = decorrelation_module or ToRGB(transform="klt")
+        self.decorrelate = (
+            decorrelation_module.cpu() if decorrelation_module is not None else None
+        )
         if init is not None:
+            assert not init.is_cuda
             assert init.dim() == 3 or init.dim() == 4
-            if decorrelate_init:
-                assert self.decorrelate is not None
+            if decorrelate_init and self.decorrelate is not None:
                 init = (
                     init.refine_names("B", "C", "H", "W")
                     if init.dim() == 4
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -209,6 +209,9 @@ class PixelImage(ImageParameterization):
 
 
 class LaplacianImage(ImageParameterization):
+    """
+    TODO: Fix divison by 6 in setup_input when init is not None.
+    """
     def __init__(
         self,
         size: Tuple[int, int] = None,
@@ -386,7 +389,9 @@ class SharedImage(ImageParameterization):
 
 
 class NaturalImage(ImageParameterization):
-    r"""Outputs an optimizable input image.
+    r"""TODO: Resolve device issues with default ToRGB instance, and init tensors.
+    
+    Outputs an optimizable input image.
 
     By convention, single images are CHW and float32s in [0,1].
     The underlying parameterization can be decorrelated via a ToRGB transform.
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -425,7 +425,7 @@ class NaturalImage(ImageParameterization):
         parameterization: ImageParameterization = FFTImage,
         squash_func: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
         decorrelation_module: Optional[nn.Module] = None,
-        decorrelate_init: bool = False,
+        decorrelate_init: bool = True,
     ) -> None:
         super().__init__()
         self.decorrelate = decorrelation_module or ToRGB(transform="klt")
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -430,12 +430,7 @@ class NaturalImage(ImageParameterization):
         decorrelate_init: bool = True,
     ) -> None:
         super().__init__()
-        # Deep copy to avoid issue with creating class instance in the function
-        # signature
-        if isinstance(decorrelation_module, ToRGB):
-            decorrelation_module = deepcopy(decorrelation_module)
         self.decorrelate = decorrelation_module
-
         if init is not None:
             assert init.dim() == 3 or init.dim() == 4
             if decorrelate_init and self.decorrelate is not None:
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -116,7 +116,6 @@ class FFTImage(ImageParameterization):
         )
         scale = scale * ((self.size[0] * self.size[1]) ** (1 / 2))
         spectrum_scale = scale[None, :, :, None]
-        self.register_buffer("spectrum_scale", spectrum_scale)
 
         if init is None:
             coeffs_shape = (
@@ -131,8 +130,10 @@ class FFTImage(ImageParameterization):
             )  # names=["C", "H_f", "W_f", "complex"]
             fourier_coeffs = random_coeffs / 50
         else:
+            spectrum_scale = spectrum_scale.to(init.device)
             fourier_coeffs = self.torch_rfft(init) / spectrum_scale
 
+        self.register_buffer("spectrum_scale", spectrum_scale)
         self.fourier_coeffs = nn.Parameter(fourier_coeffs)
 
     def rfft2d_freqs(self, height: int, width: int) -> torch.Tensor:
@@ -390,7 +391,7 @@ class SharedImage(ImageParameterization):
 
 class NaturalImage(ImageParameterization):
     r"""TODO: Resolve device issues with default ToRGB instance, and init tensors.
-    
+
     Outputs an optimizable input image.
 
     By convention, single images are CHW and float32s in [0,1].
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -432,11 +432,10 @@ class NaturalImage(ImageParameterization):
         super().__init__()
         # Deep copy to avoid issue with creating class instance in the function
         # signature
-        self.decorrelate = (
-            deepcopy(decorrelation_module)
-            if isinstance(decorrelation_module, ToRGB)
-            else decorrelation_module
-        )
+        if isinstance(decorrelation_module, ToRGB):
+            decorrelation_module = deepcopy(decorrelation_module)
+        self.decorrelate = decorrelation_module
+
         if init is not None:
             assert init.dim() == 3 or init.dim() == 4
             if decorrelate_init and self.decorrelate is not None:
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -213,6 +213,7 @@ class LaplacianImage(ImageParameterization):
     """
     TODO: Fix divison by 6 in setup_input when init is not None.
     """
+
     def __init__(
         self,
         size: Tuple[int, int] = None,
@@ -390,9 +391,7 @@ class SharedImage(ImageParameterization):
 
 
 class NaturalImage(ImageParameterization):
-    r"""TODO: Resolve device issues with default ToRGB instance, and init tensors.
-
-    Outputs an optimizable input image.
+    r"""Outputs an optimizable input image.
 
     By convention, single images are CHW and float32s in [0,1].
     The underlying parameterization can be decorrelated via a ToRGB transform.
@@ -431,11 +430,14 @@ class NaturalImage(ImageParameterization):
         decorrelate_init: bool = True,
     ) -> None:
         super().__init__()
+        # Deep copy to avoid issue with creating class instance in the function
+        # signature
         self.decorrelate = (
-            decorrelation_module.cpu() if decorrelation_module is not None else None
+            deepcopy(decorrelation_module)
+            if isinstance(decorrelation_module, ToRGB)
+            else decorrelation_module
         )
         if init is not None:
-            assert not init.is_cuda
             assert init.dim() == 3 or init.dim() == 4
             if decorrelate_init and self.decorrelate is not None:
                 init = (
--- a/captum/optim/models/_common.py
+++ b/captum/optim/models/_common.py
@@ -204,7 +204,8 @@ class SkipLayer(torch.nn.Module):
     like ReLU for circuits research.
 
     This layer works almost exactly the same way that nn.Indentiy does, except it also
-    ignores any additional arguments passed to the forward function.
+    ignores any additional arguments passed to the forward function. Any layer replaced
+    by SkipLayer must have the same input and output shapes.
 
     See nn.Identity for more details:
     https://pytorch.org/docs/stable/generated/torch.nn.Identity.html
--- a/captum/optim/models/_common.py
+++ b/captum/optim/models/_common.py
@@ -199,13 +199,36 @@ def collect_activations(
 
 class SkipLayer(torch.nn.Module):
     """
-    This layer is made to take the place of nonlinear activation layers like ReLU.
+    This layer is made to take the place of any layer that needs to be skipped over
+    during the forward pass. Use cases include removing nonlinear activation layers
+    like ReLU for circuits research.
+
+    This layer works almost exactly the same way that nn.Indentiy does, except it also
+    ignores any additional arguments passed to the forward function.
+
+    See nn.Identity for more details:
+    https://pytorch.org/docs/stable/generated/torch.nn.Identity.html
+
+    Args:
+        args (Any): Any argument. Arguments will be safely ignored.
+        kwargs (Any) Any keyword argument. Arguments will be safely ignored.
     """
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__()
 
-    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+    def forward(
+        self, x: Union[torch.Tensor, Tuple[torch.Tensor]], *args, **kwargs
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor]]:
+        """
+        Args:
+            x (torch.Tensor or tuple of torch.Tensor): The input tensor or tensors.
+            args (Any): Any argument. Arguments will be safely ignored.
+            kwargs (Any) Any keyword argument. Arguments will be safely ignored.
+        Returns:
+            x (torch.Tensor or tuple of torch.Tensor): The unmodified input tensor or
+                tensors.
+        """
         return x
 
 
--- a/captum/optim/models/_common.py
+++ b/captum/optim/models/_common.py
@@ -202,7 +202,10 @@ class SkipLayer(torch.nn.Module):
     This layer is made to take the place of nonlinear activation layers like ReLU.
     """
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__()
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         return x
 
 
--- a/captum/attr/_core/feature_ablation.py
+++ b/captum/attr/_core/feature_ablation.py
@@ -62,7 +62,7 @@ class FeatureAblation(PerturbationAttribution):
         additional_forward_args: Any = None,
         feature_mask: Union[None, Tensor, Tuple[Tensor, ...]] = None,
         perturbations_per_eval: int = 1,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> TensorOrTupleOfTensorsGeneric:
         r"""
         Args:
@@ -321,7 +321,7 @@ class FeatureAblation(PerturbationAttribution):
                     baselines,
                     feature_mask,
                     perturbations_per_eval,
-                    **kwargs
+                    **kwargs,
                 ):
                     # modified_eval dimensions: 1D tensor with length
                     # equal to #num_examples * #features in batch
@@ -373,7 +373,7 @@ class FeatureAblation(PerturbationAttribution):
         baselines,
         input_mask,
         perturbations_per_eval,
-        **kwargs
+        **kwargs,
     ):
         """
         This method is a generator which yields each perturbation to be evaluated
@@ -458,7 +458,7 @@ class FeatureAblation(PerturbationAttribution):
                 baseline,
                 num_features_processed,
                 num_features_processed + current_num_ablated_features,
-                **extra_args
+                **extra_args,
             )
 
             # current_features[i] has dimension
--- a/captum/metrics/_core/infidelity.py
+++ b/captum/metrics/_core/infidelity.py
@@ -68,7 +68,7 @@ def infidelity_perturb_func_decorator(multipy_by_inputs: bool = True) -> Callabl
         def default_perturb_func(
             inputs: TensorOrTupleOfTensorsGeneric, baselines: BaselineType = None
         ):
-            r""""""
+            r""" """
             inputs_perturbed = (
                 pertub_func(inputs, baselines)
                 if baselines is not None
@@ -380,7 +380,7 @@ def infidelity(
         """
 
         def call_perturb_func():
-            r""""""
+            r""" """
             baselines_pert = None
             inputs_pert: Union[Tensor, Tuple[Tensor, ...]]
             if len(inputs_expanded) == 1:
--- a/captum/optim/__init__.py
+++ b/captum/optim/__init__.py
@@ -12,7 +12,6 @@ from captum.optim._utils.image.common import (  # noqa: F401
     show,
     weights_to_heatmap_2d,
 )
-from captum.optim._utils.reducer import ChannelReducer, posneg  # noqa: F401
 
 __all__ = [
     "InputOptimization",
--- a/captum/optim/_core/optimization.py
+++ b/captum/optim/_core/optimization.py
@@ -111,7 +111,7 @@ class InputOptimization(Objective, Parameterized):
         self,
         stop_criteria: Optional[StopCriteria] = None,
         optimizer: Optional[optim.Optimizer] = None,
-        loss_summarize_fn: Optional[Callable] = default_loss_summarize,
+        loss_summarize_fn: Optional[Callable] = None,
         lr: float = 0.025,
     ) -> torch.Tensor:
         r"""Optimize input based on loss function and objectives.
@@ -131,6 +131,7 @@ class InputOptimization(Objective, Parameterized):
         stop_criteria = stop_criteria or n_steps(512)
         optimizer = optimizer or optim.Adam(self.parameters(), lr=lr)
         assert isinstance(optimizer, optim.Optimizer)
+        loss_summarize_fn = loss_summarize_fn or default_loss_summarize
 
         history = []
         step = 0
--- a/captum/optim/_core/optimization.py
+++ b/captum/optim/_core/optimization.py
@@ -138,7 +138,7 @@ class InputOptimization(Objective, Parameterized):
             while stop_criteria(step, self, history, optimizer):
                 optimizer.zero_grad()
                 loss_value = loss_summarize_fn(self.loss())
-                history.append(loss_value)
+                history.append(loss_value.clone().detach().cpu())
                 loss_value.backward()
                 optimizer.step()
                 step += 1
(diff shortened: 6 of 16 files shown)


ProGamerGov:
I have also just realized that `InputOptimization.optimize` returns `history` as an `ImageTensor`, which shouldn't be happening.

NarineK:
Thank you for the fixes, @ProGamerGov ! It doesn't seem to fail on master. Are those python files different on optim-wip branch ?

ProGamerGov:
@NarineK I think the files are slightly different, but @vivekmig also changed the same files in the most recent PR for the master branch: https://github.com/pytorch/captum/pull/658/commits/5cf38cb3e5efe8a250f99e7f3c707a2e30ec9340

ProGamerGov:
Looks like lint_test_py36_pip and lint_test_py37_conda tests are now failing due to this error:

```
    @staticmethod
    def __new__(
        cls: Type["ImageTensor"],
        x: Union[List, np.ndarray, torch.Tensor] = [],
        *args,
        **kwargs,
    ) -> torch.Tensor:
        if isinstance(x, torch.Tensor) and x.is_cuda:
            x.show = MethodType(cls.show, x)
            x.export = MethodType(cls.export, x)
            return x
        else:
>           return super().__new__(cls, x, *args, **kwargs)
E           TypeError: object.__new__(ImageTensor) is not safe, use Tensor.__new__()
```

ProGamerGov:
I found the source of the `TypeError` when subclassing `torch.Tensor`! it's a bug with the nightly PyTorch builds. The `lint_test_py36_pip` test uses `torch-1.9.0.dev20210501+cpu`  and the `lint_test_py37_conda` test uses `torch==1.9.0.dev20210501` , and thus they both fail on tests involving `ImageTensor`.

I made an issue post on the PyTorch repo here: https://github.com/pytorch/pytorch/issues/57421

ProGamerGov:
The issue with  `ImageTensor`'s `__new__` function is now resolved in the latest version of the PyTorch nightly build!

ProGamerGov:
I have discovered a new issue with the code:

Running these two lines of code:

```
image_param = opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()
image_param = opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()
```

Results in this error:

```
RuntimeError                              Traceback (most recent call last)

<ipython-input-3-f944a33c48db> in <module>()
      1 image_param = opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()
----> 2 image_param = opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()

2 frames

/content/captum/captum/optim/_param/image/images.py in __init__(self, size, channels, batch, init, parameterization, squash_func, decorrelation_module, decorrelate_init)
    439                     else init.refine_names("C", "H", "W")
    440                 )
--> 441                 init = self.decorrelate(init, inverse=True).rename(None)
    442             if squash_func is None:
    443 

/usr/local/lib/python3.7/dist-packages/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
    887             result = self._slow_forward(*input, **kwargs)
    888         else:
--> 889             result = self.forward(*input, **kwargs)
    890         for hook in itertools.chain(
    891                 _global_forward_hooks.values(),

/content/captum/captum/optim/_param/image/transforms.py in forward(self, x, inverse)
    108         flat = x.flatten(("H", "W"), "spatials")
    109         if inverse:
--> 110             correct = torch.inverse(self.transform) @ flat
    111         else:
    112             correct = self.transform @ flat

RuntimeError: Tensor for argument #3 'mat2' is on CPU, but expected it to be on GPU (while checking arguments for addmm)
```

ProGamerGov:
So, currently `NaturalImage` is setup like this:

```
class NaturalImage(ImageParameterization):
    def __init__(
        self,
        size: Tuple[int, int] = (224, 224),
        channels: int = 3,
        batch: int = 1,
        init: Optional[torch.Tensor] = None,
        parameterization: ImageParameterization = FFTImage,
        squash_func: Optional[Callable[[torch.Tensor], torch.Tensor]] = None,
        decorrelation_module: Optional[nn.Module] = None,
        decorrelate_init: bool = True,
    ) -> None:
        super().__init__()
        self.decorrelate = decorrelation_module or ToRGB(transform="klt")
        if init is not None:
            assert init.dim() == 3 or init.dim() == 4
            if decorrelate_init:
                assert self.decorrelate is not None
                init = (
                    init.refine_names("B", "C", "H", "W")
                    if init.dim() == 4
                    else init.refine_names("C", "H", "W")
                )
                init = self.decorrelate(init, inverse=True).rename(None)
            if squash_func is None:

                def squash_func(x: torch.Tensor) -> torch.Tensor:
                    return x.clamp(0, 1)

        else:
            if squash_func is None:

                squash_func = torch.sigmoid

        self.squash_func = squash_func
        self.parameterization = parameterization(
            size=size, channels=channels, batch=batch, init=init
        )

    def forward(self) -> torch.Tensor:
        image = self.parameterization()
        if self.decorrelate is not None:
            image = self.decorrelate(image)
        image = image.rename(None)  # TODO: the world is not yet ready
        return ImageTensor(self.squash_func(image))
```

If you want to disable color decorrelation / recorrelation, then you either need to set `NaturalImage`'s `decorrelate` variable to `None` or you need to set `decorrelation_module` to be an empty module. I think that there may be a more elegant solution to turning the color decorrelation on and off, but I haven't figured anything out yet.

ProGamerGov:
Well I have no idea what happened here:

```
self = <tests.optim.core.test_optimization.TestInputOptimization testMethod=test_input_optimization>

    def test_input_optimization(self) -> None:
        if torch.__version__ <= "1.2.0":
            raise unittest.SkipTest(
                "Skipping InputOptimization test due to insufficient Torch version."
            )
        model = BasicModel_ConvNet_Optim()
        loss_fn = opt.loss.ChannelActivation(model.layer, 0)
        obj = opt.InputOptimization(model, loss_function=loss_fn)
        n_steps = 5
        history = obj.optimize(opt.optimization.n_steps(n_steps, show_progress=False))
>       self.assertTrue(history[0] > history[-1])
E       AssertionError: ImageTensor(False) is not true
```

My fix for `_rand_select` is completely unrelated, so I don't understand why this error showed up.

ProGamerGov:
The FFTImage function `rfft2d_freqs` has a line that seems to cause the very issue that it claims to fix:

```
    def rfft2d_freqs(self, height: int, width: int) -> torch.Tensor:
        """Computes 2D spectrum frequencies."""
        fy = self.torch_fftfreq(height)[:, None]
        # on odd input dimensions we need to keep one additional frequency
        wadd = 2 if width % 2 == 1 else 1
        fx = self.torch_fftfreq(width)[: width // 2 + wadd]
        return torch.sqrt((fx * fx) + (fy * fy))
```

I think that we can remove these lines:

```
        # on odd input dimensions we need to keep one additional frequency
        wadd = 2 if width % 2 == 1 else 1
```

And then just always add 1 to the line:

```
        fx = self.torch_fftfreq(width)[: width // 2 + 1]
```

Hopefully this doesn't break anything? Though I don't think that any of our current tutorials or tests were using inputs with odd widths, and the issue that the fix was attempting to solve is seemingly not present in PyTorch. The issue does however appear to be a problem with the NumPy FFTImage class, and the TensorFlow [code that Lucid uses](https://github.com/tensorflow/lucid/blob/master/lucid/optvis/param/spatial.py#L52).

NarineK:
> @NarineK I think the files are slightly different, but @vivekmig also changed the same files in the most recent PR for the master branch: [5cf38cb](https://github.com/pytorch/captum/commit/5cf38cb3e5efe8a250f99e7f3c707a2e30ec9340). When we merge with the master branch we can updated everything, but for the moment updating the files will ensure that the tests keep passing.

I see - or we can sync optim-wip branch with master so that we get @vivekmig's  changes. In that case we will not have conflicts with the master branch.

NarineK:
> I found the source of the `TypeError` when subclassing `torch.Tensor`! it's a bug with the nightly PyTorch builds. The `lint_test_py36_pip` test uses `torch-1.9.0.dev20210501+cpu` and the `lint_test_py37_conda` test uses `torch==1.9.0.dev20210501` , and thus they both fail on tests involving `ImageTensor`.
> 
> I made an issue post on the PyTorch repo here: [pytorch/pytorch#57421](https://github.com/pytorch/pytorch/issues/57421)

Sounds great! Thank you! It looks like they fixed it.

NarineK:
> #3 'mat2' is

Interesting, according to code snippet it looks like it worked first time calling `opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()` but failed second time ?

NarineK:
> code that Lucid uses

According to lucid implementation it looks like we need to add 2 for odd cases and later remove that one additional pixel. What was the original problem that you saw with line: `wadd = 2 if width % 2 == 1 else 1` ? What was failing ?

ProGamerGov:
> 
> 
> > #3 'mat2' is
> 
> Interesting, according to code snippet it looks like it worked first time calling `opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()` but failed second time ?

The issue was that by creating the ToRGB instance in the init function, and it creates a single instance that will be passed to future constructor calls as default. It's apparently a well-known trap when using default parameters:

```
# Example code
import torch
import torch.nn as nn

class Test1(nn.Module):
    def __init__(self):
        super().__init__()
        self.register_buffer("test_tensor", torch.ones(5))

    def forward(self, x):
        return x * self.test_tensor

class Test2(nn.Module):
    def __init__(self, x, test1 = Test1()):
        super().__init__()
        self.test1 = test1
        x = self.test1(x)

    def forward(self, x):
        return self.test1(x)

t = Test2(torch.ones(5)).cuda()
t = Test2(torch.ones(5)).cuda() # Fails
```

```
t = Test2(torch.ones(5), Test1()).cuda()
t = Test2(torch.ones(5), Test1()).cuda() # Works
```

---

> 
> 
> > code that Lucid uses
> 
> According to lucid implementation it looks like we need to add 2 for odd cases and later remove that one additional pixel. What was the original problem that you saw with line: `wadd = 2 if width % 2 == 1 else 1` ? What was failing ?


The original issue was that `fourier_coeffs` and `spectrum_scale` would have a size mistmatch if the width given to FFTImage either through the `size` argument, or the `init` tensor was an odd number.

This code here would result in the following error:

```
image = opt.images.FFTImage((512, 405))
out = image()
```

```
RuntimeError                              Traceback (most recent call last)

<ipython-input-3-358d0813d735> in <module>()
      1 image = opt.images.FFTImage((512, 405))
----> 2 out = image()

1 frames

/content/captum/captum/optim/_param/image/images.py in forward(self)
    183     def forward(self) -> torch.Tensor:
    184         h, w = self.size
--> 185         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
    186         output = self.torch_irfft(scaled_spectrum)
    187         return output.refine_names("B", "C", "H", "W")

RuntimeError: The size of tensor a (203) must match the size of tensor b (204) at non-singleton dimension 3
```

And the same issue occurred when the init tensor size had an odd width:

```
image = opt.images.FFTImage(init=torch.ones(1, 3, 512, 405))
out = image()
```

```
RuntimeError                              Traceback (most recent call last)

<ipython-input-5-48fc824a1f55> in <module>()
----> 1 image = opt.images.FFTImage(init=torch.ones(1, 3, 512, 405))
      2 out = image()

/content/captum/captum/optim/_param/image/images.py in __init__(self, size, channels, batch, init)
    132             fourier_coeffs = random_coeffs / 50
    133         else:
--> 134             fourier_coeffs = self.torch_rfft(init) / spectrum_scale
    135 
    136         self.fourier_coeffs = nn.Parameter(fourier_coeffs)

RuntimeError: The size of tensor a (203) must match the size of tensor b (204) at non-singleton dimension 3
```

ProGamerGov:
@NarineK The lines I altered in `feature_ablation.py` and `infidelity.py` are now the same as in the master branch, so there won't be any merging conflicts do those changes.

ProGamerGov:
I looked into the FFTImage size thing, and there is a clear deviation in Ludwig's first Captum iteration of FFTImage and the Lucid code.


The Captum code create the image with the shape given from the `size` variable, and then this same shape is used to create the scale / frequency tensor:

```
        coeffs_shape = (channels, size[0], size[1] // 2 + 1, 2)
        random_coeffs = torch.randn(
            coeffs_shape
        )  # names=["C", "H_f", "W_f", "complex"]
        self.fourier_coeffs = nn.Parameter(random_coeffs / 50)

        frequencies = FFTImage.rfft2d_freqs(*size)
```

https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R160-R166

While Lucid derives the shape of the image from the shape of the frequency tensor (that the if statement changes with an odd width value):

```
    batch, h, w, ch = shape
    freqs = rfft2d_freqs(h, w)
    init_val_size = (2, batch, ch) + freqs.shape
```
https://github.com/tensorflow/lucid/blob/master/lucid/optvis/param/spatial.py#L66-L67


Therefore, I think the special behavior for widths that have an odd value is required for the Lucid way of doing things and not the new Captum FFTImage. I think that Ludwig may have just forgotten to update `rfft2d_freqs` to match his changes in the `init` function.

NarineK:
> > > #3 'mat2' is
> > 
> > 
> > Interesting, according to code snippet it looks like it worked first time calling `opt.images.NaturalImage(init=torch.ones(3, 1, 1)).cuda()` but failed second time ?
> 
> The issue was that by creating the ToRGB instance in the init function, and it creates a single instance that will be passed to future constructor calls as default. It's apparently a well-known trap when using default parameters:
> 
> ```
> # Example code
> import torch
> import torch.nn as nn
> 
> class Test1(nn.Module):
>     def __init__(self):
>         super().__init__()
>         self.register_buffer("test_tensor", torch.ones(5))
> 
>     def forward(self, x):
>         return x * self.test_tensor
> 
> class Test2(nn.Module):
>     def __init__(self, x, test1 = Test1()):
>         super().__init__()
>         self.test1 = test1
>         x = self.test1(x)
> 
>     def forward(self, x):
>         return self.test1(x)
> 
> t = Test2(torch.ones(5)).cuda()
> t = Test2(torch.ones(5)).cuda() # Fails
> ```
> 
> ```
> t = Test2(torch.ones(5), Test1()).cuda()
> t = Test2(torch.ones(5), Test1()).cuda() # Works
> ```
> 
> > > code that Lucid uses
> > 
> > 
> > According to lucid implementation it looks like we need to add 2 for odd cases and later remove that one additional pixel. What was the original problem that you saw with line: `wadd = 2 if width % 2 == 1 else 1` ? What was failing ?
> 
> The original issue was that `fourier_coeffs` and `spectrum_scale` would have a size mistmatch if the width given to FFTImage either through the `size` argument, or the `init` tensor was an odd number.
> 
> This code here would result in the following error:
> 
> ```
> image = opt.images.FFTImage((512, 405))
> out = image()
> ```
> 
> ```
> RuntimeError                              Traceback (most recent call last)
> 
> <ipython-input-3-358d0813d735> in <module>()
>       1 image = opt.images.FFTImage((512, 405))
> ----> 2 out = image()
> 
> 1 frames
> 
> /content/captum/captum/optim/_param/image/images.py in forward(self)
>     183     def forward(self) -> torch.Tensor:
>     184         h, w = self.size
> --> 185         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
>     186         output = self.torch_irfft(scaled_spectrum)
>     187         return output.refine_names("B", "C", "H", "W")
> 
> RuntimeError: The size of tensor a (203) must match the size of tensor b (204) at non-singleton dimension 3
> ```
> 
> And the same issue occurred when the init tensor size had an odd width:
> 
> ```
> image = opt.images.FFTImage(init=torch.ones(1, 3, 512, 405))
> out = image()
> ```
> 
> ```
> RuntimeError                              Traceback (most recent call last)
> 
> <ipython-input-5-48fc824a1f55> in <module>()
> ----> 1 image = opt.images.FFTImage(init=torch.ones(1, 3, 512, 405))
>       2 out = image()
> 
> /content/captum/captum/optim/_param/image/images.py in __init__(self, size, channels, batch, init)
>     132             fourier_coeffs = random_coeffs / 50
>     133         else:
> --> 134             fourier_coeffs = self.torch_rfft(init) / spectrum_scale
>     135 
>     136         self.fourier_coeffs = nn.Parameter(fourier_coeffs)
> 
> RuntimeError: The size of tensor a (203) must match the size of tensor b (204) at non-singleton dimension 3
> ```
> 
> The Lucid code also talks about slicing the image tensor to resolve the issue, but in our case `spectrum_scale` is the bigger tensor and I'm not sure that we should be slicing it.

Thank you for the explanation, @ProGamerGov! Is there a specific reason why we chose to use buffer instead of instance field. If we use it as instance field it shouldn't cause any problems.
```
# Example code
import torch
import torch.nn as nn

class Test1(nn.Module):
    def __init__(self):
        super().__init__()
        self.test_tensor = torch.ones(5)
        print(self._buffers)

    def forward(self, x):
        return x * self.test_tensor

class Test2(nn.Module):
    def __init__(self, x, test1 = Test1()):
        super().__init__()
        self.test1 = test1
        x = self.test1(x)

    def forward(self, x):
        return self.test1(x)

t = Test2(torch.ones(5)).cuda()
t = Test2(torch.ones(5)).cuda() # Works
```

NarineK:
> @NarineK The lines I altered in `feature_ablation.py` and `infidelity.py` are now the same as in the master branch, so there won't be any merging conflicts caused by those changes.

If they are exactly the same then there won't be merge conflicts but we will have same changes in different commits. I think it is fine for now since those are minor changes but it is better to sync up with master for larger changes.

ProGamerGov:
@NarineK Ludwig's original Captum code had both `FFTImage` and `ToRGB` using buffers. I'm not sure why he chose to do it that way, but I can change it if you think that they should be set as `self` variables.


> If they are exactly the same then there won't be merge conflicts but we will have same changes in different commits. I think it is fine for now since those are minor changes but it is better to sync up with master for larger changes.

Yeah, we can sync up with the master branch in separate PR. I was just trying to avoid it in this PR so that the commit history wasn't filled with all the master commits, like what happened with SK's PR.

ProGamerGov:
Oh, I think I understand why buffers were used now. They make it so that when a `NaturalImage` instance is placed onto the GPU, the buffers are also placed on the GPU.


This code works when `spectrum_scale` (FFTImage) and `transform` (ToRGB) are buffers, but results in a device error when I just define them as `self` variables:

```
obj = opt.InputOptimization(
    model,
    opt.loss.ChannelActivation(model.mixed4a, 476),
    input_param=opt.images.NaturalImage((224,224)).to(device),
)
obj.optimize()

obj.input_param().show()
```

```
/content/captum/captum/optim/_param/image/images.py in forward(self)
    180 
    181     def forward(self) -> torch.Tensor:
--> 182         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
    183         output = self.torch_irfft(scaled_spectrum)
    184         return output.refine_names("B", "C", "H", "W")

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

NarineK:
> I looked into the FFTImage size thing, and there is a clear deviation in Ludwig's first Captum iteration of FFTImage and the Lucid code.
> 
> The Captum code create the image with the shape given from the `size` variable, and then this same shape is used to create the scale / frequency tensor:
> 
> ```
>         coeffs_shape = (channels, size[0], size[1] // 2 + 1, 2)
>         random_coeffs = torch.randn(
>             coeffs_shape
>         )  # names=["C", "H_f", "W_f", "complex"]
>         self.fourier_coeffs = nn.Parameter(random_coeffs / 50)
> 
>         frequencies = FFTImage.rfft2d_freqs(*size)
> ```
> 
> * The second last dimension in `coeffs_shape` is always ` 2 + 1` and never  `2 + 2`, so making the `rfft2d_freqs` always use ` 2 + 1` prevents a size mismatch.
> 
> https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R160-R166
> 
> While Lucid derives the shape of the image from the shape of the frequency tensor (that the if statement changes with an odd width value):
> 
> ```
>     batch, h, w, ch = shape
>     freqs = rfft2d_freqs(h, w)
>     init_val_size = (2, batch, ch) + freqs.shape
> ```
> 
> https://github.com/tensorflow/lucid/blob/master/lucid/optvis/param/spatial.py#L66-L67
> 
> Therefore, I think the special behavior for widths that have an odd value is required for the Lucid way of doing things and not the new Captum FFTImage. I think that Ludwig may have just forgotten to update `rfft2d_freqs` to match his changes in the `init` function.

Thank you for looking deeper into this:
It looks like we are dividing by  2 here:
https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R160-R166
and also here:
https://github.com/pytorch/captum/pull/656/files#diff-d1326a272667e088fe9934dd175f0be589edf7594ee01b7463451a5266c56b47R141
Is it necessary to do it in booth places ?


> Oh, I think I understand why buffers were used now. They make it so that when a `NaturalImage` instance is placed onto the GPU, the buffers are also placed on the GPU.
> 
> This code works when `spectrum_scale` (`FFTImage`) and `transform` (`ToRGB`) are buffers, but results in a device error when I just define them as `self` variables:
> 
> ```
> obj = opt.InputOptimization(
>     model,
>     opt.loss.ChannelActivation(model.mixed4a, 476),
>     input_param=opt.images.NaturalImage((224,224)).to(device),
> )
> obj.optimize()
> 
> obj.input_param().show()
> ```
> 
> ```
> /content/captum/captum/optim/_param/image/images.py in forward(self)
>     180 
>     181     def forward(self) -> torch.Tensor:
> --> 182         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
>     183         output = self.torch_irfft(scaled_spectrum)
>     184         return output.refine_names("B", "C", "H", "W")
> 
> RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
> ```
> 
> I guess we could override `.to()` & `.cuda()` as a possible solution? Or we could override `_apply` so that both `.to()` and `.cuda()` works. Though I think that buffers are the recommend solution to this problem.



> Oh, I think I understand why buffers were used now. They make it so that when a `NaturalImage` instance is placed onto the GPU, the buffers are also placed on the GPU.
> 
> This code works when `spectrum_scale` (`FFTImage`) and `transform` (`ToRGB`) are buffers, but results in a device error when I just define them as `self` variables:
> 
> ```
> obj = opt.InputOptimization(
>     model,
>     opt.loss.ChannelActivation(model.mixed4a, 476),
>     input_param=opt.images.NaturalImage((224,224)).to(device),
> )
> obj.optimize()
> 
> obj.input_param().show()
> ```
> 
> ```
> /content/captum/captum/optim/_param/image/images.py in forward(self)
>     180 
>     181     def forward(self) -> torch.Tensor:
> --> 182         scaled_spectrum = self.fourier_coeffs * self.spectrum_scale
>     183         output = self.torch_irfft(scaled_spectrum)
>     184         return output.refine_names("B", "C", "H", "W")
> 
> RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
> ```
> 
> I guess we could override `.to()` & `.cuda()` as a possible solution? Or we could override `_apply` so that both `.to()` and `.cuda()` works. Though I think that buffers are the recommend solution to this problem.


I think that it is tricky to put a module on the device because pytorch puts actual tensors (parameters) related to that module on the device and not the module per se. For example we cannot  call `.device` on the module. And this makes it unclear what is on the device and what not.
https://discuss.pytorch.org/t/how-to-check-if-model-is-on-cuda/180

To be clear we can explicitly put the `transform` - ed tensors on the device where the actual input is before we multiply with it.

We can set: 
self.transform.to(x.device)

before using it and use instance fields instead. Would that work ?

NarineK:
> I looked into the FFTImage size thing, and there is a clear deviation in Ludwig's first Captum iteration of FFTImage and the Lucid code.
> 
> The Captum code create the image with the shape given from the `size` variable, and then this same shape is used to create the scale / frequency tensor:
> 
> ```
>         coeffs_shape = (channels, size[0], size[1] // 2 + 1, 2)
>         random_coeffs = torch.randn(
>             coeffs_shape
>         )  # names=["C", "H_f", "W_f", "complex"]
>         self.fourier_coeffs = nn.Parameter(random_coeffs / 50)
> 
>         frequencies = FFTImage.rfft2d_freqs(*size)
> ```
> 
> * The second last dimension in `coeffs_shape` is always ` 2 + 1` and never  `2 + 2`, so making the `rfft2d_freqs` always use ` 2 + 1` prevents a size mismatch.
> 
> https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R160-R166
> 
> While Lucid derives the shape of the image from the shape of the frequency tensor (that the if statement changes with an odd width value):
> 
> ```
>     batch, h, w, ch = shape
>     freqs = rfft2d_freqs(h, w)
>     init_val_size = (2, batch, ch) + freqs.shape
> ```
> 
> https://github.com/tensorflow/lucid/blob/master/lucid/optvis/param/spatial.py#L66-L67
> 
> Therefore, I think the special behavior for widths that have an odd value is required for the Lucid way of doing things and not the new Captum FFTImage. I think that Ludwig may have just forgotten to update `rfft2d_freqs` to match his changes in the `init` function.

Thank you for looking into this, @ProGamerGov !  As you mentioned for coeffs_shape Ludwig used `2 + 1` and that is the inconsistency. I don't know exactly why he did it. We can try to ask and understand it better. Either way, it would be good to document those differences in the code so that we don't forget.

In terms of results, not shapes, are we getting different outcomes ?

ProGamerGov:
@NarineK Yeah, moving the tensor to the input device for every forward call would work! But would that incur any sort of performance penalty for constantly moving the tensor to the target device? I tried testing it out, and there didn't seem to be any obvious effect to the time it takes to complete 512 iterations.

Alternatively, we could just include an additional parameter to `NaturalImage` to disable `ToRGB` or set the `decorrelation_module` to `SkipLayer`.


> Thank you for looking into this, @ProGamerGov ! As you mentioned for coeffs_shape Ludwig used 2 + 1 and that is the inconsistency. I don't know exactly why he did it. We can try to ask and understand it better. Either way, it would be good to document those differences in the code so that we don't forget.
> 
> In terms of results, not shapes, are we getting different outcomes ?
> 
> Also there is a division by 50 magic number (this isn't related to the dimensionality) that I don't see in lucid
> 
> https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R164).

Ludwig hasn't been actively lately, so we might have to wait a bit to ask him why he made certain design choices. Chris may be able to help, but he is a bit busy at the moment as well. Our version of `FFTImage` produces different results. We don't divide the output by 4, normalize the fft operations, and I think there's a small difference with what is done to the `scale` variable before it's saved as `spectrum_scale`.

I think that he intended Captum's `FFTImage` to be the successor to the Lucid's `FFTImage`, but I did create a Captum version of the old `FFTImage` a while back and found the results to be less detailed.

ProGamerGov:
> Thank you for looking deeper into this:
> It looks like we are dividing by 2 here:
> https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R160-R166
> and also here:
> https://github.com/pytorch/captum/pull/656/files#diff-d1326a272667e088fe9934dd175f0be589edf7594ee01b7463451a5266c56b47R141
> Is it necessary to do it in booth places ?

Yes, I think it's necessary as that way we get the right matching size for 5 dimensional fft tensors:

```
import torch
import torch.fft
x = torch.randn(1, 3, 512, 405)
print(torch.view_as_real(torch.fft.rfftn(x, s=(512, 405))).shape)
# outputs torch.Size([1, 3, 512, 203, 2])
# 405 // 2 + 1 = 203
```

NarineK:
> @NarineK Yeah, moving the tensor to the input device for every forward call would work! But would that incur any sort of performance penalty for constantly moving the tensor to the target device? I tried testing it out, and there didn't seem to be any obvious effect to the time it takes to complete 512 iterations.
> 
> Alternatively, we could just include an additional parameter to `NaturalImage` to disable `ToRGB` or set the `decorrelation_module` to `SkipLayer`.
> 
> I've been disabling the color decorrelation like this for experiments:
> 
> ```
> image = opt.images.NaturalImage(
>     parameterization=opt.images.PixelImage,
>     init=init_image.cpu(),
>     decorrelation_module=opt.models.SkipLayer(), # Disable color decorrelation
>     squash_func=lambda x: x, # Disable squash function
> ).to(device)
> ```
> 
> > Thank you for looking into this, @ProGamerGov ! As you mentioned for coeffs_shape Ludwig used 2 + 1 and that is the inconsistency. I don't know exactly why he did it. We can try to ask and understand it better. Either way, it would be good to document those differences in the code so that we don't forget.
> > In terms of results, not shapes, are we getting different outcomes ?
> > Also there is a division by 50 magic number (this isn't related to the dimensionality) that I don't see in lucid
> > https://github.com/pytorch/captum/pull/412/files#diff-d9ef468be9704729ff7c3bd65ad5b115e206b5f418039e18cb333e123428dde5R164).
> 
> Ludwig hasn't been actively lately, so we might have to wait a bit to ask him why he made certain design choices. Chris may be able to help, but he is busy at the moment as well I think. Our version of `FFTImage` produces different results that appear to be superior to Lucid's `FFTImage` results. We don't divide the output by 4, normalize the fft operations, and I think there's a small difference with what is done to the `scale` variable before it's saved as `spectrum_scale`.
> 
> I think that he intended Captum's `FFTImage` to be the successor to the Lucid's `FFTImage`, but I did create a Captum version of the old `FFTImage` a while back and found the results to be less detailed.

I don't know or think that there will be performance issues. It is simply linking the device. I don't think that `.to(device)` is costly. It is most probably a simple linking.

Alternatively, we could also pass device as an optional argument to the constructor. The user needs to specify explicitly, but we can specifically describe that the device is used for the transformation tensor.

ProGamerGov:
@NarineK I think that we could just do `self.decorrelate = decorrelation_module.cpu()` to ensure that the decorrelation_module is always on the CPU for the start, and then we can still create the `ToRGB` instance in the `__init__` function signature . The `init` tensor is already expected to be on the CPU, if it is not set to `None`. So, this might be a more elegant solution?

This way we would keep the `.to()` and `.cuda()` behavior of the buffers, and means that we don't have to change the code in `ToRGB`. The original intended behavior of the `decorrelation_module` where it can be set to None to disable color decorrelation would also be preserved.

ProGamerGov:
@NarineK I've implemented the `decorrelation_module.cpu()` fix and resolved the Conda test issue. I think that all of the issues raised in this PR have been resolved now, and thus it may be ready for merging?

NarineK:
> @NarineK I've implemented the `decorrelation_module.cpu()` fix and resolved the Conda test issue. I think that all of the issues raised in this PR have been resolved now, and thus it may be ready for merging?

This means that the tensor coming out of RGB transformation will always be on cpu. Since this is not critical we can leave it as is. I haven't reviewed the rest of PR yet. Let me have a quick look.

captum/optim/_param/image/images.py
```diff
@@ -428,11 +425,13 @@ def __init__(
         decorrelate_init: bool = True,
     ) -> None:
         super().__init__()
-        self.decorrelate = decorrelation_module
+        self.decorrelate = (
+            decorrelation_module.cpu() if decorrelation_module is not None else None
+        )
         if init is not None:
+            assert not init.is_cuda
```
NarineK, line 432: Enforcing the tensor to be on CPU without warning why, will make the user to think that there is a bug in the code.
We should explicitly put is on the CPU. I assume this is because of `decorrelation_module.cpu()`

I think that it would be good to print a warning message why we put `decorrelation_module` on the CPU and explain it in the documentation.

Perhaps also adding a TODO so that we can make the module more flexible and don't have to move to CPU device.

captum/optim/models/_common.py
```diff
@@ -202,7 +202,10 @@ class SkipLayer(torch.nn.Module):
     This layer is made to take the place of nonlinear activation layers like ReLU.
     """
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def __init__(self, *args, **kwargs) -> None:
```
NarineK, line 205: why are the args and kwargs passed if they are not used ? Are there any real scenarios that we need it ? Is this because we replace ReLU with SkipLayer ?
I could imagine, for example, that x is a tuple of tensors and we would need to return that tuple.

captum/optim/models/_common.py
```diff
@@ -202,7 +202,10 @@ class SkipLayer(torch.nn.Module):
...
+        super().__init__()
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
```
NarineK, line 208: why are the args and kwargs passed if they are not used ? We should document it because it is unclear why this is needed.

captum/optim/models/_common.py
ProGamerGov: I made the change because users may have models using `activ(inplace=False)` or `activ(False)` where `activ = torch.nn.ReLU`. This is same way that `torch.nn.Identity` works: https://pytorch.org/docs/stable/generated/torch.nn.Identity.html

I'll add the type hint for tuples of tensors.

captum/optim/models/_common.py
ProGamerGov: I made the change for situations like replacing decorrelation module as it specifies an additional argument in the forward pass. I'll add some documentation explaining why I added args and kwargs.

captum/optim/_param/image/images.py
ProGamerGov: The init tensor had to be put on the CPU because `FFTImage` expected it to be on the CPU, but I just fixed it so that `spectrum_scale` is now placed on to the init tensor's device.

In order to avoid any future issues with the decorrelation module being created in the function signature, I added a deepcopy line that only runs when it's a `ToRGB` instance.

I also removed the device assertion and made `ToRGB` place the transform on the input tensor's device as well, so that it becomes a no-op when the buffer is placed on the target device. 

The device issues should now be resolved I think?

ProGamerGov: I removed the deepcopy line as it was redundant.

captum/optim/models/_common.py
NarineK: yes, this will work for ReLU with one input tensor and specific cases like ReLUs but it will have issues with custom modules that take tuples of tensors and modules such as nn.Linear, etc. I think that perhaps it would be good to give more specific name to it or describe that this is a skip layer that assumes that the inputs and outputs have to have the same shape and that it is the first tensor that is returned as an output.

captum/optim/_param/image/images.py
NarineK: It looks cleaner now.

Here squash_func is not assigned, right? I wonder why we use different `squash_func`s depending on whether `init` is provided or not.
```
if squash_func is None: 
 def squash_func(x: torch.Tensor) -> torch.Tensor: 
   return x.clamp(0, 1)
```

NarineK (approved):
LGTM! Thank you for addressing the comments, @ProGamerGov!
  I left two minor comments.

captum/optim/_param/image/images.py
ProGamerGov: @NarineK The different squash function for when init tensors are provided was something I found to produce better results with images.

The TODO line is because named dimensions are not fully supported by PyTorch yet.

NarineK: interesting, why are inner functions better in this case ?
`rename` is used in line 440 too. I we can leave it but ideally we should use it for the PT versions that support it.

PR #659, opened 2021-04-28
WIP: Migrate to register_full_backward_hook

vivekmig:
Modifying all module backward hooks to utilize the new register_full_backward hook API documented [here](register_full_backward_hook). This new API resolves many issues we previously encountered with backward module hooks.

Since this API is available only in torch 1.8, allowing a fall-back option to the original backward hook approach.

### DIFF
--- a/captum/_utils/gradient.py
+++ b/captum/_utils/gradient.py
@@ -85,6 +85,17 @@ def undo_gradient_requirements(
             input.requires_grad_(False)
 
 
+def register_backward_hook(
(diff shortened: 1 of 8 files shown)


vivekmig:
Closing PR, will revisit this after torch 1.9 releases. Differences in hook behavior for intermediate activations inhibit this migration currently (reported as an issue for core PyToch https://github.com/pytorch/pytorch/issues/57157 ), will be able to migrate once resolved in the next issue.

PR #662, opened 2021-04-29
Fix LRP Support for Dropout Modules

vivekmig:
This fixes LRP to work appropriately with nn.Dropout modules, since dropout backward hooks add an additional grad_input element of None which is not supported with the current backward_hook_activation.

### DIFF
--- a/captum/attr/_core/feature_ablation.py
+++ b/captum/attr/_core/feature_ablation.py
@@ -554,7 +554,7 @@ class FeatureAblation(PerturbationAttribution):
         )
 
     def _get_feature_counts(self, inputs, feature_mask, **kwargs):
-        """ return the numbers of input features """
+        """return the numbers of input features"""
         if not feature_mask:
             return tuple(inp[0].numel() if inp.numel() else 0 for inp in inputs)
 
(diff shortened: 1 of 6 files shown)


NarineK (approved):
Looks great thank you for the fix. It can be that the behavior of backward hook will change in the future releases of PT. We might want to keep eye on it for future releases. 
https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.register_backward_hook
If can be that `register_full_backward_hook` won't have that issue any more.

vivekmig:
> Looks great, thank you for the fix. It can be that the behavior of backward hook will change in the future releases of PT. We might want to keep eye on it for future releases.
> https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.register_backward_hook
> If can be that `register_full_backward_hook` won't have that issue any more.

Yes, you're right, register_full_backward_hook doesn't have the issue anymore, I tested this initially. I wanted to just switch to register_full_backward_hook and avoid adding this special case, but unfortunately because of the issue identified in #659, I postponed that migration and added this condition for now.

PR #666, opened 2021-05-03
Fixes for Boolean Input Tensors

vivekmig:
This adds support for boolean input tensors to perturbation-based methods (Shapley Value Sampling and Lime) with corresponding unit tests.

### DIFF
--- a/captum/attr/_core/lime.py
+++ b/captum/attr/_core/lime.py
@@ -571,15 +571,18 @@ def default_from_interp_rep_transform(curr_sample, original_inputs, **kwargs):
     ), "Must provide baselines to use default interpretable representation transfrom"
     feature_mask = kwargs["feature_mask"]
     if isinstance(feature_mask, Tensor):
-        binary_mask = curr_sample[0][feature_mask].to(original_inputs.dtype)
-        return binary_mask * original_inputs + (1 - binary_mask) * kwargs["baselines"]
+        binary_mask = curr_sample[0][feature_mask].bool()
+        return (
+            binary_mask.to(original_inputs.dtype) * original_inputs
+            + (~binary_mask).to(original_inputs.dtype) * kwargs["baselines"]
+        )
     else:
         binary_mask = tuple(
-            curr_sample[0][feature_mask[j]] for j in range(len(feature_mask))
+            curr_sample[0][feature_mask[j]].bool() for j in range(len(feature_mask))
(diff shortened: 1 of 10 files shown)


captum/attr/_core/lime.py
```diff
@@ -571,15 +571,18 @@ def default_from_interp_rep_transform(curr_sample, original_inputs, **kwargs):
...
-        return binary_mask * original_inputs + (1 - binary_mask) * kwargs["baselines"]
+        binary_mask = curr_sample[0][feature_mask].bool()
+        return (
+            binary_mask.to(original_inputs.dtype) * original_inputs
+            + (~binary_mask).to(original_inputs.dtype) * kwargs["baselines"]
+        )
     else:
         binary_mask = tuple(
-            curr_sample[0][feature_mask[j]] for j in range(len(feature_mask))
+            curr_sample[0][feature_mask[j]].bool() for j in range(len(feature_mask))
```
NarineK, line 581: Is this change related to inputs being boolean or is it so that we can use `~` operator ?

tests/attr/test_shapley.py
```diff
@@ -39,6 +40,29 @@ def test_simple_shapley_sampling_with_mask(self) -> None:
...
+        self._shapley_test_assert(
+            net,
+            inp,
+            [35.0, 35.0, 35.0],
+            feature_mask=torch.tensor([[0, 0, 1]]),
+            perturbations_per_eval=(1, 2, 3),
+        )
+
+    def test_simple_shapley_sampling_boolean_with_baseline(self) -> None:
+        net = BasicModelBoolInput()
```
NarineK, line 55: Another way of testing would be using `BasicModel_MultiLayer()` with the transformed input and seeing if we get the same result but if we manually verified expected value, then it looks good to me.

NarineK (approved):
LGTM! Thank you for the fix and test cases!

captum/attr/_core/lime.py
vivekmig: Yes, this is necessary since if this mask is an int / float, multiplying by a boolean still casts product to int / float causing the masked input to not be a boolean.

tests/attr/test_shapley.py
vivekmig: Yes, I double checked the values manually, they seem correct.

Issue #664, opened 2021-05-03
Addition of a LRP tutorial notebook


JohannesK14:
## 📚 Documentation

I would like to add an LRP tutorial/ notebook to help other people with LRP.

I suggest using this tutorial as a reference: [Python LRP tutorial](https://git.tu-berlin.de/gmontavon/lrp-tutorial)
I already wrote the code using captum and was able to produce visually similar results.

bilalsal:
Thank you very much @JohannesK14 for the initiative!

An LRP tutorial is indeed due. How about your submit a PR of your tutorial (ideally as a Jupyter notebook) under https://github.com/pytorch/captum/tree/master/tutorials?

We can then test the notebook and provide any needed feedback.

Thanks in advance!

JohannesK14:
Ok.

NaokoGit:
Hi @JohannesK14 

Thanks for adding LRP tutorial notebook. I'm just executing your ongoing tutorial notebook but I am wondering why the output result (Overlayed LRP) is different from the original one (Python LRP tutorial)? How can I get summed over actual pixel-wise contributions using captum?

craymichael:
Closing because #1105 superseded this issue and landed the LRP tutorial.

The published tutorial is available here: https://captum.ai/tutorials/TorchVision_Interpret

PR #668, opened 2021-05-10
Lrp tutorial added

JohannesK14:
#### Reference Issues
[664](https://github.com/pytorch/captum/issues/664)

#### Changes
Addition of an LRP tutorial and a sample image. The tutorial replicates the results of the basic [PyTorch LRP implementation](https://git.tu-berlin.de/gmontavon/lrp-tutorial) provided by the LRP authors using the LRP class of captum.


NarineK:
Thank you for the tutorial, @JohannesK14! We will provide review comments soon!

A quick question: Where is the `castle.jpeg` from ? We need to know the copyright permissions and the attribution required for that image.

NarineK:
Thank you very much for working on this tutorial, @JohannesK14 ! Overall looks great to me. 

One ask: 
Since we are adding more tutorials to Captum I'm thinking that perhaps we can combine this with the resnet one and call it `Resnet_VGG_TorchVision_Interpret.ipynb`, if we want to showcase LRP example explicitly for VGG. The content of this tutorial can be added to the bottom of the `Resnet_VGG_TorchVision_Interpret.ipynb` file. 
In the top of the`Resnet_VGG_TorchVision_Interpret` file we can briefly describe that we are showcasing LRP for VGG as well. 

Feel free to add an example of LRP for Resnet as well in that same notebook.
https://github.com/pytorch/captum/blob/master/tutorials/Resnet_VGG_TorchVision_Interpret.ipynb

Couple more minor comments:
 "As one see" -> "As one can see" ?
"Currently implemented in captum are " -> "Currently implemented rules in captum are " ?   Also, do you mind adding here a link to the documentation for the recommended rules ?

NarineK (approved):
Thank you very much for working on this tutorial, @JohannesK14 ! Overall looks great to me.

One ask:
Since we are adding more tutorials to Captum I'm thinking that perhaps we can combine this with the resnet one and call it Resnet_VGG_TorchVision_Interpret.ipynb, if we want to showcase LRP example explicitly for VGG. The content of this tutorial can be added to the bottom of the Resnet_VGG_TorchVision_Interpret.ipynb file.
In the top of theResnet_VGG_TorchVision_Interpret file we can briefly describe that we are showcasing LRP for VGG as well.

Feel free to add an example of LRP for Resnet as well in that same notebook.
https://github.com/pytorch/captum/blob/master/tutorials/Resnet_VGG_TorchVision_Interpret.ipynb

If you combine those tutorials, I'd also suggest that you do minor adjustments for the website text as well:
https://github.com/pytorch/captum/blob/master/website/tutorials.json#L17
https://github.com/pytorch/captum/blob/master/website/pages/tutorials/index.js#L70

Couple more minor comments:
 "As one see" -> "As one can see" ?
"Currently implemented in captum are " -> "Currently implemented rules in captum are " ?   Also, do you mind adding here a link to the documentation for the recommended rules ?

JohannesK14:
> Thank you for the tutorial, @JohannesK14! We will provide review comments soon!
> 
> A quick question: Where is the `castle.jpeg` from ? We need to know the copyright permissions and the attribution required for that image.

Hello @NarineK ,

I took the image from here: https://git.tu-berlin.de/gmontavon/lrp-tutorial 

I could ask the author for the copyright permission.

JohannesK14:
Regarding your second comment:

Of course, I can add the tutorial to the Resnet_TorchVision_Interpret.ipynb. But then I would suggest omitting VGG and just use Resnet and also the swan picture. Otherwise, it will look a little bit thrown together, I think. And Resnet is also more interesting than VGG, because there you need the implementation with back hooks for the skip connections.

And you would not need copyright permission for the castle anymore.

NarineK:
> Regarding your second comment:
> 
> Of course, I can add the tutorial to the Resnet_TorchVision_Interpret.ipynb. But then I would suggest omitting VGG and just use Resnet and also the swan picture. Otherwise, it will look a little bit thrown together, I think. And Resnet is also more interesting than VGG, because there you need the implementation with back hooks for the skip connections.
> 
> And you would not need copyright permission for the castle anymore.
> 
> I will have a look into that in the next few days.

Hi @JohannesK14, agree, let's use Resnet and the swan image instead.

NarineK:
@JohannesK14 would you have time to address the comments in this PR as discussed previously ?

HeinrichAD:
Unfortunately, the example notebook does not work anymore.

image

`RuntimeError: Output 0 of BackwardHookFunctionBackward is a view and is being modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can fix this by cloning the output of the custom Function.`

---

I like the example. Does anybody has an idea how to solve the issue?
I mean, since it's a pre-trained model, we can't just change the inline relu layer to a "normal" relu layer, for example.

JohannesK14:
@HeinrichAD I just set the inplace flag to false. Why do you see a problem there? To my knowledge, the inplace functionality could save some memory, but should deliver the same results in a VGG.

@NarineK Sorry for my long absence. I tried the switch to the ResNet architecture this morning. Unfortunately, the design of the composition of the rules for the ResNet18 architecture is more complicated than I thought it would be. I don't have the time to tinker with the rules to create a acceptable result for the visualization of the attributions, which I am very sorry for. Therefore, I switched to the picture of the swan, which also gives a pleasant result and avoids copyright issues.

Do you think I should still merge it with the ResNet tutorial as you suggested earlier?

HeinrichAD:
@JohannesK14 I looked into the source code [1]. In this case it will be the same. So no problems here.

[1] https://pytorch.org/vision/main/_modules/torchvision/models/vgg.html#vgg16

craymichael:
Closing because this PR was superseded by #1105, which built on this work and landed the LRP tutorial.

The published tutorial is available here: https://captum.ai/tutorials/TorchVision_Interpret

Thank you for the original contribution.

PR #675, opened 2021-05-26
Fix failing nodejs

NarineK:
Creating separate PR so that we can merge it quicker and don't have to wait for #672 to be reviewed.


vivekmig (approved):

PR #691, opened 2021-06-24
Moving initial captum.robust to OSS

vivekmig:
Summary: Move initial captum.robust attacks (FGSM, PGD, base classes) to open source folder

Reviewed By: aobo-y

Differential Revision: D29353978

### DIFF
--- a/captum/robust/__init__.py
+++ b/captum/robust/__init__.py
@@ -1,0 +1,5 @@
+#!/usr/bin/env python3
+
+from captum.robust._core.fgsm import FGSM  # noqa
+from captum.robust._core.perturbation import Perturbation  # noqa
+from captum.robust._core.pgd import PGD  # noqa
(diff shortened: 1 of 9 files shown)


captum/robust/_core/attack.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+from typing import Callable
+
+from captum.robust.perturbation import Perturbation
+
+
+class Attack(Perturbation):
```
NarineK, line 7: Perturbation and Attack look so similar. Maybe we can just have Attack (it already has perturb method). Is `Perturbation` class necessary ?

vivekmig: Yes, this is a good point. I originally only had Attack, but since we were generalizing to random perturbations which were independent of the model, I added the base Perturbation class to address the review comments and made it to not require the model in the constructor (while attack implies a white or black box attack which takes the model forward). I think to keep the naming generic, we could just keep the Perturbation base class and remove the Attack class, what do you think?

NarineK: That sounds good to me too. That will make things simpler.

PR #695, opened 2021-06-29
add section of sampling process in Lime tutorial

aobo-y:
Based on the discussion in https://github.com/pytorch/captum/pull/681#pullrequestreview-688647807 ,
- add trained  checkpoint of the embedding-bag model and load it by default
- add a new section to dive into the internal Lime sampling process to illustrate how the perturbed interpretable sample, transformed image and similarity are created and used.


vivekmig (approved):
Looks great, thanks for adding this 👍 ! Just a few nits:

nit: segmebnts -> segments
nit: “expeonential kernel” -> “exponential kernel”
nit: It may be worth mentioning that the assumptions mentioned in the section can be avoided by using LimeBase which provides flexibility of sampling in input space, etc., to ensure this doesn’t confuse users who want this flexibility 

Since we only use the new score for the original argmax class (rather than the new argmax class) in training the Lime interpretable model, would it be better to show that after running the forward pass rather than the Label of the perturbed sample?

aobo-y:
@vivekmig Thank you for the great catch of the sample label. I have made updates accordingly.

PR #714, opened 2021-07-09
replace np with torch linsapce in riemann approximation

aobo-y:
Reduce the dependency on numpy in approximation methods
- use `torch.linspace` in riemann methods

### DIFF
--- a/captum/attr/_utils/approximation_methods.py
+++ b/captum/attr/_utils/approximation_methods.py
@@ -4,11 +4,6 @@ from typing import Callable, List, Tuple
 
 import torch
 
-try:
-    import numpy as np
-except ImportError:
-    np = None
-
 
 class Riemann(Enum):
     left = 1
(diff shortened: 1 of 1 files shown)


captum/attr/_utils/approximation_methods.py
```diff
@@ -2,7 +2,12 @@
 from enum import Enum
 from typing import Callable, List, Tuple
 
-import numpy as np
+import torch
+
+try:
+    import numpy as np
```
NarineK, line 8: To make this more clean can we move the import into `gauss_legendre_builders`. Then we won't need try-except.

https://github.com/pytorch/captum/blob/6eb4aff67aa5a4b749b917b2ff191bb7e50afbd5/captum/attr/_utils/approximation_methods.py#L108

aobo-y: But as you can see below, I also rely on this to raise an explicit error for using `gausslegendre` when `numpy` is not available and suggesting `riemann`.  Without it, users will only receive an contextless message like `Cannot find module numpy`.

I know we are doing this for a special use case and it is unlikely to have any users encounter this. So I am also open to your suggestion.

NarineK: My worry is, since this code is going to be executed, it could make problems during packaging. If we put it in the function, the import is not going to be executed at all when we choose riemann sum. If numpy is not available an error will raise only if the user chooses gauss legendre. If you want, you can keep try-except in the `gauss_legendre_builders ` function.

aobo-y: Sure, I will move it.

NarineK: Thank you :)

NarineK (approved):
LGTM! Thank you very much for the fix, @aobo-y!

PR #718, opened 2021-07-12
fix argument type in _get_stat of Stat

aobo-y:
as title


aobo-y:
closed this fix since it need python3.7 while we still support lower version 
[ref](https://stackoverflow.com/a/33533514)

Issue #723, opened 2021-07-20
Multimodal input binary classifier with Saliency


NullCodex:
## ❓ Questions and Help

Hi Everyone,

# Question:
How can I apply saliency to a dataset composed of categorical and image data?

I am somewhat of a beginner with pytorch and the available resources are just not clicking with my use case. The ultimate goal is for me to plot the saliency of a model, but I am stuck on calculating the gradient. Any help or guidance would be much appreciated.

# What I've reviewed:
Multimodal_VQA_Captum_Insights tutorial
BERT tutorials
(These resources all have very different data structures(images/sentences) and are confusing for a beginner to translate to an easier image/categorical dataset)

# My issue
## Model
```
Model(
  (label_embedding): Embedding(10, 10)
  (model): Sequential(
    (0): Linear(in_features=1034, out_features=512, bias=True)
    (1): LeakyReLU(negative_slope=0.2, inplace=True)
    (2): Linear(in_features=512, out_features=512, bias=True)
    (3): Dropout(p=0.4, inplace=False)
    (4): LeakyReLU(negative_slope=0.2, inplace=True)
    (5): Linear(in_features=512, out_features=512, bias=True)
    (6): Dropout(p=0.4, inplace=False)
    (7): LeakyReLU(negative_slope=0.2, inplace=True)
    (8): Linear(in_features=512, out_features=1, bias=True)
  )
)
```
## Categorical
The categorical data is just a normal label such as 4

# What I tried
```
saliency = Saliency(model)
grads = saliency.attribute((input, label), target=None)
```

where input is just a (1, 32, 32) image. I set target=None since it's a binary classifier.

# Failure output
`One of the differentiated Tensors does not require grad`

Since the label is not a float it does not require grad, is there a way i can use the saliency method to capture the grads?

bilalsal:
Hi @NullCodex 

did you try `label.requires_grad = True`?

NullCodex:
Yes, but since label is not a float, it says it does not require grad.

NullCodex:
The label data is for instance `tensor([2])`.

bilalsal:
I see. Maybe to narrow down the issue, did you try other Captum algorithms? Like IntegratedGradients?

NullCodex:
I am pretty sure it's because captum doesn't support grads with integers. I tried other algorithms as well and none of them works. Are there any working examples where `tensor([2])` generates grads or require grads.

NullCodex:
[...]

PR #727, opened 2021-07-25
Enhance error message for missing rule

nanohanno:
This PR provides additional information to an error message that is triggered when neither an explicit LRP rule is defined nor a default rule exists. The current error message is quite short and does not give enough information to the user on how to solve the problem as discussed in issue #723 .

Feel free to propose a different wording.


NarineK (approved):
LGTM! Thanks for the fix, @nanohanno !

Issue #686, opened 2021-06-20
Receiving empty output in captum insights eventhough insights flask server shows request success in log


KasirajanA:
## ❓ Questions and Help

Hello,

Ran the [example.py](https://github.com/pytorch/captum/blob/bf8b15e1fb8bdc7b6e226d656b6a0c4d4825993c/captum/insights/example.py) after making below mentioned changes to formatted_data_iter function.

- Reduced batch size to 1
- Updated the function to only yield 4 results

```# Sample code
def formatted_data_iter():
    dataset = torchvision.datasets.CIFAR10(
        root="data/test", train=False, download=True, transform=transforms.ToTensor()
    )
    dataloader = iter(
        torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=False, num_workers=1)
    )
    for i in range(4):
        images, labels = next(dataloader)
        yield Batch(inputs=images, labels=labels)
```

Only for first fetch request getting output in captum insights. From second fetch request getting empty page in captum insights eventhough getting http response with status code 200.

**Observation:**
For each fetch request in captum insights UI, requested number of samples is fetched from the data iter [(Ref in source)](https://github.com/pytorch/captum/blob/bf8b15e1fb8bdc7b6e226d656b6a0c4d4825993c/captum/insights/attr_vis/app.py#L442) Since the data iter is not reinitialised,  it is exploited when multiple fetch requests are made and it throws StopIteration exception. But the exception is caught and neglected [(Ref in source)](https://github.com/pytorch/captum/blob/bf8b15e1fb8bdc7b6e226d656b6a0c4d4825993c/captum/insights/attr_vis/app.py#L470). Because of which there is no data to process and we are getting empty response in captum insights. 

Is this a expected behaviour or a bug?

Attached the modified example.py for reference.
[example.zip](https://github.com/pytorch/captum/files/6682459/example.zip)

NarineK:
cc: @Reubend

Reubend:
Hi there @KasirajanA thanks for reporting this and my apologies for the slow reply. This is definitely a bug and I've just sent out #728 to address it.

Reubend:
This should now be fixed by #728 
Let me know if there are any further issues.

PR #728, opened 2021-08-03
Fix empty data loading in Insights

Reubend:
Previously, when Captum Insights had already shown all existing batches from the dataset, it would stop showing data at all after fetches. This change makes it recycle some batches, so that the user can keep clicking on "Fetch Data" to see the changes that their setting have on the results.

This was reported in #686

### DIFF
--- a/captum/insights/attr_vis/app.py
+++ b/captum/insights/attr_vis/app.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 from collections import namedtuple
+from itertools import cycle
 from typing import (
     Any,
     Callable,
@@ -199,6 +200,7 @@ class AttributionVisualizer:
         self._outputs: List[VisualizationOutput] = []
         self._config = FilterConfig(prediction="all", classes=[], num_examples=4)
         self._dataset_iter = iter(dataset)
+        self._dataset_cache: List[Batch] = []
 
     def _calculate_attribution_from_cache(
         self, input_index: int, model_index: int, target: Optional[Tensor]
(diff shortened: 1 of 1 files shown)


captum/insights/attr_vis/app.py
```diff
@@ -439,7 +441,22 @@ def _calculate_vis_output(
         return results if results else None
 
     def _get_outputs(self) -> List[Tuple[List[VisualizationOutput], SampleCache]]:
-        batch_data = next(self._dataset_iter)
+        # If we run out of new betches, then we need to
```
NarineK, line 444: betches -> batches ?

NarineK (approved):

NarineK:
LGTM! @bilalsal, @edward-io do you have any comments on this issue ?

captum/insights/attr_vis/app.py
```diff
@@ -199,6 +200,7 @@ class scores.
         self._outputs: List[VisualizationOutput] = []
         self._config = FilterConfig(prediction="all", classes=[], num_examples=4)
         self._dataset_iter = iter(dataset)
```
miguelmartin75, line 202: wouldn't your PR be simpler to construct this as `cycle(iter(dataset))` since it seems you're effectively reimplementing this?

Reubend: It would be much simpler that way, but my concern is about memory usage. Per the [docs](https://docs.python.org/3/library/itertools.html#itertools.cycle), it seems that cycle
> may require significant auxiliary storage (depending on the length of the iterable)

Which suggests to me that it stores the entire thing in an array and then loops around. If the dataset is huge, then I think this could be an issue, because every batch would get stored in memory. That's why I made the cache only store a few batches instead, rather than using a cycle over the entire dataset. 

However, if you don't think the memory is an issue, I could do it this way instead. It would probably only be an issue if somebody requested a _lot_ of attributions.

miguelmartin75: Makes sense. Thanks for clarifying. Probably best to assume memory will be an issue.

edward-io (approved):
Looks good to me!

PR #743, opened 2021-08-30
fix the usage of 2 baseline components

bilalsal:
In construct_whole_bert_embeddings() we should be using:
- ref_token_type_ids instead of  token_type_ids
- ref_position_ids and position_ids


NarineK (approved):
LGTM! Thank you for fixing it, @bilalsal 1

PR #747, opened 2021-09-01
change TCAV to use captum/_utils/fb/av.py

99warriors:
Summary:
AV.load now returns a Dataset, not a DataLoader
Also, activations for a single layer and multiple concepts are no longer stored as a single DataLoader (which also generates labels).  Instead, they are stored as a list of Datasets (not DataLoaders).  This list is used to create a Dataset (not DataLoader) that also generates labels.  I think it is better to pass around Datasets for as long as possible, because they are more flexible.

Differential Revision: D30566560

### DIFF
--- a/captum/_utils/av.py
+++ b/captum/_utils/av.py
@@ -1,0 +1,504 @@
+#!/usr/bin/env python3
+
+import glob
+import os
+import re
+import warnings
+from typing import Any, List, Optional, Tuple, Union
+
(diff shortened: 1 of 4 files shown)


NarineK:
@99warriors, it looks like some of the tests and linter failed. Do you mind looking into them ?

NarineK:
Closing since the changes have been merged

PR #775, opened 2021-09-25
Fix flaky tests

crawlingcub:
I ran captum's test-suite several times by removing all seed setting statements such as `np.random_seed(1234)`.  This revealed some tests which are extremely flaky. For instance, the test `Test::test_multi_sample_shapley_batch_scalar_float` in `tests/attr/test_shapley.py` failed 272 out of 500 times that I ran. Similarly, other 3 tests  `Test::test_multi_sample_shapley_batch_scalar_*` which execute the same assertion also have a similar failure rate. 

To fix these tests, I tried different values of `n_samples` in the `_single_input_multi_sample_batch_scalar_shapley_assert` function which is used by all these tests. Using a value of `2500` seems to ensure that all these tests pass (i tried running the tests up to 5000 times) without seeds. This change will ensure that the tests never fail if the seeds/rng implementation changes. Also, the change in execution time of the tests is only ~1s. 

Please let me know if this change seems reasonable. If yes, I can also look into some other tests which are flaky as well. I will also be happy to incorporate any other suggestions that you may have. Thanks!

### DIFF
--- a/tests/attr/test_shapley.py
+++ b/tests/attr/test_shapley.py
@@ -333,7 +333,7 @@ class Test(BaseTest):
             feature_mask=mask,
             perturbations_per_eval=(1,),
             target=None,
-            n_samples=2500
+            n_samples=2500,
         )
 
     def _single_int_input_multi_sample_batch_scalar_shapley_assert(
--- a/tests/attr/test_shapley.py
+++ b/tests/attr/test_shapley.py
@@ -333,7 +333,7 @@ class Test(BaseTest):
             feature_mask=mask,
             perturbations_per_eval=(1,),
             target=None,
-            n_samples = 2500
+            n_samples=2500
         )
 
     def _single_int_input_multi_sample_batch_scalar_shapley_assert(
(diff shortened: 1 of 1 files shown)


tests/attr/test_shapley.py
```diff
@@ -333,6 +333,7 @@ def _single_input_multi_sample_batch_scalar_shapley_assert(
             feature_mask=mask,
             perturbations_per_eval=(1,),
             target=None,
+            n_samples = 2500
```
vivekmig, line 336: Looks like there's a flake8 line issue in the CircleCI results, would you be able to fix this?

vivekmig (approved):
Thanks so much for this fix @crawlingcub ! This fix looks good, it should be fine to increase n_samples for stability as long as runtime doesn't increase too much. For other occurrences, it may be better to alternatively increase the threshold used for comparison rather than increase n_samples if runtime is affected more substantially.

crawlingcub:
Thanks @vivekmig. I fixed the linting issue. I will also look into the other tests after this!

crawlingcub:
@vivekmig seems like the checks passed. Can you please merge the changes? Let me know if anything else is needed.

PR #805, opened 2021-11-30
change AV.load to not return None

99warriors:
Summary:
`captum/_utils/av.AV.load` previous returned `None` when AV was not found.  This made type checking difficult.

This diff changes so that a `RuntimeError` is raised if the AV is not found.

tests.utils.fb._core.test_av is changed to reflect this new convention

The only file that directly calls `AV.load` is `tcav.py` - all other files only call `AV.generate_dataset_activations`, which then calls `AV.load`.

Reviewed By: NarineK

Differential Revision: D32102878


NarineK:
@99warriors, let's look into failing test cases

PR #806, opened 2021-12-01
apply katex to render math in api docs

aobo-y:
Adopt the same `katex` package as [PyTorch](https://github.com/pytorch/pytorch/blob/4a8f27445d8f1905b7ecd27fe7fb076cad0ec0d3/docs/source/conf.py#L57) to support rendering math equations in API docs

Updated [captum.metrics.infidelity](http://captum.ai/api/metrics.html#infidelity) as an example
Screen Shot 2021-11-30 at 10 10 10 PM


NarineK (approved):
LGTM! Thank you for fixing this, @aobo-y!

PR #807, opened 2021-12-01
fix broken links in the doc of GradientShap

aobo-y:
sphinx does not allow breaking url into multiple lines. This pr simply fixes the broken links

### DIFF
--- a/captum/attr/_core/gradient_shap.py
+++ b/captum/attr/_core/gradient_shap.py
@@ -25,14 +25,11 @@ from captum.log import log_usage
 class GradientShap(GradientAttribution):
     r"""
     Implements gradient SHAP based on the implementation from SHAP's primary
-    author. For reference, please, view:
-
-    https://github.com/slundberg/shap\
-    #deep-learning-example-with-gradientexplainer-tensorflowkeraspytorch-models
-
-    A Unified Approach to Interpreting Model Predictions
-    http://papers.nips.cc/paper\
-    7062-a-unified-approach-to-interpreting-model-predictions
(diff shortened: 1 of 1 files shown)


captum/attr/_core/gradient_shap.py
```diff
@@ -25,14 +25,11 @@
...
-    #deep-learning-example-with-gradientexplainer-tensorflowkeraspytorch-models
-
-    A Unified Approach to Interpreting Model Predictions
-    http://papers.nips.cc/paper\
-    7062-a-unified-approach-to-interpreting-model-predictions
+    author. For reference, please view the original
+    `implementation
+    <https://github.com/slundberg/shap#deep-learning-example-with-gradientexplainer-tensorflowkeraspytorch-models>`_
+    and the paper: `A Unified Approach to Interpreting Model Predictions
+    <https://papers.nips.cc/paper/7062-a-unified-approach-to-interpreting-model-predictions>`_
```
NarineK, line 32: Do we need <http: ... >  for breaking lines into next line ?
Otherwise we don't need <http: ... >  ?

https://github.com/pytorch/captum/blob/master/captum/attr/_core/integrated_gradients.py#L44

aobo-y: Sorry for the confusion. It is not for breaking lines. 
\`xxx \<http:...\>\`_ in sphinx is equivalent to \[xxx\]\(http:....\) in markdown. It is to make hyperlink with text.
Even with \<..\>, sphinx still not allow breaking the url into mult-lines

PR #808, opened 2021-12-02
build sphinx from src code

aobo-y:
The module path in sphinx config is wrong. It is set to `~/captum/captum` but it should be `~/captum/` so all `.rst` files can correctly access the class through full module path, e.g., `.. autoclass:: captum.concept.TCAV`.

The site can be successfully built so far was because the src code was installed as a module in ci https://github.com/pytorch/captum/blob/master/scripts/install_via_pip.sh#L46 So sphinx actually use `captum` from installed module instead of the src code. This is very inconvenient for local development of docs and may potentially go wrong as another captum is installed in the current env.

The fix force the sphinx to use captum from src code.


NarineK (approved):
LGTM! Thank you!

PR #812, opened 2021-12-04
support empty inputs in Lime & KernelShap

aobo-y:
Empty inputs work in methods like `FeatureAblation` & `ShapleyValueSampling`, but not in `Lime` and `KernelShap`.
This update correctly handles empty tensors like `tensor([[],[],[]])` whose shape is (3,0) in `Lime` and `KernelShap`.
Added corresponding tests.


vivekmig (approved):

PR #820, opened 2021-12-12
Optim-wip: Fix & Update WeightVisualization Notebook

ProGamerGov:
This notebook was originally created with an early version of the optim module, so it was nonfunctional and required updating.

GitHub's diff checker is reporting that the entire file changed, but I only changed a few lines of code.

* Fixed the WeightVisualization notebook so that it works with the latest version of the optim module.

* Updated the WeightVisualization notebook to use loss comprehension for faster rendering times.


ProGamerGov:
@NarineK This PR and https://github.com/pytorch/captum/pull/822 should be simple to review as they only contain a small number of fixes to existing tutorial notebooks.

ProGamerGov:
As Git wasn't able to properly show the changes, I'll list them below (about 90 lines were changed):

* The `show` function and it's associated imports were removed as the function is part of the optim code library.
* Transforms were missing the 's' in their calls, as previously they were accessed via `transform.ExampleTransform`, instead of `transforms.ExampleTransform`.
* The visualization functions were updated to use batch targeting to render all the given targets at once.

ProGamerGov:
These images show the changes in this PR, as GitHub is not showing the correct file diff:
diff_1_wv
diff_2_wv
diff_3_wv
diff_4_wv
diff_5_wv

ProGamerGov:
@NarineK I added a set of images from an online diff checker as GitHub's diff checker isn't working on this PR.

PR #821, opened 2021-12-13
Optim-wip: Add JIT support to all transforms & some image parameterizations

ProGamerGov:
* Added JIT support to all transforms.

* Added JIT tests for all applicable transforms.

* There's currently a bug that prevents `CenterCrop` from working with JIT, so I added the `@torch.jit.ignore` decorator to it's forward function to avoid the bug. This is so that it won't throw an error if someone uses `torch.jit.script` on it. https://github.com/pytorch/pytorch/issues/69938 

* JIT is also told to ignore `NChannelsToRGB` due to it's unsupported inner functions.

* The `ToRGB` didn't support JIT due to the usage of named dimensions, so I implemented a second forward function that is only used when using JIT.

* Add JIT support `NaturalImage`, `FFTImage`, & `PixelImage`.

* Improved `NaturalImage` & `FFTImage` tests, and test coverage.

* Added `ImageParameterization` instance support for `NaturalImage`. This improvement should make it easier to use parameterization enhancements like SharedImage, and will be helpful for custom parameterizations that don't use the standard input variable set (size, channels, batch, & init).

* Added asserts to verify `NaturalImage` parameterization inputs are instances or types of `ImageParameterization`.

* Some transform type hints had to be changed due to a bug in JIT's Union support: https://github.com/pytorch/pytorch/issues/69434

* Fixed bug with `RandomCrop` transform.

* Replaced default `RandomScale` transform with Interpolation based variant. Renamed the old variant to `RandomScaleAffine`.

* `CenterCrop` / `center_crop` now adds padding if the crop size is larger than the input dimensions.

* Add distributions support to both versions of `RandomScale`. Ludwig wanted this based off of his original PR.

* Changed `NumSeqOrTensorType` hint to `NumSeqOrTensorOrProbDistType`.

* Added more comprehensive testing to applicable transforms. Test coverage reported by pytest should now be 100% minus the version specific tests.

* Transform torch version checks required for module forward functions are performed in their  `__init__` function so that they work with JIT.

* Add `TransformationRobustness()` transform.
    * Similar to Lucid's `standard_transforms`, the `TransformationRobustness()` is a convenient helper for applying the standard set of transforms in the correct order. The default values have also been tuned to optimal values.
    * It performs the following transforms: A padding transform, one or more `RandomSpatialJitter` transforms, a `RandomScale` transform, a `RandomRotation` transform, a single `RandomSpatialJitter` transform, and then a final optional CenterCrop or Pad To transform.
    * There's a pretty significant improvement to visualization quality when using `TransformationRobustness()`, and it makes it easier to use transform robustness properly..
    * Links to Lucid's versions: https://github.com/tensorflow/lucid/blob/master/lucid/optvis/transform.py#L164, https://github.com/tensorflow/lucid/blob/master/lucid/recipes/activation_atlas/render.py#L127
    * Ludwig's original Captum PR had the same transform planned: https://github.com/ludwigschubert/captum/blob/f1fd0729dece59564a7c10b7b397617d8a09a247/captum/optim/param/transform.py#L88
    * It was also used for feature visualization with OpenAI's CLIP model: https://github.com/openai/CLIP-featurevis/blob/master/example_facets.py#L61

```
    pad_module = torch.nn.ConstantPad2d(2, value=0.5)
    transforms = opt.transforms.TransformationRobustness(padding_transform=pad_module)
```

JIT support for the InceptionV1 model was added to https://github.com/pytorch/captum/pull/655

This PR only changes / adds around 711 lines of code, with the other 1306 lines being from tests that feature a lot of redundant / simple code.

### DIFF
--- a/captum/optim/_param/image/images.py
+++ b/captum/optim/_param/image/images.py
@@ -620,7 +620,7 @@ class NaturalImage(ImageParameterization):
                 nn.Parameter tensor, or stacking init images.
                 Default: 1
             parameterization (ImageParameterization, optional): An image
-                parameterization class.
+                parameterization class, or instance of an image parameterization class.
                 Default: FFTImage
             squash_func (Callable[[torch.Tensor], torch.Tensor]], optional): The squash
                 function to use after color recorrelation. A funtion or lambda function.
@@ -632,8 +632,14 @@ class NaturalImage(ImageParameterization):
                 Default: True
         """
         super().__init__()
+        if not isinstance(parameterization, ImageParameterization):
+            # Verify uninitialized class is correct type
+            assert issubclass(parameterization, ImageParameterization)
+        else:
+            assert isinstance(parameterization, ImageParameterization)
+
         self.decorrelate = decorrelation_module
-        if init is not None:
+        if init is not None and not isinstance(parameterization, ImageParameterization):
             assert init.dim() == 3 or init.dim() == 4
             if decorrelate_init and self.decorrelate is not None:
                 init = (
@@ -647,9 +653,11 @@ class NaturalImage(ImageParameterization):
                 squash_func = self._clamp_image
 
         self.squash_func = torch.sigmoid if squash_func is None else squash_func
-        self.parameterization = parameterization(
-            size=size, channels=channels, batch=batch, init=init
-        )
+        if not isinstance(parameterization, ImageParameterization):
+            parameterization = parameterization(
+                size=size, channels=channels, batch=batch, init=init
+            )
+        self.parameterization = parameterization
 
     @torch.jit.export
     def _clamp_image(self, x: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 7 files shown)


ProGamerGov:
Looks like there was an error with the Insights install in the tests. Everything optim related is working fine however.

ProGamerGov:
Upon observation of the performances of both `RandomScale` variations, the Affine transform one performs better than the interpolation one.

ProGamerGov:
I may be able to get the better quality lanczos resizing mode working for `RandomScale`'s as a better alternative to the current `bilinear`default. It turns out that lanczos interpolation can be accomplished via Conv2d followed by `bicubic` interpolation. It actually fits in really neatly with `RandomScale`'s code as it's an optional operation before running `F.interpolate`.

ProGamerGov:
@NarineK Once this PR is merged, I'll create a PR adding `lanczos` interpolation to `RandomScale` with only about 24 lines of code, plus the testing code. Lanczos is higher quality than the default `bilinear` interpolation, but it's a little bit slower.

ProGamerGov:
I've been using the `TransformationRobustness` in all of the newer notebooks due to it's easy of use and dramatic improvement in visualization quality.

The optimizing with transparency tutorial notebook requires this PR for the `TransformationRobustness` transform and the fixes made to `CenterCrop`.

PR #822, opened 2021-12-13
Optim-wip: Fix the Channel Attr notebook image download & metadata filename

ProGamerGov:
GitHub's diff checker is reporting that the entire file changed, but I only changed a few lines of code.

* The image that this tutorial was using is no longer accessible as the hosting container for it was removed. I uploaded the image to GitHub and now the tutorial downloads it from GitHub instead.

* Previously the filename in the metadata was listed as "CaptumNotebookCells.ipynb" instead of "ChannelAttribution_OptimViz.ipynb".


NarineK:
@ProGamerGov, do you mind pointing out here which lines did you change ?

ProGamerGov:
@NarineK I changed the rendering to use batch targeting so that we didn't have to render each item in the list separately. I also changed the image download link, and updated the transform path names: `optimviz` -> `opt`, `transform` -> `transforms`.

NarineK:
@ProGamerGov, do you mind looking into mypy checks and  fixing them ?

ProGamerGov:
@NarineK The MyPy error associated with the optim model is fixed in the optim-wip branch already

NarineK:
@ProGamerGov , do you mind then rebasing from the optim-wip branch so that the errors get fixed ?

ProGamerGov:
@NarineK Okay, I've done that and the error is no longer present, There will hopefully be no errors once the fixes from https://github.com/pytorch/captum/pull/935 are in play

NarineK:
Thank you, @ProGamerGov!  I left comments in https://github.com/pytorch/captum/pull/935

PR #823, opened 2021-12-13
make av.generate_dataset_activations return AVDatasets

99warriors:
Summary:
AV.generate_dataset_activations now returns a list of AVDatasets (or a single AVDataset, depending on whether layers is a str or list of str).  These AVDatasets are explicitly used in similarity_influence.py.

Also modified existing tests.

optional: This change does not affect the total complexity, since we are just moving calls to `AV.load`, not creating new calls.  We are making calls to `AV.load` all at once, which is fine for memory.  This may lead to a longer wait (without other computation in between that would print out some indicator of progress).  To let the user know `AV.load` is being called, `AV.load` now prints out that the method is being started and stopped and how many files were found by `glob`.

Differential Revision: D32749369


NarineK:
@99warriors, it looks like @i-jones fixed typehint issues. The errors will bee fixed after that PR gets merged.

https://github.com/pytorch/captum/pull/843/files#diff-82c35b4e66c58e6ded430a9f35324015f6683a3912c5562cc0442e9bac340d0dR87

PR #828, opened 2021-12-20
Optim-wip: Composable loss improvements

ProGamerGov:
This PR adds a few simple improvements to the `CompositeLoss` class and it's features.

* Added support for `__pos__` and `__abs__` unary operators to `CompositeLoss`. These appear to be the only other basic operators that make sense to add support for.
* Made `CompositeLoss`'s reduction operation a global variable that can be changed by users. This should improve the generality of the optim module, and it makes it possible to disable this aspect of `CompositeLoss`.
* Added composable `torch.mean` and `torch.sum` reduction operations to `CompositeLoss`. These are common operations, so there are likely use cases that can benefit from them. Example usage: `loss_fn.mean()` & `loss_fn.sum()`.
* Added tests for the changes listed above.

I'm also working on ways for users to create their own custom composite loss operations, similar to the `sum_loss_list` function, for future PRs.

### DIFF
--- a/captum/optim/_core/loss.py
+++ b/captum/optim/_core/loss.py
@@ -16,6 +16,10 @@ def _make_arg_str(arg: Any) -> str:
     return arg[:15] + "..." if too_big else arg
 
 
+# Reduction op for CompositeLoss loss composability size mismatch avoidance
+REDUCTION_OP: Callable[[torch.Tensor], torch.Tensor] = torch.mean
+
+
 class Loss(ABC):
     """
     Abstract Class to describe loss.
@@ -39,6 +43,12 @@ class Loss(ABC):
 
     def __neg__(self) -> "CompositeLoss":
         return module_op(self, None, operator.neg)
+
+    def __pos__(self) -> "CompositeLoss":
+        return module_op(self, None, operator.pos)
+
+    def __abs__(self) -> "CompositeLoss":
+        return module_op(self, None, operator.abs)
 
     def __add__(self, other: Union[int, float, "Loss"]) -> "CompositeLoss":
         return module_op(self, other, operator.add)
@@ -68,7 +78,7 @@ class Loss(ABC):
         if isinstance(other, (int, float)):
 
             def loss_fn(module: ModuleOutputMapping) -> torch.Tensor:
-                return operator.truediv(other, torch.mean(self(module)))
+                return operator.truediv(other, REDUCTION_OP(self(module)))
 
             name = self.__name__
             target = self.target
@@ -86,7 +96,7 @@ class Loss(ABC):
         if isinstance(other, (int, float)):
 
             def loss_fn(module: ModuleOutputMapping) -> torch.Tensor:
-                return operator.pow(other, torch.mean(self(module)))
+                return operator.pow(other, REDUCTION_OP(self(module)))
 
             name = self.__name__
             target = self.target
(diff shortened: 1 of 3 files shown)


ProGamerGov:
While investigating JIT support for InputOptimization (or at least the relevant parts), I discovered that it's creating ghost hooks that never go away.

The hooks are registered in the initialization function, and if the `optimize` class function is not run then the hooks are never removed as can be seen by using the follow function:

```
def find_and_remove_hooks(m):
    print("Hooks")
    c = 0
    module_name = type(m).__name__
    print(m._forward_hooks)
    for k, v in m._forward_hooks.items():
        c+=1
    print("All " + str(c) + " hooks found")
```

aobo-y:
Hi, thank you for making this, but I may miss some context/history here. Why do we need the "composable loss" in Captum? 

Pytorch has already provided a convention for loss: `function`/`callable-module` wrapping some `tensor` operations. For example, if I need a loss made of others
```py
def new_loss(output_tensor):
  return nn.SomeLoss(output_tensor) + some_other_loss(output_tensor) + torch.linalg.norm(output_tensor)
```

Pytorch `tensor` has supported these basic arithmetic operations to modify & combine loss tensors. Why are we composing `function/callable-module` with arithmetic operations, instead of composing tensors? Pytorch supports more operation than what we have. And the "composable loss" cannot be composed with any existing Pytorch losses.

I think our optim loss can work the same without "composable loss" and even be more flexible. For example
```py
deepdream = DeepDream(target)
layeractivation = LayerActivation(target)
def new_loss(targets_to_values: ModuleOutputMapping):
  loss = deepdream(targets_to_values) + layeractivation(targets_to_values)
  # can also use pytorch loss
  return loss + nn.SomeLoss(targets_to_values[target])
```

captum/optim/_core/loss.py
```diff
@@ -138,18 +170,110 @@ def loss_fn(module: ModuleOutputMapping) -> torch.Tensor:
...
+
+        name = loss_op_fn.__name__ + "(" + loss.__name__ + ")"
+        target = loss.target
+    return CompositeLoss(loss_fn, name=name, target=target)
+
+
 class BaseLoss(Loss):
     def __init__(
         self,
         target: Union[nn.Module, List[nn.Module]] = [],
```
aobo-y, line 265: can the `target` be `List[nn.Module]`? Many losses below directly access `targets_to_values[self.target]`. Did I miss anything?

ProGamerGov: @aobo-y Losses like `ActivationInterpolation` have multiple targets (Faceted loss as well in an upcoming PR), but `BaseLoss` works off using a single `target` variable. 
 
The `BaseLoss` class is called in the `__init__` functions of loss classes like so:

```
# Single target
BaseLoss.__init__(self, target, batch_index)

# Multiple targets
BaseLoss.__init__(self, [target1, target2])
```

The loss class itself will indicate via `target: List[nn.Module]` type hint that multiple targets are supported / required, or it is handled things internally by passing the targets as a list to `BaseLoss` like in `ActivationInterpolation`.

The `ActivationInterpolation` loss class can be found here: https://github.com/ProGamerGov/captum/blob/optim-wip-composable-loss-improvements/captum/optim/_core/loss.py#L506

aobo-y: Sure, but cases like `DeepDream` and some others directly inherits BaseLoss's init definition, where `target` can be a list while actually it should not https://github.com/ProGamerGov/captum/blob/optim-wip-composable-loss-improvements/captum/optim/_core/loss.py#L393-L407 

If these losses have different assumptions of what their targets should be, why do we abstract the `target` into the base class. The base class `BaseLoss` does not need `target` anyway. Each class can define their own `target` in `__init__`. Or we can have 2 other intermediate abstract classes `SingleTargetLoss` `MultiTargetsLoss`

But anyway, this is just for discussion. It has nth related to this PR. We can leave it to future updates if needed.

ProGamerGov: Oh, yeah I see what you mean now. In the original code, I think that Ludwig had `SingleTargetObjective` & `MultiObjective` for handling these cases: https://github.com/ludwigschubert/captum/blob/f1fd0729dece59564a7c10b7b397617d8a09a247/captum/optim/optim/objectives.py#L108

It'd probably be best to leave this to a future PR if decide on the changes

ProGamerGov:
@aobo-y Originally Captum’s loss functions [were setup similar]( https://github.com/pytorch/captum/pull/500#discussion_r515453337) to the simple class-like functions like [Lucid uses](https://github.com/tensorflow/lucid/blob/master/lucid/optvis/objectives.py). Upon review we then changed the losses to use classes [instead](https://github.com/pytorch/captum/pull/527).

Ludwig (one of the main Lucid developers) designed the initial optim module to utilize a Lucid-like composable loss system. One of the main benefits of the composable loss system is ease of use and built-in target tracking (the list of targets has to created regardless of whether not we use composable losses). It also allows for easy-to-use handling of things like batch specific targeting.

ProGamerGov:
This PR can be skipped for now.

PR #832, opened 2021-12-31
Optim-wip: Add new StackImage parameterization & JIT support for SharedImage

ProGamerGov:
* Added `SimpleTensorParameterization` as a workaround for JIT not supporting `nn.ParameterList`. It also helps `StackImage` support tensor inputs.
* Added JIT support for `SharedImage`.
* Added new parameterization called `StackImage`, that stacks multiple parameterizations (that are can be on different devices) along the batch dimension.

Maybe we should try to find a name for the parameterizations that augment the existing parameterizations?


ProGamerGov:
Oops, this PR still has the commits from the other branch.

PR #837, opened 2022-01-07
Migrate to register_full_backward_hook

vivekmig:
Modifying all module backward hooks to utilize the new register_full_backward hook API documented here. This new API resolves many issues we previously encountered with backward module hooks.

Since this API is available only in torch 1.8, allowing a fall-back option to the original backward hook approach.

Due to issues described [here](https://github.com/pytorch/pytorch/issues/57157), we are also deprecating attribution with respect to neuron outputs for NeuronDeepLift, NeuronGuidedBackprop, and NeuronDeconvolution; these methods require attributing with respect to neuron input (which is typically equivalent to attributing with respect to the previous layer output).

Additionally, in-place modules are no longer supported for full backward hooks, so these are no longer supported for DeepLift, LRP, and GuidedBackprop / Deconvolution and corresponding variants. Documentation has been updated accordingly.


tests/attr/test_lrp.py
```diff
@@ -124,19 +124,6 @@ def test_lrp_simple_repeat_attributions(self) -> None:
         output_after = model(inputs)
         assertTensorAlmostEqual(self, output, output_after)
 
-    def test_lrp_simple_inplaceReLU(self) -> None:
```
NarineK, line 127: @vivekmig, there are several pre-trained models that have `inplace=True` for ReLU. I think @nanohanno, specifically needed it for his models, if I remember correctly. Does this mean that we need to make that change in the pre-trained models ?

captum/attr/_core/deep_lift.py
```diff
@@ -107,7 +108,9 @@ def __init__(self, model: Module, multiply_by_inputs: bool = True) -> None:
         r"""
         Args:
 
-            model (nn.Module):  The reference to PyTorch model instance.
+            model (nn.Module):  The reference to PyTorch model instance. Model cannot
+                        contain any in-place nonlinear submodules; these are not
+                        supported by the register_full_backward_hook PyTorch API.
```
NarineK, line 113: maybe add: `API starting from v1.8.0 `?

captum/_utils/gradient.py
```diff
@@ -84,6 +84,27 @@ def undo_gradient_requirements(
...
 
+def register_backward_hook(
+    module: Module, hook: Callable, attr_obj: Any
+) -> torch.utils.hooks.RemovableHandle:
+    # Special case for supporting output attributions for neuron methods
+    # This can be removed after deprecation of neuron output attributions
+    # for NeuronDeepLift, NeuronDeconvolution, and NeuronGuidedBackprop
+    # in v0.6.0
+    if (
+        hasattr(attr_obj, "skip_new_hook_layer")
```
NarineK, line 95: nit: is this `skip_new_hook_layer` or `use_old_backward_hook`  ?

tests/attr/test_guided_backprop.py
```diff
@@ -120,7 +121,8 @@ def _guided_backprop_matching_assert(
         test_input: TensorOrTupleOfTensorsGeneric,
     ):
         out = model(test_input)
-        attrib = GuidedBackprop(model)
+        model_copy = copy.deepcopy(model)
```
NarineK, line 124: nit: why do we do `deepcopy` here ?

vivekmig: deepcopy is used here since once full_backward_hooks are placed on a module, then standard backward hooks can no longer be added. Since this is using neuron guided backprop (with layer output), this falls back to standard backward hooks within NeuronGuidedBackprop but only full backward hooks in GuidedBackprop, requiring a copy of the model rather than using the same model.

NarineK (approved):
LGTM! Approving, made minor doc suggestions.

tests/attr/test_lrp.py
vivekmig: Yes, I think we would need to change this for pre-trained models as well. I didn't see any explicit documentation of this limitation, but obtained an error with toy models using in-place modules, as below. Maybe @albanD can provide more context or documentation?

```
import torch
import torch.nn as nn

class SampleRelu(nn.Module):
    def __init__(self):
        super().__init__()
        self.relu = nn.ReLU(inplace=True)

    def forward(self, inp):
        return self.relu(inp)
relu_mod = SampleRelu()

def grad_hook(mod, inp, out):
  print("HOOK")
hook = relu_mod.relu.register_full_backward_hook(grad_hook)
inp = torch.randn(3,4, requires_grad=True)
out = relu_mod(inp)
```
Result:
```
RuntimeError: Output 0 of BackwardHookFunctionBackward is a view and is being modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can fix this by cloning the output of the custom Function.
```

albanD: Hey!
The reason is that for inplace ops, the input and output to your nn.Module is effectively the same Tensor. So it is not clear which gradient you expect to see there: the same or two different things?

The other thing is that for Module with multiple inputs or outputs, we can't technically do it. We can special case the one input/output case though if you think the input's gradient should be for the value before it was modified and the output's for the value after it was modified?

andresfp14: Ok, I had this exact same issue half an hour ago.... as a dirty fix to my model, I just changed all my inplace Relu to not in place ones:
```python
def correct_inplace_relu(model):
    for child_name, child in model.named_children():
        if isinstance(child, nn.ReLU):
            setattr(model, child_name, nn.ReLU())
        else:
            correct_inplace_relu(child)
correct_inplace_relu(model)
```
Afterwards, everything works fine.

expectopatronum:
Does this warning mean the output for GuidedBackPropagation will not be correct? (I am currently using the following pretrained torchvision models: Alexnet, VGG16, Resnet50, DenseNet121)

PR #839, opened 2022-01-10
Optim-wip: Add better colorspace support, image grids, & user agent to ImageTensor

ProGamerGov:
* Added color space support to `save_tensor_as_image` & `ImageTensor.export`.
* Added image grid creation support to `ImageTensor.export` , `ImageTensor.show` , `show` & `save_tensor_as_image` via a new `make_grid_image` function. This makes it easier for users to view multiple images stacked across the batch dimension without having to scroll horizontally.
* Added user agent to `ImageTensor.open` as sites like Wikipedia require user agents.


tests/optim/utils/image/common.py
```diff
@@ -8,7 +8,182 @@
...
+                        [0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0],
+                        [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                        [0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0],
+                        [0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 0.0],
+                        [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                    ]
+                ]
+            ]
+        )
+        self.assertEqual(list(expected_output.shape), [1, 1, 7, 10])
```
NarineK, line 32: The transformation doesn't seem very clear. Do you mind adding more details how 6 x 1 x 2 x 2 transforms into 1 x 1 x 7 x 10 ? `nrow` argument looks a bit unclear to me.

NarineK:
@ProGamerGov, where is this "make image grid" functionality used now ? Are we using it in activation atlas ?

ProGamerGov:
@NarineK The functionality is currently used for when visualizations are displayed to the user, and when they wish to save them as images. Previously users would have to scroll a lot horizontally when viewing multiple visualizations stacked across the batch dimension. The activation atlas grids are different as not every cell has a visualization, and the atlas grid is defined based on the dimensionality reduced point graph. 

In this PR for displaying images we simply specify how many images per row and if we want any padding.

tests/optim/utils/image/common.py
ProGamerGov: The is 7 x 10 because have 3, 2 by 2 tensors in each row, which gives us two rows as we have 6 tensors in total. For this test we have 3 * 2 rows, which gives us 6 tensors with a border of size 1. For the vertical portion the border has a total size of 3, while there are 4 border values horizontally. The border doesn't double in size when it separates two tensors, and thus the values are not the same as if we simply padded the inputs with a border and then concatenated them.

ProGamerGov:
```
(3* 1) + 1+ (3* 2) = 10
(2* 1) + 1+ (2* 2) = 7

# Border has a size of 1, and we are using 3 in a row:
total_w = (n_row * border) + border + (n_row * tensor_dim_w)
total_h = (n_column * border) + border + (n_column * tensor_dim_h)
```

ProGamerGov: I've added the above comment to the tests!

ProGamerGov: I've added the above comment to the tests!

ProGamerGov:
@NarineK Currently this is what happens for example if a user uses the show function to display 12 visualizations stacked across the batch dimension: 
no_grid_example

With this PR, the same user could specify `nrow=4` and create a much better looking display for visualizations:
with_grid_example

NarineK:
Thank you for the explanation, @ProGamerGov ! In that case it's would be better to call it `images_per_row`. I originally thought that we are expected to see three rows.

ProGamerGov:
@NarineK Ah, I see the problem! I've changed `nrow` to `images_per_row` and corrected the documentation for the `images_per_row` argument.

PR #844, opened 2022-01-21
Update lrp.py

ericotjo001:
Tested on pretrained ResNet50. Without the additional "if layer.rule is not None:", the following error will be raised:

Traceback (most recent call last):
  File "/home/ericotjoaubu/.local/lib/python3.9/site-packages/captum/attr/_core/lrp.py", line 272, in _check_and_attach_rules
    layer.rule.relevance_input = defaultdict(list)  # type: ignore
AttributeError: 'NoneType' object has no attribute 'relevance_input'


ericotjo001:
Will retry pulling request after signing cla.

PR #850, opened 2022-01-27
Optim-wip: Add Class Activation Atlas Tutorial Notebook

ProGamerGov:
This PR simplifies the commit history of: https://github.com/pytorch/captum/pull/730

This PR adds the `ClassActivationAtlas_OptimViz.ipynb` tutorial notebook, which is the final activation atlas tutorial. I've also updated the class atlas notebook based on the feedback for the main atlas tutorial.


NarineK:
@ProGamerGov, sorry for getting back late to this PR. I've just had a quick look and it looks like some of the things that we showcase here we already have in the tutorial that we've already merged:  
https://github.com/pytorch/captum/blob/optim-wip/tutorials/optimviz/atlas/ActivationAtlas_OptimViz.ipynb

For example `Activation Atlas Shape` and `Dimensionality Reduction For Coordinate Grid Creation`. Do we want to focus on multi-class only in this tutorial ?

ProGamerGov:
@NarineK I think that we can probably just remove the 'Activation Atlas Shape' section as it's covered in the main atlas tutorial, and it's not particularly important for the class atlas tutorial. I'll also change the first 'Dimensionality Reduction For Coordinate Grid Creation' section's text to be more simplified like in part 2 of the tutorial:

> After filtering our activation samples, we can now create the atlas xy coordinate graph. To do this, we'll use UMAP to reduce the number of channels down to 2. This will give us the x and y axes of our graph.

NarineK:
> [ProGamerGov](/ProGamerGov)

@ProGamerGov, if it is text description, similarity it is fine. If those are functions that are shared between 2 tutorials you can put them in a python file  under:

`tutorials/optimviz/atlas/ ... `

NarineK:
> > [ProGamerGov](/ProGamerGov)
> 
> @ProGamerGov, if it is text description, similarity it is fine. If those are functions that are shared between 2 tutorials you can put them in a python file under:
> 
> `tutorials/optimviz/atlas/ ... `

To avoid duplicates you can also refer to the other notebook from this notebook. So that the users can check out the other one first.

ProGamerGov:
@NarineK So actually other than the two items that you mentioned, I don't think there's any other duplication of the other atlas tutorial.

NarineK (approved):
@ProGamerGov, thank you for the PR and sorry for the  delay reviewing it. 

Here are some minor comments:

1. nit: multi-class ? (seen at several places)
2. The class-specific version that only focuses on the specified class. -> Maybe this bullet point can come first ?
3. So, there are two versions: `class-specific`  and `class-specific with context` ? What is the context in that case ? Maybe you can describe it a little bit ?
4. In the section `Dimensionality Reduction For Coordinate Grid Creation` do you mind adding a description on how to interpret sample activation atlas ?
5. we will used Captum.optim's  -> we will use  ... ?
6. Wouldn't we get duplicate indices here if some indices in `class_indices` overlap with `top_positive_class_indices` ?
7. Sample Whitening -> Do you mind moving this closer to the section where it is used ?
8. In the bottom of `Rendering The Single Class Activation Atlas Visualizations` section maybe you can write a little bit interpretation of the visualized class activation atlas. E.g. `On the atlas we recognize high level concepts associated to scuba diver such as ... ` ?
9. For the `xy coordinate grid` do you mind labelling the axes ?

ProGamerGov:
@NarineK 

1. Fixed!
2. Done!
3. The bullet-point section already explains the differences, I think.
4. The text explains this already I think? The channels are reduced to 2, and are treated as the x and y axes.
5. Fixed!
6. Yeah, it seems like there are a few duplicate indices. The research paper's [supplementary code](https://github.com/tensorflow/lucid/blob/master/notebooks/activation-atlas/class-activation-atlas.ipynb) also seemingly allows for duplicates, so it could be intentional.
7. The sampling whitening is used in both sections of the tutorial, so I placed it at the top in order to avoid repeating it for both sections. It's also possible for the user to skip the first atlas section as well.
8. Done!
9. Not 100% sure what you mean here?

NarineK:
> 6\. supplementary code

Thank you, @ProGamerGov!
4. Does it say that x and y axes are similarity scores ? Maybe such kind of description if it doesn't
9. I meant to add label to x and y axes in the plot. For example, for y-axis  on the plot you can write `Similarity Score` and for the X-axis `Class Differences`. You describe it in the text but it would bee good to have it as a description of axes as well

ProGamerGov:
@NarineK Okay, I've added the labels to the second plotted graph! For the first plotted graph, we can probably leave it as is, as it's created the same way as the graph in the other atlas tutorial.

NarineK:
Thank you, @ProGamerGov! Merging ...

Issue #835, opened 2022-01-03
NaN-value attributions when using DeepLIFT during fp16 training


aarzchan:
Hi! I'm trying to compute attributions for BigBird via DeepLIFT during fp16 training, but I'm getting NaN for many of the attribution values. However, if I do fp32 training, none of the attribution values are NaN. I guess the fp16 precision is truncating some near-zero intermediate values to zero, resulting in division-by-zero. Would appreciate any ideas on how to work around this issue!

For reference, the `inputs` and `baselines` parameters for my DeepLIFT attribute function call are constructed as follows:

`inputs`
- Feed input ids into BigBird word embedding layer to get input embeddings.
- Use input embeddings as inputs for DeepLIFT.

`baselines`
- Create baseline ids by setting all non-special tokens in the input ids to PAD.
- Feed baseline ids into BigBird word embedding layer to get baseline embeddings.
- Use baseline embeddings as baselines for DeepLIFT.

bilalsal:
Hi @aarzchan , thank you for reporting this issue!

Does it help if you use an fp16 tensor for the `baselines` parameter in `DeepLift`?

aarzchan:
Hi @bilalsal! Thanks for your comment, and apologies for my delayed response. I was busy with some other tasks recently.

I've tried casting different combinations of `inputs` and `baselines` as `Float16` type, but I get the same output.

If I cast `inputs` and/or `baselines` as `Float32` type, then I get the error `RuntimeError: expected scalar type Double but found Float`, since the model parameters are `Float16` type.

bilalsal:
I see. Thank you @aarzchan . Is it easy to create a minimal Colab notebook demonstrating the error? Would love to play with this further.

aarzchan:
Hi @bilalsal, here is a [Colab notebook](https://colab.research.google.com/drive/1UN4j42EMV--6bPPgegEefhOSNWrKwzX6?usp=sharing) demonstrating the error. I shared it with your FB email.

NarineK:
@aarzchan, @bilalsal  has this issue been addressed after we configured the eps in https://github.com/pytorch/captum/pull/854 ?

bilalsal:
Hi @NarineK. I shared a [Colab notebook](https://colab.research.google.com/drive/12HThc6JNQTYTNZrvGm2c1pSu0m9zZoN0) with Aaron demonstrating the fix.
I think we can close this issue, unless @aarzchan has further remarks / questions.

[...]

PR #854, opened 2022-02-03
Make the 'eps' parameter user-adjustable in the .attribute() method.

bilalsal:
Important to fix numerical stability issues as in #835

### DIFF
--- a/captum/attr/_core/deep_lift.py
+++ b/captum/attr/_core/deep_lift.py
@@ -103,11 +103,12 @@ class DeepLift(GradientAttribution):
     https://pytorch.org/blog/optimizing-cuda-rnn-with-torchscript/
     """
 
-    def __init__(self,
-                 model: Module,
-                 multiply_by_inputs: bool = True,
-                 eps: float = 1e-10,
-                 ) -> None:
+    def __init__(
+        self,
+        model: Module,
+        multiply_by_inputs: bool = True,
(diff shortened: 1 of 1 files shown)


captum/attr/_core/deep_lift.py
```diff
@@ -322,7 +328,7 @@ def attribute(  # type: ignore
                activations. The hooks and attributes will be removed
             after the attribution is finished"""
         )
-
+        self.eps = eps
```
NarineK, line 331: Thank you for the pr, @bilalsal ! We usually don't set any state variables in the attribute computation method. We can either pass this in the constructor or pass to the method as an argument instead of making it an instance variable.

bilalsal: Thanks for the hint Narine! Yes, the constructor is the safer place (this also reduces the complexity of using .attribute(), it is unlikely users want to use different eps for the same model).

NarineK: Thank you, @bilalsal ! Looks good. There is a small formatting issue. That you can probably also reproduce with `ufmt format .` Let's address that before merging.

NarineK: cc: @bilalsal

NarineK (approved):
Thank you, @bilalsal ! Let's merge this.

PR #858, opened 2022-02-10
Optim-wip: Fix bug with nn.Sequential targets

ProGamerGov:
Some PyTorch modules that can be used as targets have an `__iter__` attribute, like `torch.nn.Sequential`. The current target system only checks for the `__iter__` attribute to determine if a target is already in a list or not, and thus doesn't work with targets like Sequential modules. This PR fixes this bug.

This PR also corrects `InputOptimization`'s model variable type hint to reflect the fact that it's optional.

### DIFF
--- a/captum/optim/_core/loss.py
+++ b/captum/optim/_core/loss.py
@@ -127,16 +127,8 @@ def module_op(
             return math_op(torch.mean(self(module)), torch.mean(other(module)))
 
         name = f"Compose({', '.join([self.__name__, other.__name__])})"
-        target = (
-            self.target
-            if hasattr(self.target, "__iter__")
-            and not isinstance(self.target, nn.Module)
-            else [self.target]
-        ) + (
-            other.target
-            if hasattr(other.target, "__iter__")
-            and not isinstance(other.target, nn.Module)
-            else [other.target]
+        target = (self.target if isinstance(self.target, list) else [self.target]) + (
+            other.target if isinstance(other.target, list) else [other.target]
         )
     else:
         raise TypeError(
@@ -724,12 +716,7 @@ def sum_loss_list(
     target = [
         target
         for targets in [
-            [loss.target]
-            if not (
-                hasattr(loss.target, "__iter__")
-                and not isinstance(loss.target, nn.Module)
-            )
-            else loss.target
+            [loss.target] if not isinstance(loss.target, list) else loss.target
             for loss in loss_list
         ]
         for target in targets
--- a/captum/optim/_core/loss.py
+++ b/captum/optim/_core/loss.py
@@ -724,7 +724,12 @@ def sum_loss_list(
     target = [
         target
         for targets in [
-            [loss.target] if not hasattr(loss.target, "__iter__") else loss.target
+            [loss.target]
+            if not (
+                hasattr(loss.target, "__iter__")
+                and not isinstance(loss.target, nn.Module)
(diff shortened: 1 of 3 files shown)


captum/optim/_core/loss.py
```diff
@@ -128,8 +128,16 @@ def loss_fn(module: ModuleOutputMapping) -> torch.Tensor:
...
-            self.target if hasattr(self.target, "__iter__") else [self.target]
-        ) + (other.target if hasattr(other.target, "__iter__") else [other.target])
+            self.target
+            if hasattr(self.target, "__iter__")
+            and not isinstance(self.target, nn.Module)
+            else [self.target]
+        ) + (
+            other.target
+            if hasattr(other.target, "__iter__")
+            and not isinstance(other.target, nn.Module)
```
aobo-y, line 138: This of course will work! But we have already clearly defined the type of `self.target` as l`Union[nn.Module, List[nn.Module]]`. So we only need to check if it is a `list` and no need to worry about other `iterable`. We can just do
```py
if isinstance(self.target, list)
```

ProGamerGov: Yeah, I think that the `hasattr(obj, __iter__)` is a holdover from when SK was putting together the first iteration of the comprehension system, when weren't 100% sure how it was going to work.

I've replaced the relevant lines with `isinstance(self.target, list)`!

captum/optim/_core/optimization.py
```diff
@@ -56,15 +56,17 @@ def __init__(
                         optimization.
             lr (float): The learning rate to use with the Adam optimizer.
         """
-        self.model = model
+        self.model = model or nn.Identity()
```
aobo-y, line 59: add `optional` to the type in the corresponding doc (comments above) as well? btw, what is the use case of not giving a `model`

captum/optim/_core/optimization.py
```diff
@@ -56,15 +56,17 @@ def __init__(
...
-        if hasattr(loss_function.target, "__iter__"):
+        assert hasattr(loss_function, "target")
+        if isinstance(loss_function.target, list):
             self.hooks = ModuleOutputsHook(loss_function.target)
         else:
             self.hooks = ModuleOutputsHook([loss_function.target])
         self.input_param = input_param or NaturalImage((224, 224))
         if isinstance(self.model, Iterable):
-            self.input_param.to(next(self.model.parameters()).device)
+            if list(self.model.parameters()) != []:
```
aobo-y, line 68: covert the generator to a list means we have to iterate the entire generator and store them, which is unnecessarily expensive. If you just want to check if the model has parameters, try call `next` once with fallback
```py
param = next(self.model.parameters(), None)
if param:
  self.input_param.to(param.device)
```

ProGamerGov: @aobo-y Thanks for the tip! I've changed the code to your suggested less intensive way of checking for parameters to use as a source for what device to use.

captum/optim/_core/optimization.py
ProGamerGov: @aobo-y I thought that listing optional in the docs meant that there was a preset value for the variable? I've added it now though.

The reasoning behind making the model optional is that for example in some cases the image parameterization may be a generative model. Thus it doesn't make a lot of sense to require a second model be used in these cases unless the user wants to link them.

PR #860, opened 2022-02-11
Add utilities for per-sample param gradients

vivekmig:
This PR adds utilities for computing per-sample param gradients, refactoring and improving upon the initial version of autograd hacks utilized.

Particularly this implementation now supports:
- Most conv layers with custom settings of padding, stride, dilation etc.
- Multi-dimensional inputs to linear models (> 2d)
- GPU inputs / outputs
- Repeated usage of a single nn.module

### DIFF
--- a/captum/_utils/sample_gradient.py
+++ b/captum/_utils/sample_gradient.py
@@ -4,6 +4,7 @@ from typing import Tuple, Union, cast, Iterable
 
 import torch
 from captum._utils.common import _format_tensor_into_tuples
+from captum._utils.gradient import register_backward_hook
 from torch import Tensor
 from torch.nn import Module
 
@@ -25,6 +26,8 @@ def linear_param_grads(
     (weight and bias). If reset = True, any current sample_grad values are reset,
     otherwise computed gradients are accumulated and added to the existing
     stored gradients.
+
+    Inputs with more than 2 dimensions are only supported with torch 1.8 or later
     """
     if reset:
         _reset_sample_grads(module)
@@ -116,7 +119,7 @@ class SampleGradientWrapper:
                 module.register_forward_hook(self._forward_hook_fn)
             )
             self.backward_hooks.append(
-                module.register_full_backward_hook(self._backward_hook_fn)
+                register_backward_hook(module, self._backward_hook_fn, None)
             )
 
     def _forward_hook_fn(
--- a/captum/_utils/sample_gradient.py
+++ b/captum/_utils/sample_gradient.py
@@ -1,0 +1,144 @@
+from collections import defaultdict
+from enum import Enum
+from typing import Tuple, Union
+
+import torch
+from torch import Tensor
+from torch.nn import Module
+
+from captum._utils.common import _format_tensor_into_tuples
+
+
+def linear_param_grads(
+    module: Module, activation: Tensor, gradient_out: Tensor, reset: bool = False
+):
+    if reset:
+        module.weight.sample_grad = 0
+        if module.bias is not None:
+            module.bias.sample_grad = 0
+    print(gradient_out.shape)
+    print(activation.shape)
+
+    module.weight.sample_grad += torch.einsum(
+        "n...i,n...j->nij", gradient_out, activation
+    )
+    print(module.weight.sample_grad.shape)
+    if module.bias is not None:
+        module.bias.sample_grad += gradient_out
+
+
+def conv2d_param_grads(
+    module: Module, activation: Tensor, gradient_out: Tensor, reset: bool = False
+):
+    if reset:
+        module.weight.sample_grad = 0
+        if module.bias is not None:
+            module.bias.sample_grad = 0
+
+    batch_size = activation.shape[0]
+    unfolded_act = torch.nn.functional.unfold(
+        activation,
+        module.kernel_size,
+        dilation=module.dilation,
+        padding=module.padding,
+        stride=module.stride,
+    )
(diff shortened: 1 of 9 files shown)


captum/_utils/sample_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+        module.weight.sample_grad = 0  # type: ignore
+        if module.bias is not None:
+            module.bias.sample_grad = 0  # type: ignore
+    module.weight.sample_grad += torch.einsum(  # type: ignore
+        "n...i,n...j->nij", gradient_out, activation
+    )
+    if module.bias is not None:
+        module.bias.sample_grad += torch.einsum(  # type: ignore
+            "n...i->ni", gradient_out
+        )
```
NarineK, line 33: Does this assume that `gradient_out` and `activation` are 2 dimensional?  For a linear model they don't have to be 2 dimensional.

NarineK (commented):
Thank you for working on this PR, @vivekmig! Looks great. I left couple minor comments. Let's also add Akhil's test cases as discussed. Do GPU test cases work because I don't see that we keep track of the device ids ?

captum/_utils/sample_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+from typing import Tuple, Union, cast, Iterable
+
+import torch
+from captum._utils.common import _format_tensor_into_tuples
+from torch import Tensor
+from torch.nn import Module
+
+
+def linear_param_grads(
+    module: Module, activation: Tensor, gradient_out: Tensor, reset: bool = False
```
NarineK, line 12: could this also be tuples of tensors instead of tensors ?

captum/_utils/sample_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+        self.activation_dict[module].append(inp_tuple[0].clone().detach())
+
+    def _backward_hook_fn(
+        self,
+        module: Module,
+        grad_input: Union[Tensor, Tuple[Tensor, ...]],
+        grad_output: Union[Tensor, Tuple[Tensor, ...]],
+    ):
+        grad_output_tuple = _format_tensor_into_tuples(grad_output)
+        self.gradient_dict[module].append(grad_output_tuple[0].clone().detach())
```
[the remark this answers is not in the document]
vivekmig: We support a single GPU, e.g. model place on device cuda:0. nn.DataParallel isn't supported; that would require keeping track of a devices, etc. The previous approach only supported CPU and didn't also support a single GPU.

captum/_utils/sample_gradient.py
```diff
@@ -0,0 +1,176 @@
...
+            sample_grad_fn = SUPPORTED_MODULES[type(module)]
+            activations = self.activation_dict[module]
+            gradients = self.gradient_dict[module]
+            assert len(activations) == len(gradients), (
+                "Number of saved activations do not match number of saved gradients."
+                " This may occur if multiple forward passes are run without calling"
+                " reset or computing param gradients."
+            )
+            for i, (act, grad) in enumerate(
+                zip(activations, list(reversed(gradients)))
```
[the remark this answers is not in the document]
vivekmig: Reversing grads since when a module is used multiple times, the activations will be aligned with the reverse order of the gradients, since the order is reversed in backprop. Sure, will add a description.

captum/_utils/sample_gradient.py
vivekmig: This should work with any number of dimensions; the new test model also explicitly tests a 3D input to nn.Linear. The previous version only supported 2D inputs.

captum/_utils/sample_gradient.py
vivekmig: For now, all supported modules (nn.Conv and nn.Linear) only operate on single tensors, so for now this should only be a tensor (we extract a single tensor even if a tuple is obtained from the hook). This line also only affects nn.Linear. If we need to add support for new modules which operate on multiple inputs, we should definitely look into this further.

captum/_utils/sample_gradient.py
NarineK: @vivekmig, I asked because I tried n > 2 and it failed for linear model. I probably set something incorrectly.
```
import torch

a = torch.randn(4, 5)
b = torch.randn(4, 4)

# Passes
torch.einsum(  # type: ignore
        "n...i,n...j->nij", a, b
    ).shape

a = torch.randn(4, 5, 4)
b = torch.randn(4, 4, 4)

# Fails
torch.einsum(  # type: ignore
        "n...i,n...j->nij", a, b
    ).shape
```

NarineK (approved):
LGTM! Thank  you!

captum/_utils/sample_gradient.py
vivekmig: I think it works as long as all dimensions between the first and last are the same for both tensors. The '...' should serve to sum any dimensions in between the first and last. Your example seems to work fine when I run in colab.

In the previous version, it seems the tensors were these shapes:
```
a = torch.randn(4, 5, 4)
b = torch.randn(4, 4, 4)
```

I think that might have caused an issue, since the intermediate dimension should be the same for both, and the linear module only changes the size of the last dimension.

NarineK: Screen Shot 2022-02-21 at 12 30 22 PM

That example is failing for me. Maybe it is dependent  on PT version ?

vivekmig: Interesting, yes, from the documentation, seems like this might not be supported for versions before torch 1.8. I will double check and update the documentation, thanks for catching this!

PR #861, opened 2022-02-11
Optim-wip: Fix issue with `get_model_layers`

ProGamerGov:
There was an issue where multiple separators were not used, and thus a layer like: `layer4[5].relu3` was incorrectly listed as `layer4.[5].relu3` (with a period between the list index and the module name, rather than no separator).

### DIFF
--- a/captum/optim/models/_common.py
+++ b/captum/optim/models/_common.py
@@ -16,19 +16,16 @@ def get_model_layers(model: nn.Module) -> List[str]:
     """
     layers = []
 
-    def get_layers(net: nn.Module, prefix: List = []) -> None:
-        if hasattr(net, "_modules"):
-            for name, layer in net.named_children():
-                if layer is None:
-                    continue
-                name_str = "".join(
-                    [
-                        ("[" + str(i) + "]" if str(i).isdigit() else "." + i)
-                        for i in prefix + [name]
-                    ]
-                ).strip()
-                layers.append(name_str[1:] if name_str[0] == "." else name_str)
-                get_layers(layer, prefix=prefix + [name])
+    def get_layers(net: nn.Module, prefix: str = "") -> None:
+        for name, layer in net.named_children():
+            delimiter = "." if prefix else ""
+            name_str = (
+                f"{prefix}[{name}]"
+                if str(name).isdigit()
+                else f"{prefix}{delimiter}{name}"
+            )
+            layers.append(name_str)
+            get_layers(layer, prefix=name_str)
 
     get_layers(model)
     return layers
--- a/captum/optim/models/_common.py
+++ b/captum/optim/models/_common.py
@@ -18,12 +18,16 @@ def get_model_layers(model: nn.Module) -> List[str]:
 
     def get_layers(net: nn.Module, prefix: List = []) -> None:
         if hasattr(net, "_modules"):
-            for name, layer in net._modules.items():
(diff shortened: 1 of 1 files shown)


captum/optim/models/_common.py
```diff
@@ -21,9 +21,13 @@ def get_layers(net: nn.Module, prefix: List = []) -> None:
...
-                separator = "" if str(name).isdigit() else "."
-                name = "[" + str(name) + "]" if str(name).isdigit() else name
-                layers.append(separator.join(prefix + [name]))
+                name_str = "".join(
+                    [
+                        ("[" + str(i) + "]" if str(i).isdigit() else "." + i)
+                        for i in prefix + [name]
+                    ]
+                ).strip()
+                layers.append(name_str[1:] if name_str[0] == "." else name_str)
```
aobo-y, line 30: do you think we can use the string `name_str` as the `prefix` argument? so no need to repeat the `list -> str` conversion for the prefix part in each recursive call.

ProGamerGov: I don't think that we can, as the naming requires knowing the name of the module that the current module is inside.

aobo-y: I mean sth like this, where the `prefix` contains the chain of modules
```py
def get_layers(net: nn.Module, prefix:str = "") -> None:
  for name, layer in net._modules.items():
    delimiter = "." if prefix else ""
    name_str = f"{prefix}[{name}]" if str(name).isdigit() else f"{prefix}{delimiter}{name}"
    layer_append(name_str)
    get_layers(layer, prefix=name_str)
```

`prefix` sounds more naturally as a variable for string and seems we do not need to save the list. Let me know if I miss sth

captum/optim/models/_common.py
```diff
@@ -21,9 +21,13 @@ def get_layers(net: nn.Module, prefix: List = []) -> None:
             for name, layer in net._modules.items():
```
aobo-y, line 21: btw, do we assume every module has a name? What if the name is `None`? do we want to handle it like `layer1.None.layer3`

ProGamerGov: @aobo-y Looking at the `nn.Module` class, using `None` for the name does not appear to be possible: https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L370

The variable has to be `torch._six.string_classes`, and one cannot do `None = nn.Module()` for initial creation of the module.

aobo-y: @ProGamerGov then we should use `named_children()` https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_children

This is the public api to get the immediate children. Underneath, it does the same as your implementation https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L1700 , but we can avoid accessing the private property while no need to check `layer is None`

ProGamerGov: Okay, I've replaced `net._modules.items()` with `net.named_children()`! 

I think the usage of the private property might have been a holdover from back before the `named_children()` function existed.

captum/optim/models/_common.py
ProGamerGov: @aobo-y Ah, I see, and your code suggestion looks good! I've replaced the existing code with it.

aobo-y:
🚀

PR #863, opened 2022-02-12
Optim-wip: Move inception test helpers to separate file

ProGamerGov:
This change will make it easier to use these helpers for testing other models.

### DIFF
--- a/tests/optim/helpers/models.py
+++ b/tests/optim/helpers/models.py
@@ -3,27 +3,10 @@ from typing import Type
 import torch
 
 
-def _check_layer_in_model(
-    self,
-    model: torch.nn.Module,
-    layer: Type[torch.nn.Module],
-) -> None:
-    def check_for_layer_in_model(model, layer) -> bool:
-        for name, child in model._modules.items():
-            if child is not None:
-                if isinstance(child, layer):
-                    return True
-                if check_for_layer_in_model(child, layer):
-                    return True
-        return False
-
-    self.assertTrue(check_for_layer_in_model(model, layer))
-
-
-def _check_layer_not_in_model(
-    self, model: torch.nn.Module, layer: Type[torch.nn.Module]
-) -> None:
-    for name, child in model._modules.items():
-        if child is not None:
-            self.assertNotIsInstance(child, layer)
-            _check_layer_not_in_model(self, child, layer)
+def check_layer_in_model(model: torch.nn.Module, layer: Type[torch.nn.Module]) -> bool:
+    for _, child in model._modules.items():
+        if child is None:
+            continue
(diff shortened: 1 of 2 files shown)


tests/optim/helpers/models.py
```diff
@@ -0,0 +1,29 @@
...
+
+import torch
+
+
+def _check_layer_in_model(
+    self,
+    model: torch.nn.Module,
+    layer: Type[torch.nn.Module],
+) -> None:
+    def check_for_layer_in_model(model, layer) -> bool:
```
aobo-y, line 11: If you agree with the above suggestion, we don't need another nested inner function here.

```py
def check_layer_in_model(model, layer):
  for _, child in model._modules.items():
    if child is None:
      continue
    if isinstance(child, layer) or check_layer_in_model(child, layer):
      return True
  return False
```

tests/optim/helpers/models.py
```diff
@@ -0,0 +1,29 @@
...
+                if isinstance(child, layer):
+                    return True
+                if check_for_layer_in_model(child, layer):
+                    return True
+        return False
+
+    self.assertTrue(check_for_layer_in_model(model, layer))
+
+
+def _check_layer_not_in_model(
```
aobo-y, line 23: we may not even need this `layer_not_in_model` function, just `not check_layer_in_model(model, layer)`

ProGamerGov: @aobo-y 

In addition to using `not`, we can also use `self.assertFalse`

```
self.assertFalse(check_layer_in_model(model, layer))
self.assertTrue(check_layer_in_model(model, layer))
```

tests/optim/helpers/models.py
```diff
@@ -0,0 +1,29 @@
+from typing import Type
+
+import torch
+
+
+def _check_layer_in_model(
+    self,
```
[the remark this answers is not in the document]
ProGamerGov: The original idea was to follow after `assertTensorAlmostEqual`'s input variable setup, but it does appear simpler to return bool outputs instead.

tests/optim/helpers/models.py
```diff
@@ -0,0 +1,29 @@
+from typing import Type
+
+import torch
+
+
+def _check_layer_in_model(
```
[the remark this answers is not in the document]
ProGamerGov: Oops, yeah I'm not sure how I missed removing the underscore!

tests/optim/helpers/models.py
ProGamerGov: Yeah, we can just use the one function with bool outputs to test for the presence and absence of certain layers.

PR #885, opened 2022-03-03
add tracincp tutorial, logging

99warriors:
Summary:
- add tutorial to website
- add logging

Differential Revision: D34621342

### DIFF
--- a/captum/influence/_core/tracincp_fast_rand_proj.py
+++ b/captum/influence/_core/tracincp_fast_rand_proj.py
@@ -17,6 +17,7 @@ from captum.influence._utils.nearest_neighbors import (
     NearestNeighbors,
     AnnoyNearestNeighbors,
 )
+from captum.log import log_usage
 from torch import Tensor
 from torch.nn import Module
 from torch.utils.data import DataLoader, Dataset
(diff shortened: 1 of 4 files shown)


captum/influence/_core/tracincp_fast_rand_proj.py
```diff
@@ -663,6 +664,7 @@ def _self_influence(self):
         )
         raise NotImplementedError
 
+    @log_usage()
```
NarineK, line 667: I think that since this method is not even abstract I'm skeptical that the logging will work for child classes. @vivekmig, @aobo-y  what do you think ?

aobo-y: I read through the code and actually feel it should word. Abstract or not does not matter. The `log_usage` can decorate any functions. But of course, juts based on my brain parsing. @99warriors  have you tested the logging internally?

NarineK: yeah, it looks like we have the definitions only in the abstract class. I kept thinking about the attributions, where we have overrides. It probably should be fine but it would be good to test.

Issue #792, opened 2021-10-19
Support for Identity layer in LRP


soupault:
## 🐛 Bug

Hi! Thank you for the great tool!

It appears that LRP fails if the model includes `Identity` layer (https://pytorch.org/docs/stable/generated/torch.nn.Identity.html):

```python
  File "???.py", line 523, in my_fn
    attrs = lrp.attribute(inputs=xs,
  File "???/python3.8/site-packages/captum/log/__init__.py", line 35, in wrapper
    return func(*args, **kwargs)
  File "???/python3.8/site-packages/captum/attr/_core/lrp.py", line 183, in attribute
    self._check_and_attach_rules()
  File "???/python3.8/site-packages/captum/attr/_core/lrp.py", line 283, in _check_and_attach_rules
    raise TypeError(
TypeError: Module type <class 'torch.nn.modules.linear.Identity'> is not supported.No default rule defined.
```

## Expected behavior

To my understanding, `Identity` should act as no-op for LRP. It seems currently it is just missing from the sanitizing function (https://github.com/pytorch/captum/blob/9b33f13aabdb0efe2d5404dfe6166d2c8c5d2a7c/captum/attr/_core/lrp.py#L267-L289). To solve the issue, I believe the layer should be handled within this branch - https://github.com/pytorch/captum/blob/9b33f13aabdb0efe2d5404dfe6166d2c8c5d2a7c/captum/attr/_core/lrp.py#L279-L280 .

## Environment
Describe the environment used for Captum

```
 - Captum / PyTorch Version (e.g., 1.0 / 0.4.0): 1.8.2 / 0.4.0
 - How you installed Captum / PyTorch (`conda`, `pip`, source): conda

soupault:
@bilalsal absolutely! I will work on it in the coming days.

PR #889, opened 2022-03-05
Add Identity layer handling to LRP

soupault:
This is one of the better implementations I have come up with. Let me know your thoughts.

Closes https://github.com/pytorch/captum/issues/792.

### DIFF
--- a/captum/attr/_core/lrp.py
+++ b/captum/attr/_core/lrp.py
@@ -408,6 +408,7 @@ class LRP(GradientAttribution):
 
 
 SUPPORTED_LAYERS_WITH_RULES = {
+    nn.Identity: EpsilonRule,
     nn.MaxPool1d: EpsilonRule,
     nn.MaxPool2d: EpsilonRule,
     nn.MaxPool3d: EpsilonRule,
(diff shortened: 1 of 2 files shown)


captum/attr/_core/lrp.py
```diff
@@ -408,6 +408,7 @@ def _forward_fn_wrapper(self, *inputs: Tensor) -> Tensor:
 
 
 SUPPORTED_LAYERS_WITH_RULES = {
+    nn.Identity: lambda: EpsilonRule(epsilon=0.0),
```
NarineK, line 411: Thank you for  the PR, @soupault! I have  a question regarding epsilon=0.0.

 It can be that the output of the identity is 0  and we might hit division by zero ?

https://github.com/pytorch/captum/blob/7c08d06428f867e0a487c092b06f06514c498cbe/captum/attr/_utils/lrp_rules.py#L64

NarineK: If we want no-op for Identity we could potentially add it a list like `SUPPORTED_NON_LINEAR_LAYERS ` ?

soupault: Thank you @NarineK for the review! Division by 0 is definitely possible, yet I would say rather unrealistic when Identity is used with any other layer and real data. In any case, better to set the rule to the common Epsilon. No-op is surely the best fit here, but I feel it might be confusing for further contributors to have the layer under `_NON_LINEAR_`.

PR #891, opened 2022-03-08
Run influence tests using annoy.

i-jones:
Summary:
Some of the influence tests require the annoy package to run. This change adds the annoy dependencies so these tests are run instead of skipped.
This also adds annoy as a dependency for the captum bento kernel.

Differential Revision: D34729957

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ DEV_REQUIRES = (
         "usort==0.6.4",
         "ufmt",
         "scikit-learn",
+        "annoy",
     ]
 )
 


NarineK:
@aobo-y, @vivekmig , do you mind confirming this change ?

setup.py
```diff
@@ -61,7 +61,7 @@ def report(*args):
 
 TUTORIALS_REQUIRES = INSIGHTS_REQUIRES + ["torchtext", "torchvision"]
 
-TEST_REQUIRES = ["pytest", "pytest-cov"]
```
aobo-y, line 64: Hi @i-jones , I feel `annoy` is not just a test deps like `pytest`. It is used in our [main codebase](https://github.com/pytorch/captum/blob/ed49c1bbbdd4a880b5c39a4177e73de7cc7521eb/captum/influence/_utils/nearest_neighbors.py#L103) to provide functionality as well (cc @99warriors to confirm).

We did not make it [mandatory](https://github.com/pytorch/captum/blob/master/setup.py#L149) because it is not widely needed and we try to keep our default package small. Users who need it may manually install it.

These optional packages (the other one is `scikit-learn`) are needed in development. So maybe add it to https://github.com/pytorch/captum/blob/master/setup.py#L69 instead? Our testing ci will install all packages within ["develop"](https://github.com/pytorch/captum/blob/ed49c1bbbd/scripts/install_via_pip.sh#L47)

PR #895, opened 2022-03-11
update CI linting and testing jobs

aobo-y:
- extract linting related steps (flake8, ufmt, sphinx) to a dedicated ci job & only run it once as they are unlikely to be impacted by different pytorch version
- remove linting & testing from the job for website deployment
- remove no longer needed installation of linting deps in conda

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,7 +161,6 @@ jobs:
       - checkout
       - pip_install:
           args: "-v 1.6"
-      - mypy_check
       - unit_tests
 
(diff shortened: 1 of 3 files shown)


aobo-y:
@vivekmig changed CI as discussed in this PR except `black target_version` and `mypy`

setting `black target_version` to `py36` caused some new errors of linting, fixed in a separate pr https://github.com/pytorch/captum/pull/896

adding `mypy` to other pytorch versions will also cause a lot of new errors, better leave it to separate task

Issue #680, opened 2021-06-11
Strange import issue --> AttributeError: module 'captum' has no attribute 'attr'


adaruna3:
## 🐛 AttributeError: module 'captum' has no attribute 'attr'

When importing captum with just `import captum`, the package is missing modules like attr. Running `dir(captum)` shows that the other modules, like `attr` are missing. However, if a statement like `from captum.attr._utils.input_layer_wrapper import ModelInputWrapper` is included after `import captum`, the captum package then has the missing modules. Again, verified by running a `dir(captum)`. Is this expected behavior?

## To Reproduce

1. Run the code below:
```
import captum
print(dir(captum))
print("---------------")
from captum.attr._utils.input_layer_wrapper import ModelInputWrapper
print(dir(captum))
exit()
```
1. The above should get an output of the following:
```
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__']
---------------
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_utils', 'attr', 'log']
```

## Expected behavior

I assume all modules should be available after `import captum`. If this behavior is not expected, please close and disregard.

## Environment

I installed captum using the suggested conda method. See details below. After noticing this issue in an environment I was developing in with many other packages, I made a new, clean environment that only contained captum and any dependencies conda installed. Same issue.

 - Captum / PyTorch Version (e.g., 1.0 / 0.4.0): 0.3.1
 - OS (e.g., Linux): Ubuntu 18.04.4 LTS
 - How you installed Captum / PyTorch (`conda`, `pip`, source): conda
 - command: `conda install captum -c pytorch`
 - Python version: 3.6.13

NarineK:
Thank you for raising the issue, @adaruna3 ! We will look into it.

vivekmig:
Hi @adaruna3 , I think this is likely expected behavior, since importing a module doesn't automatically import all submodules. This would only occur if the main captum __init__ imported the specific submodules, which is not currently the case, so you would need to either do `import captum.attr` or `from captum.attr import *` to import the attr submodule. These links also discuss similar issues ([1](https://stackoverflow.com/questions/23829845/dirpackage-not-listing-out-containing-modules-in-python?rq=1), [2](https://stackoverflow.com/questions/57461671/dir-function-not-showing-all-package-contents))

[...]

PR #912, opened 2022-03-24
Add support for captum toplevel-import.

dkrako:
This fixes issue #680 Strange import issue --> AttributeError: module 'captum' has no attribute 'attr'

In most python packages, you can import the toplevel package, like numpy, scipy, torch, etc.. and then access the submodules simply by the dot-operator. Like you can use `import numpy` and after that you can use any submodules à la `numpy.random.uniform`.

With this PR, you can just `import captum` and then for example use `captum.attr.DeepLift` or `captum.robust.Perturbation` instead of having to import both. It's just a small convenience, and I think there are more people that expect this kind of import to work but don't bother to create an issue out of this.

I hope this PR is considered as helpful.

### DIFF
--- a/captum/__init__.py
+++ b/captum/__init__.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
-import captum.attr as attr
-import captum.concept as concept
-import captum.influence as influence
-import captum.insights as insights
-import captum.log as log
-import captum.metrics as metrics
-import captum.robust as robust
+import captum.attr as attr  # noqa
+import captum.concept as concept  # noqa
+import captum.influence as influence  # noqa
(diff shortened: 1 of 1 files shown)


NarineK:
Thank you for the PR, @dkrako! @aobo-y, @vivekmig do you mind checking whether there is import runtime performance related issues with this change ?

vivekmig:
Thanks so much for adding this @dkrako ! 

> Thank you for the PR, @dkrako! @aobo-y, @vivekmig do you mind checking whether there is import runtime performance related issues with this change ?

This seems fine, ```import captum``` is slower, but comparable to importing captum.attr currently, so should be fine considering the benefits of having this import.

PR #919, opened 2022-04-07
Optim-wip: Fix duplicated target bug

ProGamerGov:
Currently a hook is created in `ModuleOutputsHook` for ever instance of a target in the target list. Each captured set of activations for the same hook overwrites the previous set of activations, potentially leading to negative performance impacts. This bug also causes the of the warning messages in `ModuleOutputsHook` to repeat every iteration.

This PR solves the issue by ensuring that duplicate target values are removed.

### DIFF
--- a/captum/optim/_core/loss.py
+++ b/captum/optim/_core/loss.py
@@ -130,6 +130,9 @@ def module_op(
         target = (self.target if isinstance(self.target, list) else [self.target]) + (
             other.target if isinstance(other.target, list) else [other.target]
         )
+
+        # Filter out duplicate targets
+        target = list(dict.fromkeys(target))
     else:
         raise TypeError(
             "Can only apply math operations with int, float or Loss. Received type "


ProGamerGov:
The `test_py37_conda` test failed due to a stochastic error, and thus it can be ignored.

captum/optim/_core/loss.py
```diff
@@ -129,6 +129,9 @@ def loss_fn(module: ModuleOutputMapping) -> torch.Tensor:
         target = (self.target if isinstance(self.target, list) else [self.target]) + (
             other.target if isinstance(other.target, list) else [other.target]
         )
+
+        # Filter out duplicate targets
+        target = list(dict.fromkeys(target))
```
NarineK, line 134: @ProGamerGov, why would someone pass duplicated target here ? Shouldn't we set an assert here ?

ProGamerGov: @NarineK There are a few reason why there can be duplicates here.

For example, optimization with transparency will be using `NaturalImage` or a transform as the target for one or more alpha channel related objectives. If the user is working with a CLIP model, an L2 penalty objective will also be using one of the same targets.

ProGamerGov: Using multiple different penalties on the same target will also create duplicates without that line.

NarineK: @ProGamerGov, are you saying that in this line the targets will be duplicated (example from Optimizing with Transparency Notebook) ? 
```
loss_fn = loss_fn - MeanAlphaChannelPenalty(transforms[0])
loss_fn = loss_fn - (9 * BlurActivations(transforms[0], channel_index=3))

``` 

Is it because we are concatenating the other target to the current target here ?
```
        target = (self.target if isinstance(self.target, list) else [self.target]) + (
            other.target if isinstance(other.target, list) else [other.target]
        )
```
I was thinking why are we concatenating the targets in the above line ?
It looks like we are not concatenating if `self.target` is a list but otherwise we concatenate it with `other.target` . I was wondering if we could elaborate this logic a bit.

ProGamerGov: @NarineK Yes, that those lines will result in a duplicated target because we concatenate the target lists for every operation involving multiple loss objectives. I'll about doing a more detailed write-up of how it works in another PR.

NarineK: @ProGamerGov, but we don't concatenate them if  `self.target` is a list ? You can perhaps rework this PR since it is very small or we need to document that the logic requires refinement in this PR before we merge it.

ProGamerGov: @NarineK It gets a bit complicated, but it might be a bit more complicated to change at the moment. Most loss objectives store a `self.target` value that is then called to collect the target activations: 

```
    def __call__(self, targets_to_values: ModuleOutputMapping) -> torch.Tensor:
        activations = targets_to_values[self.target]
        return activations
```

The `self.target` value them becomes a list when combined with another loss objective in a resulting `CompositeLoss` instance. The two original objectives can still call and use their own `self.target` value. The `ModuleOutputsHook` module and `InputOptimization` module both use a list of targets as well, but `InputOptimization` does not overwrite the original `self.target` to make it's list.

PR #922, opened 2022-04-08
Optim-wip: Correct `RandomScale` transform

ProGamerGov:
The `RandomScale` transform should actually be using `F.interpolate` rather than ` F.affine_grid` & `F.grid_sample`, so I've changed it. The old version of `RandomScale` however seems to be useful enough to warrant keeping it, so I've renamed it to `RandomScaleAffine`.

* The `_rand_select` function is no longer being used by any transforms, so I've deleted it.

* Replaced default `RandomScale` transform with Interpolation based variant. Renamed the old variant to `RandomScaleAffine`.

* Added JIT support for both versions of `RandomScale`.  This required changing some of the type hints.

* Add distributions support to both versions of `RandomScale`. Ludwig wanted this based off of his original PR.

* Added more tests to improve testing coverage.

* Changed `NumSeqOrTensorType` hint to `NumSeqOrTensorOrProbDistType`.

This PR contains code that was part of: https://github.com/pytorch/captum/pull/821, to make it easier to review.


ProGamerGov:
Closing this PR as it is no longer needed

PR #935, opened 2022-05-08
Optim-wip: Add the pre-trained InceptionV1 Places365 model

ProGamerGov:
See: https://github.com/pytorch/captum/pull/655 for more details on this PR


NarineK:
@ProGamerGov, is it possible to also include the test fixes ? CircleCI is now failing.

ProGamerGov:
@NarineK I fixed that black linting errors. The mypy test in the in the CircleCI test_py36_pip_release will fail though as there is a somewhat complex issue with type hinting in one of the other Captum modules.

captum/attr/_core/noise_tunnel.py
```diff
@@ -205,7 +205,7 @@ def compute_expected_attribution_and_sq(attribution):
 
             attribution = attribution.view(attribution_shape)
             expected_attribution = attribution.mean(dim=1, keepdim=False)
-            expected_attribution_sq = torch.mean(attribution ** 2, dim=1, keepdim=False)
+            expected_attribution_sq = torch.mean(attribution**2, dim=1, keepdim=False)
```
NarineK, line 208: @ProGamerGov, this PR has also formatted the files in the non optim related code. Perhaps you can bring this back. Also, maybe you have a different version of code formatter that's why it changed the formatting ?

NarineK:
@ProGamerGov, let's remove the changes that aren't related to optim and copy the fix related to flask. Is this flask error related to flask version ?

captum/attr/_core/noise_tunnel.py
ProGamerGov: @NarineK The black tests fail currently because the optim-wip branch's version of black was never fixed to a specific version. I've reverted the changes.

Issue #936, opened 2022-05-09
DeprecationWarning: Importing display from IPython.core.display is deprecated


AndreasGerken:
The way display is imported in IPython should be changed.

Bug in [visualization.py:15](https://github.com/pytorch/captum/blob/1c5bd8a3c0a974e9773ea5787b61bf0a368b084f/captum/attr/_utils/visualization.py#L15):
```
captum/attr/_utils/visualization.py:15: DeprecationWarning: Importing display from IPython.core.display is deprecated since IPython 7.14, please import from IPython display
  from IPython.core.display import HTML, display
```

The Line in visualization should be changed from 
```
from IPython.core.display import display, HTML
```
to
```
from IPython.display import display, HTML
```

Using IPython 8.3.0 in Python 3.9.12.

vivekmig:
Thanks for identifying this issue @AndreasGerken ! We will push a fix for this issue shortly.

YudiGuzman:
Thanks @AndreasGerken, for sharing how to fix this issue.

PR #937, opened 2022-05-11
Update IPython import to resolve deprecation warning

vivekmig:
Address #936 by switching IPython import from IPython.core.display to from IPython.display


NarineK (approved):
LGTM! Thank you for the fix!

PR #940, opened 2022-05-12
Fix version check bug

ProGamerGov:
By default: `"1.8.0" > "1.10.0"` will be equal to True, despite 1.10 being a later version that 1.8.0. This PR fixes this issue.


vivekmig (approved):
Looks great, thanks for this fix @ProGamerGov !

PR #969, opened 2022-06-06
Add gpu support to tracincp rand projection

NarineK:
Adds gpu support to tracincp rand projection. 
Cleaned up un-passed args to _load_flexible_state_dict


vivekmig (approved):
Looks great :) !

PR #974, opened 2022-06-11
Fix `set_all_random_seeds` testing function

ProGamerGov:
A seed value was being passed to the function, but it was ignored. This PR fixes that.


vivekmig (approved):
Thanks so much for fixing this @ProGamerGov !

PR #535, opened 2020-11-21
Add missing type hints to master in accordance with PEP-0484

ProGamerGov:
While working on the Optim-WIP branch I discovered that some of the `__init__` functions lacked the `-> None` type hints that [PEP-0484](https://www.python.org/dev/peps/pep-0484/) specifies. It seemed pretty easy to automate a fix, so I went ahead and added them after figuring out the regex code I needed.

~~The failed test looks like it's unrelated to my changes. Mypy didn't show any issues when I ran it locally.~~

PR #976, opened 2022-06-11
Add missing type hints in accordance with PEP-0484

ProGamerGov:
Adding `-> None` to `__init__` functions is specified by [PEP-0484](https://www.python.org/dev/peps/pep-0484/) and is really easy to do with regex, so I went ahead and did it for the `captum/` & `tests/` directories.

I used this to do it: https://stackoverflow.com/questions/64948233/how-to-add-none-to-the-end-of-init-functions-with-notepad

If this PR seems a bit familiar, it's because I submitted a similar one 2 years ago: https://github.com/pytorch/captum/pull/535

### DIFF
--- a/captum/_utils/av.py
+++ b/captum/_utils/av.py
@@ -47,7 +47,7 @@ class AV:
             identifier: Optional[str] = None,
             layer: Optional[str] = None,
             num_id: Optional[str] = None,
-        ):
+        ) -> None:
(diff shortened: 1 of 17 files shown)


NarineK:
Thank you, @ProGamerGov

NarineK (approved):
LGTM

Issue #988, opened 2022-06-30
Manual install fails due to missing insight dependency


ProGamerGov:
## 🐛 Bug

Manual install fails due to missing dependencies.

## To Reproduce

Start without `flask_compress` preinstalled.

Install manually:

```
git clone https://github.com/pytorch/captum.git
cd captum
pip install -e .
```

They try running:

```
import captum.attr
```

And you get the following error.

```
---------------------------------------------------------------------------

ModuleNotFoundError                       Traceback (most recent call last)

[<ipython-input-3-5589eed498a3>](https://localhost:8080/#) in <module>()
----> 1 import captum.attr

4 frames

[/content/captum/captum/insights/attr_vis/server.py](https://localhost:8080/#) in <module>()
      9 from captum.log import log_usage
     10 from flask import Flask, jsonify, render_template, request
---> 11 from flask_compress import Compress
     12 from torch import Tensor
     13 

ModuleNotFoundError: No module named 'flask_compress'


---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.
---------------------------------------------------------------------------
```

This issue seems to come from this PR: https://github.com/pytorch/captum/pull/912

ProGamerGov:
Possible fixes would be removing the insights import from the main `__init__.py` file, or changing the Insight module's `__init__.py` file to this:

```
from warnings import warn

try:
    from captum.insights.attr_vis import AttributionVisualizer, Batch  # noqa
except (ImportError, AssertionError) as e:
    warning_msg = (
        "Missing required libraries for Captum's Insights module."
        + " This warning can be ignored if not using Insights."
        + " {}".format(e)
    )
    warn(warning_msg)
```

PR #992, opened 2022-07-04
Remove the `insights` module from the main `__init__.py` file

ProGamerGov:
Fixes: https://github.com/pytorch/captum/issues/988

According the `setup.py` and the `README`, users wishing to use insights need to use the custom install option provided or install the modules separately. https://github.com/pytorch/captum/blob/master/setup.py#L54, https://github.com/pytorch/captum/blob/master/README.md#installation

Therefore the insights module should not be loaded in the `__init__.py` file, and users will have to call it like before the changes proposed in https://github.com/pytorch/captum/pull/912, and added to the master branch in https://github.com/pytorch/captum/commit/9305b109417ca24ee6893075e03f3da241e59252

### DIFF
--- a/captum/__init__.py
+++ b/captum/__init__.py
@@ -2,7 +2,6 @@
 import captum.attr as attr  # noqa
 import captum.concept as concept  # noqa
 import captum.influence as influence  # noqa
-import captum.insights as insights  # noqa
 import captum.log as log  # noqa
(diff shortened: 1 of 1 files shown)


NarineK (approved):
Thank you, @ProGamerGov ! This was a recent addition - We didn't see the issue.

PR #994, opened 2022-07-14
modify tracin self influence helpers

99warriors:
Summary:
change `TracInCP._self_influence_batch_tracincp` and `TracInCP._self_influence_batch_tracincp` `TracInCP._self_influence_batches_tracincp_fast` to be named `self_influence`, which is now public, and now accept a DataLoader yielding batches (as well as a single batch, as before).  The modified helper function can be called by external functions to compute self influence.

The helper itself is also changed to improve efficiency, by reducing the number of times checkpoints are loaded.  The modified helper, despite being able to compute self influence scores for a dataloader yielding batches, still only loads each checkpoint once, per call.  This is because the modified helper now has an outer iteration over checkpoints, and an inner iteration over batches (the order of iteration is reversed compared to before). This helper is called by `influence` when running it in self influence mode.

The reason we cannot just increase the batch size to reduce the number of checkpoint loadings is that for large models (precisely those for which loading checkpoints is expensive), the model takes up too much memory, so that the batch size cannot be too large.

Minor change: for `influence_src_dataset` argument of all `__init__`'s, add description of what assumptions we make of the batches yielded by the dataloader.

Reviewed By: NarineK

Differential Revision: D35603078

### DIFF
--- a/captum/influence/_core/tracincp.py
+++ b/captum/influence/_core/tracincp.py
@@ -26,6 +26,7 @@ from captum._utils.gradient import (
 from captum._utils.progress import progress
 from captum.influence._core.influence import DataInfluence
 from captum.influence._utils.common import (
+    _format_inputs_dataset,
     _get_k_most_influential_helper,
     _gradient_dot_product,
     _load_flexible_state_dict,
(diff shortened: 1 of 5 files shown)


NarineK:
cc: @99warriors it looks like some of the tests related to progress bar are failing. Do you mind looking into it ?

captum/influence/_core/tracincp.py
```diff
@@ -95,7 +96,7 @@ class TracInCPBase(DataInfluence):
     def __init__(
         self,
         model: Module,
-        influence_src_dataset: Union[Dataset, DataLoader],
+        train_dataset: Union[Dataset, DataLoader],
```
NarineK, line 99: @99warriors, do you mind fixing this naming also in the tutorials ? I remember that we used them explicitly as argument names.

PR #995, opened 2022-07-15
Don't link directly to arXiv PDF files in algorithms.md

ProGamerGov:
arXiv offers multiple options for viewing a paper, so it's best to let the user decide which one they want.


NarineK (approved):
Thank you for the fix!

PR #996, opened 2022-07-16
Fix some docstrings

ProGamerGov:
The "optional" part should come last.


NarineK (approved):
LGTM! Thank you for fixing it

PR #985, opened 2022-06-30
Fix multiple Sphinx warnings & docstrings

ProGamerGov:
The warning messages take up a lot of space on the console log, and it was really simple to resolve them.

The common.rst file was also incorrectly pointing to the wrong path and some of functions were renamed since it was created, so no docs were being generated for that page. `InputBaselineXGradient` was also removed from the public rst api docs, as it's not supposed to be public.

In addition to these easy doc warning fixes, I found a module that was listed on on the API docs site, but there's not public path to use it and no tests were ever written for it. I made an issue post for it here: https://github.com/pytorch/captum/issues/989

I fixed some docstring issues like lack of consistent uppercase for `any` and `callable`, spacing, random type / case mistakes I came across, etc...

I also fixed some paths.

Issues with upgrading to later versions of Sphinx were also resolved.

---

Currently Sphinx gives the following warnings / errors for the master branch:

```
/content/captum/sphinx/source/base_classes.rst:2: WARNING: Title underline too short.

Base Classes
==========
/content/captum/sphinx/source/base_classes.rst:29: WARNING: Title underline too short.

Perturbation Attribution
^^^^^^^^^^^^^^^^^^^^^
/content/captum/sphinx/source/base_classes.rst:29: WARNING: Title underline too short.

Perturbation Attribution
^^^^^^^^^^^^^^^^^^^^^
WARNING: autodoc: failed to import function 'validate_input' from module 'captum.attr._utils.common'; the following exception was raised:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/sphinx/util/inspect.py", line 448, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'captum.attr._utils.common' has no attribute 'validate_input'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/sphinx/ext/autodoc/importer.py", line 110, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/usr/local/lib/python3.7/dist-packages/sphinx/ext/autodoc/__init__.py", line 332, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/usr/local/lib/python3.7/dist-packages/sphinx/ext/autodoc/__init__.py", line 2780, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
[...]

PR #999, opened 2022-07-27
Improve version checking

ProGamerGov:
Without the packaging library, statements like: `"1.8.0" > "1.10.0"` will be equal to True, despite v1.10 being a later version that v1.8.0.

The `packaging` library will in some cases not be already installed on a user's device, so I've also added it the `setup.py` file. It's one of the core libraries from the Python Packaging Authority, but it's not included with the base Python installation: https://packaging.python.org/en/latest/key_projects/#pypa-projects

This wasn't an issue in https://github.com/pytorch/captum/pull/940 as one the libraries in dev install has `packaging` as a dependency. So, there's no error when the tests are using the `packaging` library.

### DIFF
--- a/captum/_utils/common.py
+++ b/captum/_utils/common.py
@@ -3,7 +3,7 @@ import typing
 from enum import Enum
 from functools import reduce
 from inspect import signature
-from typing import Any, Callable, cast, Dict, List, Optional, overload, Tuple, Union
+from typing import Any, Callable, cast, Dict, List, overload, Tuple, Union
 
 import numpy as np
 import torch
@@ -18,7 +18,7 @@ from torch import device, Tensor
 from torch.nn import Module
 
 
-def _parse_version(v: str, length: Optional[int] = 3) -> Tuple[int, ...]:
+def _parse_version(v: str) -> Tuple[int, ...]:
     """
     Parse version strings into tuples for comparison.
 
@@ -29,10 +29,6 @@ def _parse_version(v: str, length: Optional[int] = 3) -> Tuple[int, ...]:
     Args:
 
         v (str): A version string.
-        length (int, optional): The expected length of the output tuple. If the output
-            is less than the expected length, then it will be padded with 0 values. Set
-            to None for no padding or length checks.
-            Default: ``3``
 
     Returns:
         version_tuple (tuple of int): A tuple of integer values to use for version
@@ -40,9 +36,6 @@ def _parse_version(v: str, length: Optional[int] = 3) -> Tuple[int, ...]:
     """
     v = [n for n in v.split(".") if n.isdigit()]
     assert v != []
-    if length is not None:
-        v += ["0"] * (length - len(v))
-        assert len(v) == length
     return tuple(map(int, v))
 
 
--- a/captum/_utils/common.py
+++ b/captum/_utils/common.py
@@ -14,6 +14,7 @@ from captum._utils.typing import (
     TensorOrTupleOfTensorsGeneric,
     TupleOrTensorOrBoolGeneric,
 )
+from packaging import version
 from torch import device, Tensor
 from torch.nn import Module
 
@@ -671,7 +672,7 @@ def _register_backward_hook(
     ):
         return module.register_backward_hook(hook)
 
-    if torch.__version__ >= "1.9":
+    if version.parse(torch.__version__) >= version.parse("1.9.0"):
         # Only supported for torch >= 1.9
         return module.register_full_backward_hook(hook)
     else:
--- a/captum/_utils/common.py
+++ b/captum/_utils/common.py
@@ -3,7 +3,7 @@ import typing
 from enum import Enum
 from functools import reduce
 from inspect import signature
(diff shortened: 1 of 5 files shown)


NarineK:
@ProGamerGov if you rebase from master #1000 will hopefully fix failing conda issue

ProGamerGov:
@NarineK It worked on this PR, but I tried it on #985 and it's still running after 3 and half hours

NarineK:
@ProGamerGov, would it be possible to use something like this:
```
def versiontuple(v):
    return tuple(map(int, (v.split("."))))
```
Especially because pytorch's versioning is non-complicated and follows the same pattern.
We usually avoid adding new dependancies to Captum if there is another possible solution without additional dependancies. I understand the advantage of package over tuple comparison but in this case pytorch's versioning is pretty straightforward and non-complex.

https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

ProGamerGov:
@NarineK That solution fails for nightly builds:

```
def versiontuple(v):
    return tuple(map(int, (v.split("."))))

output = versiontuple("1.12.0.dev20201109")
print(output)
```

```
---------------------------------------------------------------------------

ValueError                                Traceback (most recent call last)

[<ipython-input-2-32fcc066bf69>](https://localhost:8080/#) in <module>()
      4 
      5 
----> 6 output = versiontuple("1.12.0.dev20201109")
      7 print(output)

[<ipython-input-2-32fcc066bf69>](https://localhost:8080/#) in versiontuple(v)
      1 def versiontuple(v):
----> 2     return tuple(map(int, (v.split("."))))
      3 
      4 
      5 

ValueError: invalid literal for int() with base 10: 'dev20201109'
```

Though I suppose we could find a way to make it work. Are the other PyTorch libraries like Torchtext using the same versioning pattern?

The `packaging` library is also already essentially a package dependency of Captum as it is installed as a dependency during the dev install, which is why I'm suggesting it.

ProGamerGov:
@NarineK Is `captum/_utils/common.py` where we'd add the version parsing function?

I also think the Conda test might be broken again, as it's failed a couple of times now.

ProGamerGov:
@NarineK I've implemented the `_parse_version` function and supporting tests.

NarineK:
@ProGamerGov, thank you for making the changes. I was looking into the dependency graph of the libs and I noticed that mathplotlib already depends on packaging. If that's the right packaging library that I'm looking into. 

Is there need to add it here ? We already have the dependency in the non-dev setting. 
```
install_requires=["matplotlib", "numpy", "packaging", "torch>=1.6"],
```
 
Screen Shot 2022-08-13 at 11 56 30 PM

ProGamerGov:
@NarineK I just tested installing Captum & Matplotlib, and neither installed the packaging library. Its listed in Matplotlib's [setup.py](https://github.com/matplotlib/matplotlib/blob/main/setup.py), but not their [setupext.py](https://github.com/matplotlib/matplotlib/blob/main/setupext.py) (which I think is used for the general user version of the package). They also use it for testing and some of the optional modules it looks like: https://github.com/matplotlib/matplotlib/search?p=1&q=packaging

captum/_utils/common.py
```diff
@@ -18,6 +18,34 @@
...
+            to None for no padding or length checks.
+            Default: ``3``
+
+    Returns:
+        version_tuple (tuple of int): A tuple of integer values to use for version
+            comparison.
+    """
+    v = [n for n in v.split(".") if n.isdigit()]
+    assert v != []
+    if length is not None:
```
NarineK, line 43: A question: when will the length be useful ? If we always add 0s then it is not going to make big difference?
e.g. Whether we compare `(2,5) >= (2,4,5)` or `(2,5,0) >= (2,4,5)` we get the same outcome.

NarineK:
> @NarineK I just tested installing Captum & Matplotlib, and neither installed the packaging library. Its listed in Matplotlib's [setup.py](https://github.com/matplotlib/matplotlib/blob/main/setup.py), but not their [setupext.py](https://github.com/matplotlib/matplotlib/blob/main/setupext.py) (which I think is used for the general user version of the package). They also use it for testing and some of the optional modules it looks like: https://github.com/matplotlib/matplotlib/search?p=1&q=packaging

Thank you for checking @ProGamerGov, makes sense! I added a question related to the `_parse_version` function ?

captum/_utils/common.py
NarineK: I guess only in this case `(2,4,0) > (2,4)`  and `(2,4) > (2,4)` but is it likely ?

ProGamerGov: @NarineK Yeah, I realize now that we don't need to pad the version tuple with zeros. If there are zeros in the version string, then they'll be used, otherwise we don't need to add them.

I've removed the length option from the function

ProGamerGov:
@NarineK The packaging library is downloaded as a dependency in the Captum dev install, but I'm not sure which package uses it as a dependency. So, we only need the `_parse_version` function for non test related code

NarineK (approved):
Thank you very much for working on this PR, @ProGamerGov!

PR #1001, opened 2022-07-29
Notebook support for tqdm

Meghpal:
The `tqdm` progress in notebooks breaks in some cases (for me this behavior was persistent after I stopped the cell running this even once), possibly because it is not imported from the **recommmended** `tqdm.auto`

```python
from tqdm import tqdm
```
image

However, when imported from `tqdm.auto` it works flawlessly:
```python
from tqdm.auto import tqdm
```
image

### DIFF
--- a/captum/_utils/progress.py
+++ b/captum/_utils/progress.py
@@ -6,7 +6,7 @@ from time import time
 from typing import cast, Iterable, Sized, TextIO
 
 try:
-    from tqdm import tqdm
+    from tqdm.auto import tqdm
 except ImportError:
     tqdm = None
 


aobo-y:
Thank you for doing this! Worth noting though, that `tqdm.auto` is also not perfect https://github.com/tqdm/tqdm/issues/1098

> Note that this will issue a TqdmExperimentalWarning if run in a notebook since it is not meant to be possible to distinguish between jupyter notebook and jupyter console. Use auto instead of autonotebook to suppress this warning.

Considering the users of Jupyter Notebook is much more than Jupyter console, I feel this change should benefit more people

PR #1009, opened 2022-08-09
Fix failing conda test, & fix a GPU test bug

ProGamerGov:
I think that the amount of Conda install timeouts can be lessened by removing unnecessary installs: 

* Sphinx is not used for the Conda tests, so it doesn't make sense to install it.
* NumPy is a dependency of PyTorch, so there is no need to install it after installing PyTorch.

The `nodejs` install seems to be what takes up the most time, and I think its where the solver sometimes fails. Using the libmamba solver seems to prevent it from failing, but it still takes vast majority of the install time (around 20 minutes) to install `nodejs`. https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community

I also noticed a recurring error in the GPU tests, and implemented a fix for it:

```
Errors were encountered while processing:
 sane-utils
W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

This PR along with #1007 should fix the test failures that Captum is experiencing.

@vivekmig

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,7 @@ commands:
             sudo dpkg -i cuda-repo-ubuntu2004-11-4-local_11.4.2-470.57.02-1_amd64.deb
             sudo apt-key add /var/cuda-repo-ubuntu2004-11-4-local/7fa2af80.pub
             sudo apt-get update
+            sudo dpkg --configure -a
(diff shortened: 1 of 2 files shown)


vivekmig (approved):
Looks great, thanks for these fixes @ProGamerGov !

PR #1013, opened 2022-08-15
Assign variables

standardAI:
If USE_PRETRAINED_MODEL == True; then assign variables.


vivekmig (approved):
Looks great, thanks so much for this fix!

Issue #1012, opened 2022-08-14
House Price Interpretation Tutorial notebook fails


jimthompson5802:
## 🐛 Bug

Running the tutorial [Interpret regression models using Boston House Prices Dataset](https://captum.ai/tutorials/House_Prices_Regression_Interpret) results in an `IndexError` in the last cell

## To Reproduce

 
1. This [colab notebook](https://colab.research.google.com/drive/16XpYfmiM3fDed35BMosfxdp3Ly5EMjHv?usp=sharing) demonstrates the error.
1. `pip install captum` in the notebook
1.   uploaded into the notebook container the [pre-trained model](https://github.com/pytorch/captum/blob/master/tutorials/models/boston_model.pt).

```
IndexError                                Traceback (most recent call last)
<ipython-input-19-fd2cf6250225> in <module>()
      1 plt.figure(figsize=(15, 8))
      2 
----> 3 x_axis_data = np.arange(lc_attr_test.shape[1])
      4 
      5 y_axis_lc_attr_test = lc_attr_test.mean(0).detach().numpy()

IndexError: tuple index out of range
SEARCH STACK OVERFLOW
<Figure size 1080x576 with 0 Axes>
```

## Expected behavior

Error free run of the tutorial notebook.

## Environment

Google's colab notebook

```

 - Captum / PyTorch Version (0.5.0 / 1.12.1+cu113):
 - OS (e.g., Linux):
 - How you installed Captum / PyTorch (`conda`, `pip`, source): pip install captum
 - Build command you used (if compiling from source): N/A
 - Python version: 3.7.13
 - CUDA/cuDNN version: N/A
 - GPU models and configuration: colab GPU
 - Any other relevant information:  Error occurs with or without use of GPU

## Additional context

None.

**Comment by dbish** [comment]
Found the issue. There's a line that must have been edited/added at some point that assumes more then one tensor is being returned from lc.attribute, but there's only one (since only one tensor is passed in): 
```
lc_attr_test = lc.attribute(X_test, n_steps=100, attribute_to_layer_input=True)

# shape: test_examples x size_hidden
lc_attr_test = lc_attr_test[0]
```

The second line here of setting lc_attr_test to the 0th index, then sets it to the tensor index instead of the first tensor, which in turn in the next cell means that "lc_attr_test.shape[1]" throws an exception. Fix is to just take out the reassigning of "lc_attr_test = lc_attr_test[0]". I'll submit a PR fixing this.

[...]

PR #1014, opened 2022-08-19
Fixed bug in House_Prices_Regression_Interpret tutorial

dbish:
Described in issue: [1012](https://github.com/pytorch/captum/issues/1012)

**Background**
There's a line that must have been edited/added at some point that assumes more then one tensor is being returned from lc.attribute, but there's only one (since only one tensor is passed in):

lc_attr_test = lc.attribute(X_test, n_steps=100, attribute_to_layer_input=True)

# shape: test_examples x size_hidden
lc_attr_test = lc_attr_test[0]
The second line here of setting lc_attr_test to the 0th index, then sets it to the tensor index instead of the first tensor, which in turn in the next cell means that "lc_attr_test.shape[1]" throws an exception. 

**Changes**
Fix is to just take out the reassigning of "lc_attr_test = lc_attr_test[0]".

**Testing**
Tested running the notebook before the change (which showed the error described) and after, which produced the plot which was originally shown in the static tutorial page and matches expectations now (see attachment)
Screen Shot 2022-08-18 at 9 59 25 PM
.


dbish:
Looks like this tutorial hasn't been updated in a bit, it works now, but there are a few new warnings that didn't show up before in the notebook output without changing any additional code besides the line mentioned. Big one is a function deprecation warning from sklearn https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html. Maybe worth making another update or opening a new task to clean those up.

tutorials/House_Prices_Regression_Interpret.ipynb
```diff
@@ -63,7 +63,48 @@
    "cell_type": "code",
    "execution_count": 2,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
vivekmig, line 66: nit: Can we revert the change in these lines and 474-487? This will avoid including your local machine path in the notebook. We will address the deprecation warnings separately.

vivekmig (approved):
Looks great, thanks so much for fixing this! Just one small nit.

tutorials/House_Prices_Regression_Interpret.ipynb
dbish: makes sense. i'll clean those lines up

PR #1015, opened 2022-08-24
Fix GPU test sane-utils install error

ProGamerGov:
The error seems to still be occurring even with my fix. Hopefully this PR fixes it completely.

The error occurs during the `sudo apt-get update` step of the CircleCI GPU CUDA install, so we need to apply the fix before that line and not after:

```
sudo apt-key add /var/cuda-repo-ubuntu2004-11-4-local/7fa2af80.pub

sudo apt-get update

sudo apt-get --yes --force-yes install cuda
```

The error message given is:

```
Errors were encountered while processing:
 sane-utils
W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

Example of a recent error: https://app.circleci.com/pipelines/github/pytorch/captum/4300/workflows/5f58ee73-92de-41b4-9282-d996822ce567/jobs/24331


vivekmig (approved):
Looks good, thanks for fixing this!

PR #1016, opened 2022-08-27
add docstring style in developer guide

aobo-y:
Write down the agreed docstring style in the developer guide

### DIFF
--- a/docs/contribution_guide.md
+++ b/docs/contribution_guide.md
@@ -86,7 +86,7 @@ Docstring is required for all public APIs to provide users the details of the ar
 - common collection types: `list`, `tuple`, `dict`
     - [PEP 585](https://peps.python.org/pep-0585/#implementation) has deprecated the duplicate types: `List`, `Tuple`, `Dict`
     - element types: `list[int]`, `dict[int, str]`
-- other foundamental types: `Any`, `Callable`, `Interable`
+- other foundamental types: `Any`, `Callable`, `Iterable`
 - class types: `MyClass`, `external_lib.SomeClass`
 - omit `torch` for common Pytorch types: `Tensor`, `nn.Module`
 - use `or` and `,` for union types: `type1 or type2`, `type1, tyep2, or type3`
--- a/docs/contribution_guide.md
+++ b/docs/contribution_guide.md
@@ -77,3 +77,20 @@ https://captum.ai/tutorials/House_Prices_Regression_Interpret
 **Multimodal**
 - You can use VQA model and dataset described here:
 https://captum.ai/tutorials/Multimodal_VQA_Captum_Insights
+
+
+## Docstring style
+
+Docstring is required for all public APIs to provide users the details of the arguments and returns. [Our API documentation](https://captum.ai/api/) is generated from the docstring. Captum adopts a customized docstring format modified on top of [Google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html). Specifically, each argument should be listed as `arg_name (type): description` in the `Args:` section. The argument typing convention:
+- primitive types: `int`, `str`, `float`, `bool`
+- common collection types: `list`, `tuple`, `dict`
+    - [PEP 585](https://peps.python.org/pep-0585/#implementation) has deprecated the duplicate types: `List`, `Tuple`, `Dict`
+    - element types: `list[int]`, `dict[int, str]`
+- other foundamental types: `Any`, `Callable`, `Interable`
+- class types: `MyClass`, `external_lib.SomeClass`
+- omit `torch` for common Pytorch types: `Tensor`, `nn.Module`
+- use `or` and `,` for union types: `type1 or type2`, `type1, tyep2, or type3`
+    - [PEP 604](https://peps.python.org/pep-0604/) proposes to use `|` to connect types: `type1 | type2`. We may consider migration later.
+- append `optional` for argument with default value: `int, optional`
+    - append default value to the end of the description: `Default: None`
+    - Notice this is different with python's type hint `Optional[...]`, which indicate if the argument can be `None`


ProGamerGov:
@aobo-y For comparison, this is the docstring style guide I wrote for PyTorch: https://github.com/pytorch/pytorch/blob/master/CONTRIBUTING.md#docstring-type-formatting

docs/contribution_guide.md
```diff
@@ -77,3 +77,20 @@ https://captum.ai/tutorials/House_Prices_Regression_Interpret
...
+Docstring is required for all public APIs to provide users the details of the arguments and returns. [Our API documentation](https://captum.ai/api/) is generated from the docstring. Captum adopts a customized docstring format modified on top of [Google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html). Specifically, each argument should be listed as `arg_name (type): description` in the `Args:` section. The argument typing convention:
+- primitive types: `int`, `str`, `float`, `bool`
+- common collection types: `list`, `tuple`, `dict`
+    - [PEP 585](https://peps.python.org/pep-0585/#implementation) has deprecated the duplicate types: `List`, `Tuple`, `Dict`
+    - element types: `list[int]`, `dict[int, str]`
+- other foundamental types: `Any`, `Callable`, `Interable`
+- class types: `MyClass`, `external_lib.SomeClass`
+- omit `torch` for common Pytorch types: `Tensor`, `nn.Module`
+- use `or` and `,` for union types: `type1 or type2`, `type1, tyep2, or type3`
+    - [PEP 604](https://peps.python.org/pep-0604/) proposes to use `|` to connect types: `type1 | type2`. We may consider migration later.
```
ProGamerGov, line 93: I don't think that Sphinx supports using "`|`" to separate types, so using it will break the hyperlinking. Up until recently it only supported the using "`or`", but it now also supports "`of`" as well.

docs/contribution_guide.md
```diff
@@ -77,3 +77,20 @@ https://captum.ai/tutorials/House_Prices_Regression_Interpret
...
+
+
+## Docstring style
+
+Docstring is required for all public APIs to provide users the details of the arguments and returns. [Our API documentation](https://captum.ai/api/) is generated from the docstring. Captum adopts a customized docstring format modified on top of [Google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html). Specifically, each argument should be listed as `arg_name (type): description` in the `Args:` section. The argument typing convention:
+- primitive types: `int`, `str`, `float`, `bool`
+- common collection types: `list`, `tuple`, `dict`
+    - [PEP 585](https://peps.python.org/pep-0585/#implementation) has deprecated the duplicate types: `List`, `Tuple`, `Dict`
+    - element types: `list[int]`, `dict[int, str]`
+- other foundamental types: `Any`, `Callable`, `Interable`
```
ProGamerGov, line 89: The full set of capitalized types should be: "`Callable`", "`Any`", "`Iterable`", "`Iterator`", "`Generator`".

docs/contribution_guide.md
```diff
@@ -77,3 +77,20 @@ https://captum.ai/tutorials/House_Prices_Regression_Interpret
...
+- class types: `MyClass`, `external_lib.SomeClass`
+- omit `torch` for common Pytorch types: `Tensor`, `nn.Module`
```
ProGamerGov, line 91: Unless Tensor and nn are imported in the Python file, this is going to break the Sphinx hyperlinking.

If we want to display `Tensor`, then we should be using `~torch.Tensor`. The `~torch.` could be appended while generating the docs, or be present by default in docstrings themselves.

ProGamerGov: For `nn.Module`, I think the Sphinx compatible options are `torch.nn.Module`, or  `~torch.nn.Module` (displayed as simply `Module`).

docs/contribution_guide.md
aobo-y: Sure, these are just examples. Python surely has more built-in types. No need to list everything.

docs/contribution_guide.md
aobo-y: The pr is just to document the agreed styles that the team prefers to write docstring within the codebase. 

For how sphinx compiles it or how it will be displayed in the webpage, we can leave them to the future. I don't think sphinx has any issues to handle even free text typing. So for "compatible", I guess you only mean hyperlinks, which are not that important anyway given our webpage is not hyperlinked much (as well as other torch libs). We can figure out it later.

docs/contribution_guide.md
ProGamerGov: I would recommend just using "`or`" to connect types for now.

docs/contribution_guide.md
ProGamerGov: @aobo-y Yeah, for this issue I already have the code for turning `Tensor` into `~torch.Tensor` during HTML creation, so it that's basically a solved problem.

The usage of `nn.Module` however doesn't work as there is no way to ensure the intersphinx extension always recognizes it correctly, and manual replacement strategies cannot reproduce the right styling. It may show up as hyperlinked in files where `import torch.nn as nn` is used, which will then be inconsistent across Captum as not every file does that.

I would say that it's easiest to just include `torch.` with `nn.Module` for now, and then it can be changed in the future if a better solution is found. Future formatting changes would then decrease line lengths, meaning it'll be easier to do without fixing docstring line length linting errors.

docs/contribution_guide.md
ProGamerGov: @aobo-y Yeah, there are a more types and would be rather tedious to list them. Those were the types that I found when I inspected the docstrings in Captum's code, and I thought it would be easier for users as they were seemingly common. I do like your idea though of combining a bunch of the types into a single "`Interable`"!

docs/contribution_guide.md
aobo-y: again, this pr is to standardize docstring. For whatever is broken, we can fix it next, especially hyperlinks which never work as expected anyway.

Moreover, for this case, "`|`" is the recommended way in python typehints. I doubt sphinx does not support this delimiter. It should support all official typehints. You may try it with Captum if curious. But anyway, we are not migrating to it now.

docs/contribution_guide.md
aobo-y: great then! At least, we are sure about `Tensor`.

Inconsistent is ok for now. Just like no file is currently following the rules here. We can progressively fix them next. Rules here are for how developer write docstring within the code. The compiled page may look different. For example, "nn.Module" can be rendered as "torch.nn.Module". We will figure out when we move to improve the building process.

docs/contribution_guide.md
ProGamerGov: Sphinx's Python domain supports it according to the deliminator regex and supporting tests: https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/domains/python.py#L383. However, I was concerned about the napoleon extension's lack of explicit regex for detecting "`|`": https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/ext/napoleon/docstring.py

However, I just tested with Sphinx 4.3.2 and it appears to work:

formatting_test1

Sphinx's napoleon extension is really sparse on testing, so I suppose I should write some more tests in order to protect the usage of "`|`" from being broken.

PR #1017, opened 2022-08-29
doc: rectify layerlrp example

YBooks:
This PR aims to rectify the LayerLRP example since it's talking about LayerLRP and not LRP. Moreover LRP doesn't require 2 arguments.


aobo-y:
@YBooks Thank you for fixing this!

PR #1033, opened 2022-09-22
check sphinx build in ci & remove pytext rst

aobo-y:
- Require successful sphinx build in CI. Previously sphinx build will always pass even with errors (example below)
image

- Fixed the remaining error sphinx build: remove `pytext.rst`. We never really opensourced pytext model yet and its doc can never be built due to missing pytext (which is deprecated now); we can later migrate to torchtext and re-opensource it if needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ commands:
     steps:
       - run:
           name: "Run sphinx"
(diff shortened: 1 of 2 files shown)


vivekmig (approved):
Looks great, thanks for fixing this :) !

PR #1034, opened 2022-09-23
fix docstring - multiple or

aobo-y:
use only one `or` for multiple types


captum/_utils/gradient.py
```diff
@@ -724,7 +724,7 @@ def _compute_jacobian_wrt_params(
                 It is unpacked before passing to `model`, so it must be a tuple.  The
                 individual elements of `inputs` can be anything.
         labels (Tensor or None): Labels for input if computing a loss function.
-        loss_fn (torch.nn.Module or Callable or None): The loss function. If a library
+        loss_fn (torch.nn.Module or Callable, optional): The loss function. If a library
```
NarineK, line 727: For the case `labels (Tensor or None)` it is also option like loss_fn ? Do we want to fix that too. Line 795 is similar.

PR #1041, opened 2022-10-06
Changing Housing Regression tutorial to use California housing dataset

diegoolano:
Summary: Use of [the Boston dataset ](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html?highlight=load_boston)has been deprecated by sklearn due to ethical issues and so we changed the tutorial to instead use the [California Housing dataset](https://scikit-learn.org/stable/datasets/real_world.html#california-housing-dataset)


vivekmig (approved):
Looks great overall, thanks for the great work on this @diegoolano :) !

A few small comments:

In the last text cell "and spike in Layer 4." -> should probably be "Neuron 0" (this seems like a mistake in the original notebook, should've referred to Neuron rather than Layer)

In the attribution comparison, it looks like gradient SHAP now is less aligned with other methods compared to before, which is likely due to using the training data distribution as the baseline distribution (as opposed to 0). It might be good to highlight that the baseline is different for GradientShap here as a potential cause for the discrepancy ("In this tutorial we use zero-valued baselines for all features, ...") or alternatively we could consider removing GradientShap from the comparison if it's more confusing. cc @NarineK - what do you think?

vivekmig:
Also, looks like the model file isn't included in this PR, we need to update that as well. The current model is [boston_model.pt](https://github.com/pytorch/captum/blob/master/tutorials/models/boston_model.pt)

PR #1042, opened 2022-10-11
support nested progress bars for `SimpleProgress`

cyrjano:
This change support nested progress bar for `SimpleProgress` in a reasonable way (not entirely equal to tqdm).
The approach is the following: 
- Support context `__enter__` , `__exit__` method so that we can identify the parent progress bar.
- For the parent progress bar add a new line for each call (instead of going back to the original line as tqdm does) to keep the code simple.
- update the code in `Tracincp` and  `TracInCPFast` to use `width` statement.

Screen Shot 2022-10-11 at 12 43 23 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ commands:
     steps:
       - run:
           name: "Run sphinx"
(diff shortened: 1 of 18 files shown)


cyrjano:
Did some cleanup on it.

PR #1043, opened 2022-10-11
Mask for Adversarial Attacks

### LINKED ISSUE #941: Mask for Adversarial Attacks
Hi, thanks for the great work!

Do you plan to offer the possibility to define a mask for adversarial attacks (PGD, FGSM) in the future? 

I'm thinking about a mask that defines to which part of the input the adversarial perturbations should be applied.

This would make it possible to perturb only certain parts of the input, while the rest of the input would remain unchanged. 

Thus, adversarial examples could be generated much more flexibly.

Thanks a lot!

jgonik:
Adds an optional mask argument to the FGSM and PGD adversarial attacks. This mask determines which pixels are affected by the adversarial perturbations. If no mask is specified, then all pixels are affected. This PR resolves #941.

### DIFF
--- a/captum/robust/_core/fgsm.py
+++ b/captum/robust/_core/fgsm.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from typing import Any, Callable, Optional, Tuple
+from typing import Any, Callable, Optional, Tuple, Union
 
 import torch
 from captum._utils.common import (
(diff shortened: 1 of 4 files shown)


vivekmig (approved):
Looks great, thanks for the great work on this @jgonik as well as the thorough test cases! 

There do seem to be some minor mypy failures in the CircleCI test results, can you look into resolving those? You can also run these locally using scripts/run_mypy.sh

vivekmig:
Looks great, thanks! You can import this diff now, looks ready to merge.

PR #1044, opened 2022-10-11
🐛 Update features.py

Infinity-ops:
ISSUE:: When the CIFAR10/FashionMNIST greyscale images are used as the input to ImageFeature, using "from captum.insights import AttributionVisualizer" produces ==> Runtime Error: Number of Dims in Permute Does Not Match

FIX:: Existing feature.py accepts only RGB images as input, when greyscale images is passed as input to the captum insights visualizer, it raises the Runtime Error. It fixed by removing the permute operation based on the input image format.

### DIFF
--- a/captum/insights/attr_vis/features.py
+++ b/captum/insights/attr_vis/features.py
@@ -118,11 +118,20 @@ class ImageFeature(BaseFeature):
         if self.visualization_transform:
             data = self.visualization_transform(data)
 
-        data_t, attribution_t = [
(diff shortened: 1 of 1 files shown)


craymichael:
Hi @Infinity-ops, thank you for your contribution! Unfortunately, we have deprecated Captum Insights and will be removing it from Captum in the upcoming release. If you're interested in maintaining Insights, we recommend forking from v0.8.0 and landing your changes there. Thanks again!

PR #1047, opened 2022-10-14
refactor feature ablation

aobo-y:


captum/attr/_core/feature_ablation.py
```diff
@@ -316,8 +314,9 @@ def attribute(
             )
 
             total_attrib = [
+                # attribute w.r.t each output element
                 torch.zeros(
-                    (num_outputs,) + input.shape[1:],
+                    (n_outputs,) + input.shape[1:],
```
aobo-y, line 319: should we respect the output shape, instead of flatten it, like `(*initial_eval.shape, *input.shape[1:])`?

For example, if the forward return in shape(5, 2), the attribution should be in shape(5,2, ...), not shape(10, ...)

cc @vivekmig @NarineK

captum/attr/_core/feature_ablation.py
```diff
@@ -286,22 +286,20 @@ def attribute(
             if show_progress:
                 attr_progress.update()
 
+            # number of elements in the output of forward_func
+            n_outputs = initial_eval.numel() if isinstance(initial_eval, Tensor) else 1
+
+            # flattent eval outputs into 1D (n_outputs)
+            # add the leading dim for n_feature_perturbed
+            if isinstance(initial_eval, Tensor):
```
aobo-y, line 294: is it common for the `forward` not return a Tensor? What else do we expect? I would guess `float` or `double`, but they are all very easy to be converted into tensor and users can just do it by wrapping the forward. So we can explicitly define the return type of `forward`. 

Even if we really want to support non-Tensor returns, can we convert them in `_run_forward` to standardize the logic?

vivekmig (approved):
Looks great, this looks cleaner, thanks for updating this!

captum/attr/_core/feature_ablation.py
aobo-y: > the output attributions shape may no longer always be aligned with the input shape
do you mean the number of dimension would no longer be the same between input and returned attribution? E.g., if the input is `(batch_size, A, B)`  and the forward outputs `(C, D)`, the attr will be`(C*D, A, B)`. The attr does also have 3 dimensions like the input, but the shape may not be the same (where `C*D != batch_size`).

My suggestion tries to respect the output of the forward more. The dimension (or the entire shape) of the returned attr depend on the input and output shapes of the forward. E.g., if the input is in `(batch_size, A, B)` and the forward outputs `(C, D)`, the attribution will be in `(C, D, A, B)`. I feel if the user make the forward returns `(C, D)`, most likely the user also need to reshape our returned `C*D` to `(C, D)` to further aggregate or select the attribution, e.g., C is batch_size, D is n_tasks. 

I can see the reasons behind both designs. But I would suggest to not squeeze if we use the 2nd way. Coz we do not understand the exact meaning of the dims of 1. I am afraid some weird edge cases that keeping the 1 can be very convenient (users want to reuse the selection indices created for the forward output). I would trust the users to know their own model output 😆 . If they feel the dims of 1 are really annoying, they can always control it by wrapping their own model `forward = lambda *args, **kargs: forward(*args, **kargs).squeeze()`

PR #1049, opened 2022-10-21
convert forward return to tensor in FeatureAblation

aobo-y:
FeatureAblation(Permutation) never clearly define the type of `forward`'s return.  According to the [documentation](https://captum.ai/api/feature_ablation.html), `tensor` is the only acceptable type

> The forward function can either return **a scalar per example** or **a tensor of a fixed sized tensor (or scalar value)** for the full batch

However, returning a single `int` or `float` is a common use case we have already supported (ref https://github.com/pytorch/captum/pull/1047#discussion_r997314566). But our code did not explicitly raise error for unexpected types. Other types like `list`, `tuple`, or `numpy.array` may pass unexpectedly or fail in unexpected places with confusing error messages, such as we may use `list` as `torch.dtype` https://github.com/pytorch/captum/blob/5f878af6a74a5d041dc478addabfaf365d7810d5/captum/attr/_core/feature_ablation.py#L320

The PR explicitly assert the return type and convert everything into `tensor`. The assertion & conversion is done in a new private `_run_forward` wrapper instead of `_run_forward` from global utils, which is shared by many other classes. I will update others progressively and eventually push the logic to the shared `_run_forward`.

### DIFF
--- a/captum/attr/_core/feature_ablation.py
+++ b/captum/attr/_core/feature_ablation.py
@@ -279,7 +279,7 @@ class FeatureAblation(PerturbationAttribution):
 
             # Computes initial evaluation with all features, which is compared
             # to each ablated result.
-            initial_eval = self._run_forward(
+            initial_eval = self._strict_run_forward(
                 self.forward_func, inputs, target, additional_forward_args
             )
 
@@ -352,7 +352,7 @@ class FeatureAblation(PerturbationAttribution):
                     #   agg mode: (*initial_eval.shape)
                     #   non-agg mode:
                     #     (feature_perturbed * batch_size, *initial_eval.shape[1:])
-                    modified_eval = self._run_forward(
+                    modified_eval = self._strict_run_forward(
                         self.forward_func,
                         current_inputs,
                         current_target,
@@ -594,7 +594,12 @@ class FeatureAblation(PerturbationAttribution):
             or all(len(sm.shape) == 0 or sm.shape[0] == 1 for sm in feature_mask)
         )
 
-    def _run_forward(self, *args, **kwargs) -> Tensor:
+    def _strict_run_forward(self, *args, **kwargs) -> Tensor:
+        """
+        A temp wrapper for global _run_forward util to force forward output
+        type assertion & conversion.
+        Remove after the strict logic is supported by all attr classes
+        """
         forward_output = _run_forward(*args, **kwargs)
         if isinstance(forward_output, Tensor):
             return forward_output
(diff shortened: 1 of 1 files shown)


captum/attr/_core/feature_ablation.py
```diff
@@ -601,3 +593,20 @@ def _find_output_mode(
...
+        assert output_type is int or output_type is float, (
+            "the return of forward_func must be a tensor, int, or float,"
+            f" received: {forward_output}"
+        )
+
+        # using python type as torch dtype is not a documented behavior
+        # our tests expect int -> torch.int64, float -> torch.float64
+        # but this may actually depend on the machine
+        # ref: https://docs.python.org/3.10/library/stdtypes.html#typesnumeric
+        return torch.tensor(forward_output, dtype=output_type)
```
aobo-y, line 612: I inherit our original logic that passing python types to torch dtype, like `dtype=float`.  But this is not an officially documented operation. Existing tests assume it must equal to `dtype=torch.float64` https://github.com/pytorch/captum/blob/5f878af6a7/tests/attr/test_feature_ablation.py#L429

But this may be machine dependent https://docs.python.org/3.10/library/stdtypes.html#typesnumeric . 

> Floating point numbers are usually implemented using double in C; information about the precision and internal representation of floating point numbers for the machine on which your program is running is available in [sys.float_info](https://docs.python.org/3.10/library/sys.html#sys.float_info) 

Two other alternatives are:
- explicitly map python type to torch dtype: float -> torch.float64
- do not set dtype, rely on torch's default dtype (float32)

captum/attr/_core/feature_ablation.py
```diff
@@ -601,3 +593,20 @@ def _find_output_mode(
             feature_mask is None
             or all(len(sm.shape) == 0 or sm.shape[0] == 1 for sm in feature_mask)
         )
+
+    def _run_forward(self, *args, **kwargs) -> Tensor:
+        forward_output = _run_forward(*args, **kwargs)
```
vivekmig, line 598: nit: It seems a bit confusing when seeing both the instance method and original method named as _run_forward, could consider renaming this one slightly, but either way is fine.

vivekmig (approved):
Looks great, thanks! Just one note on the dtype comments.

captum/attr/_core/feature_ablation.py
aobo-y: Thx for the deep dive @vivekmig ! 
Then I will just add the comment to refer the mapping, also as a caveat.
After all, it is not a documented torch usage. May have breaking changes someday.

vivekmig: Makes sense, sounds good!

PR #1052, opened 2022-10-25
Return more helpful type error messages in _select_targets().

dkrako:
This is a very small change.

The error message we get from `_select_targets()` can be confusing when passing numpy arrays instead of torch.tensors or lists of integers.

I therefore added the actual type to the error message.


NarineK (approved):
LGTM! Thank you for the fix, @dkrako!

PR #1061, opened 2022-11-08
Fix failing kernel shap tests

NarineK:
Kernel SHAP's tests have been failing with the new PyTorch's version 1.14.0 on master. This fix resets random seeds to bring determinism to the tests. 
Here is the link to the failing tests: 
https://app.circleci.com/pipelines/github/pytorch/captum/4445/workflows/0db80fcb-08c7-4111-81c9-8e68588c5750/jobs/25508

### DIFF
--- a/tests/attr/test_kernel_shap.py
+++ b/tests/attr/test_kernel_shap.py
@@ -371,7 +371,6 @@ class Test(BaseTest):
     ) -> None:
         for batch_size in perturbations_per_eval:
             kernel_shap = KernelShap(model)
-            set_all_random_seeds(1234)
             attributions = kernel_shap.attribute(
                 test_input,
                 target=target,
--- a/tests/attr/test_kernel_shap.py
+++ b/tests/attr/test_kernel_shap.py
@@ -12,6 +12,7 @@ from tests.helpers.basic import (
     assertTensorAlmostEqual,
     assertTensorTuplesAlmostEqual,
     BaseTest,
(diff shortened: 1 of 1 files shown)


tests/attr/test_kernel_shap.py
```diff
@@ -386,6 +387,7 @@ def _kernel_shap_test_assert(
             )
 
             if expected_coefs is not None:
+                set_all_random_seeds(1234)
```
vivekmig, line 390: nit: Could also add this before line 374, so for each iteration of the loop, both attribution calls are consistent.

vivekmig (approved):

tests/attr/test_kernel_shap.py
NarineK: I remember that we discussed it and I tried but it didn't pass the tests. I think that is probably because since the seeds haven't been initialized, for some tests the expected values are computed based on the next seed value and doesn't work when I reset it. In that case we need to change the expected values for some test cases.

vivekmig: Interesting, having set_all_random_seeds in both places should make sure they are always consistent (both at 390 and 373). Was the issue when adding the line in both places or putting it only at 374? 

Putting it in both places should ensure that the result values are identical with either return_input_shape being True or False, so might help to reduce issues in the future if there are further changes. But otherwise, this change looks good!

PR #1081, opened 2022-12-05
Fix failing GPU errors for influential examples

NarineK:
Due to floating point arithmetic inaccuracies and limitations switching to double for test cases and passing dtype to projection matrix so that the the arithmetic error is within accepted range.


vivekmig (approved):

PR #1086, opened 2022-12-07
Add split_channels parameter to LayerGradCam.attribute

dzenanz:
This allows examination of each channel's contribution. That is useful if channels are something other than standard RGB, for example multi-spectral input, potentially with many spectral channels.

### DIFF
--- a/captum/attr/_core/layer/grad_cam.py
+++ b/captum/attr/_core/layer/grad_cam.py
@@ -82,6 +82,7 @@ class LayerGradCam(LayerAttribution, GradientAttribution):
         additional_forward_args: Any = None,
         attribute_to_layer_input: bool = False,
         relu_attributions: bool = False,
+        split_channels: bool = False,
     ) -> Union[Tensor, Tuple[Tensor, ...]]:
         r"""
         Args:
@@ -149,6 +150,10 @@ class LayerGradCam(LayerAttribution, GradientAttribution):
                         otherwise, by default, both positive and negative
                         attributions are returned.
                         Default: False
+            split_channels (bool, optional): Indicates whether to
+                        keep attributions split per channel.
+                        The default (False) means to sum per channels.
+                        Default: False
 
         Returns:
             *Tensor* or *tuple[Tensor, ...]* of **attributions**:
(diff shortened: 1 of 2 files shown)


dzenanz:
I signed CLA. Is some manual recheck invocation required?

dzenanz:
This PR is rather small and simple. Any chance someone will review, and maybe merge it?

NarineK:
@dzenanz, @vivekmig will look into it. Sorry for the delay.

dzenanz:
@vivekmig ping!

dzenanz:
I would still like to have this merged.

NarineK:
@split_channels, do you mind adding a test case for `split_channels` ? We have test cases for LayerGradCam you can add another test case for `split_channels` feature.

captum/attr/_core/layer/grad_cam.py
```diff
@@ -208,10 +213,16 @@ def attribute(
             for layer_grad in layer_gradients
         )
 
-        scaled_acts = tuple(
-            torch.sum(summed_grad * layer_eval, dim=1, keepdim=True)
-            for summed_grad, layer_eval in zip(summed_grads, layer_evals)
-        )
+        if split_channels:
```
NarineK, line 216: It looks like it is useful when `relu_attributions == True` because otherwise we can also sum returned value of the attributions.

dzenanz:
I will add a test case, probably tomorrow.

captum/attr/_core/layer/grad_cam.py
dzenanz: `split_channels` is independent of `relu_attributions`. They can be applied independently.

dzenanz:
The first force-push does a plain rebase.

dzenanz:
I added a test, and it passes on my computer. Please re-review.

dzenanz:
The second force-push fixes linting errors.

dzenanz:
I can't see the details of failure for "Facebook Internal - Builds & Tests".

captum/attr/_core/layer/grad_cam.py
```diff
@@ -82,6 +82,7 @@ def attribute(
         additional_forward_args: Any = None,
         attribute_to_layer_input: bool = False,
         relu_attributions: bool = False,
+        split_channels: bool = False,
```
NarineK, line 85: @dzenanz, `channel` suffix seems to be image specific. Our implementation is generic and we prefer having generic names. Do you mind giving a more generic name ? Something like: `attr_dim_summation=False` ? 
cc: @vivekmig

dzenanz: Done.

dzenanz: Now I realize this rename would need to invert the logic, and the default from False to True. I will force-push.

NarineK: Thank you very much for the quick fix, @dzenanz! I'll let @vivekmig to have a final look.

dzenanz:
The last force push is a plain rebase on master.

PR #1091, opened 2022-12-14
validate forward output in FeatureAblation

aobo-y:
Correctly validate the `forward_fun`'s output shape in `FeatureAblation` (and `FeaturePermutation`)

Abandoned previous flawed assumption of "aggregation mode", which forbid the support for multi-outputs models (ref https://github.com/pytorch/captum/pull/1047)

New logic does not care output shape when `perturbations_per_eval == 1`.  Only when `perturbations_per_eval > 1`, it require "aggregation mode", which is defined as the 1st dim of the model's output should grow with the input's batch size in the same ratio. This is achieved by actually comparing the output shape of 2 different inputs instead of making any assumption based on other user config:
- The baseline output is from the initial eval with the original inputs which we have to run anyway. 
- The expanded output is from the 1st ablated eval whose input batch size has been expanded for more feature perturbation.

This way does not even introduce any extra forward calls.

### DIFF
--- a/captum/attr/_core/feature_ablation.py
+++ b/captum/attr/_core/feature_ablation.py
@@ -53,6 +53,15 @@ class FeatureAblation(PerturbationAttribution):
         PerturbationAttribution.__init__(self, forward_func)
         self.use_weights = False
 
+        # only used when perturbations_per_eval > 1, where the 1st dim of forward's
+        # output grow as the input batch. If forward's output is aggregated,
+        # we cannot expand the batch size to include more perturbations in one call.
+        # If it's False, we will check with an additional run where
+        # perturbations_per_eval = 1 to see if the output shape is expected;
+        # but it turns to True, we will assume the model's hehavior stays
+        # consistant and no longer check again
+        self._is_output_shape_valid = False
+
     @log_usage()
     def attribute(
         self,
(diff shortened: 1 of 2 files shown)


captum/attr/_core/feature_ablation.py
```diff
@@ -362,21 +360,43 @@ def attribute(
...
+                            # check if the output is not a scalar
+                            current_output_shape
+                            and initial_output_shape
+                            # check if the output grow in same ratio, i.e., not agg
+                            and current_output_shape[0]
+                            == n_perturb * initial_output_shape[0]
+                        ), (
+                            "When perturbations_per_eval > 1, forward_func's output "
+                            "should be a tensor whose 1st dim grow with the input "
+                            f"batch size: when input batch size is {num_examples}, "
```
aobo-y, line 387: I think we can be more strict here by requiring the 1st dim of output to be the same as input batch size, instead of just require it "grow with the input batch size".

For example, if the input batch is `2`, the output's 1st dim can be `2`, `4`, or `7`; as long as
when we expand the input batch size to `4`, the output's 1st dim becomes `4`, `8`, or `14` respectively. 

But the last 2 cases are pretty weird and unlikely to happen. It may be easier to always require output's 1st dim to be batch size and it must be the same as input batch size in this "non-aggregation mode"

tests/attr/test_feature_ablation.py
```diff
@@ -345,17 +345,6 @@ def forward_func(inp):
         with self.assertRaises(AssertionError):
             _ = ablation.attribute(inp, perturbations_per_eval=2)
 
-    def test_error_agg_mode_incorrect_fm(self) -> None:
-        def forward_func(inp):
-            return inp[0].unsqueeze(0)
-
-        inp = torch.tensor([[1, 2, 3], [4, 5, 6]])
-        mask = torch.tensor([[0, 1, 2], [0, 0, 1]])
```
aobo-y, line 353: This test checks when rowwise `mask` is give, the model must not be in "aggregated mode". But I don't think `mask` has anything to do with it. Models of both batch-mode and aggregated-mode may need specify `mask`. So I simply deleted this test.

PR #1094, opened 2022-12-16
Updating version to 0.6.0

aobo-y:


dzenanz:
Will #1086 make it into `0.6`?

aobo-y:
@dzenanz Sorry, unfortunately, #1086 is not included in the v0.6 as it has not been merged before last week. I saw @NarineK & @vivekmig have already started the review, so we will include it in the next release. Thank you for your contribution to Captum.

PR #1106, opened 2023-01-28
Replace she/he with they

jessijzhao:
* Make language more inclusive by removing unnecessarily gendered language


aobo-y (approved):
Thank you for fixing it

Issue #1114, opened 2023-02-22
visualization.visualize_image_attr crashes with matplotlib 3.7


daniel-ziegler:
## To Reproduce

1. Install matplotlib 3.7.0
2. Make any call to `visualization.visualize_image_attr`

```
  File "/home/dmz/miniconda3/envs/audit_vis/lib/python3.10/site-packages/captum/attr/_utils/visualization.py", line 250, in visualize_image_attr
    plt_axis.grid(b=False)
  File "/home/dmz/miniconda3/envs/audit_vis/lib/python3.10/site-packages/matplotlib/axes/_base.py", line 3196, in grid
    self.xaxis.grid(visible, which=which, **kwargs)
  File "/home/dmz/miniconda3/envs/audit_vis/lib/python3.10/site-packages/matplotlib/axis.py", line 1655, in grid
    self.set_tick_params(which='major', **gridkw)
  File "/home/dmz/miniconda3/envs/audit_vis/lib/python3.10/site-packages/matplotlib/axis.py", line 927, in set_tick_params
    kwtrans = self._translate_tick_params(kwargs)
  File "/home/dmz/miniconda3/envs/audit_vis/lib/python3.10/site-packages/matplotlib/axis.py", line 1071, in _translate_tick_params
    raise ValueError(
[...]

PR #1134, opened 2023-04-18
TracIn FBLearner Flow bug TracInCPFastRandProj projection_dim mismatch

diegoolano:
Summary:
For F416836189, F417066355 and other runs of FBLearner Flow for computing influential embeddings using TracInCPFastRandProj, there is a possible bug of the final embedding size being of dimension 60 when "projection_dim" passed in is of 100.  Its possible that this isn't a bug per-say, but its something that we should look into to either fix or possibly document better.

Just for reference I haven't seen this issue using TracInCPFastRandProj without passing in "projection_dim" explicitly ( which is effectively just TracInCPFast such as F418560294) or when using Arnoldi Influence ( such as F418927001).

**Update, This is not a bug, but rather the intended output of the method.  I've added a test with different configurations to further explain how the projection method works using TracInCPFastRandProj

Reviewed By: NarineK

Differential Revision: D45050510

### DIFF
--- a/tests/influence/_core/test_tracin_intermediate_quantities.py
+++ b/tests/influence/_core/test_tracin_intermediate_quantities.py
@@ -250,3 +250,104 @@ class TestTracInIntermediateQuantities(BaseTest):
                 delta=0.01,  # due to numerical issues, we can't set this to 0.0
                 mode="max",
             )
+
+    @parameterized.expand(
+        [
+            (reduction, constructor, projection_dim, unpack_inputs)
+            for unpack_inputs in [False]
+            for (reduction, constructor, projection_dim) in [
+                ("sum", DataInfluenceConstructor(TracInCPFastRandProj), None),
+                ("sum", DataInfluenceConstructor(TracInCPFastRandProj), 2),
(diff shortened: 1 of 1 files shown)


NarineK:
Thank you for importing, @diegoolano! Let's look into the failing tests and address them

aobo-y:
@diegoolano @NarineK The failed tests should have nothing to do with this PR. We can ignore them.

`py36` is failing because some previous changes in `test_gradient_shap` broke `mypy`. I can fix it.

```
tests/attr/test_gradient_shap.py:235: error: Argument "delta_thresh" to "_assert_attribution_delta" has incompatible type "float"; expected "Tensor"
tests/attr/test_gradient_shap.py:256: error: Incompatible default for argument "delta_thresh" (default has type "float", argument has type "Tensor")
tests/attr/test_gradient_shap.py:272: error: Incompatible default for argument "delta_thresh" (default has type "float", argument has type "Tensor")
Found 3 errors in 1 file (checked 95 source files)
```

`py37` fails due to known issue about `matplotlib` discussed below. @vivekmig will push a fix soon
https://github.com/pytorch/captum/issues/1114

PR #1152, opened 2023-06-09
Correct default `method` argument for captum.attr.visualization.visualize_t…

teddykoker:
The default argument for `method` in `captum.attr.visualization.visualize_timeseries_attr` is currently `"individual_channels"`, which is not a valid option, resulting in an exception if used. This PR changes the default method to `"overlay_individual"`, which is what the docs indicate the default should be.


vivekmig (approved):
Thanks so much for this fix @teddykoker !

PR #1171, opened 2023-08-25
Add InterpretableInput

aobo-y:
Summary:
Define the interface for the abstract `InterpretableInput` class.
More details in the docstring

Differential Revision: D48486800

### DIFF
--- a/captum/attr/_utils/interpretable_input.py
+++ b/captum/attr/_utils/interpretable_input.py
@@ -1,0 +1,100 @@
+from typing import Any, Optional
+
+from torch import Tensor
+
+
+class InterpretableInput:
+    """
+    InterpretableInput is an adapter for different kinds of model inputs to
+    work in Captum's attribution methods. Generally, attribution methods of Captum
+    assume the inputs are numerical PyTorch tensors whose 1st dimension must be batch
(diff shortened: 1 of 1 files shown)


captum/attr/_utils/interpretable_input.py
```diff
@@ -0,0 +1,100 @@
...
+        input elements, the attribution of the interpretable features can be scattered
+        back to the model input shape.
+
+        Args:
+
+                itp_attr (Tensor): attributions of the interpretable features
+
+        Returns:
+
+                attr (Tensor): formatted attribution
```
vivekmig, line 98: Would it make sense to allow the return type to be Any here to allow a custom dict, etc. to represent the attributions in a custom format similar to input rather than a tensor?

vivekmig (approved):
Looks great, thanks for adding this!

captum/attr/_utils/interpretable_input.py
aobo-y: Maybe! My only concern is some of our downstream tasks assume the attr is tensor, like stats aggregation. We can discuss this when we face an Input type that using tensor as attr will be very confusing or inconvenient

PR #1181, opened 2023-09-01
Migrate to GitHub Actions

vivekmig:
Migrates test jobs from CircleCI to Github Actions


.github/workflows/test-pip-cpu.yml
```diff
@@ -0,0 +1,36 @@
...
+  tests:
+    strategy:
+      matrix:
+        python_version: ["3.6"]
+        pytorch_args: ["", "-n", "-v 1.6", "-v 1.7", " -v 1.8", "-v 1.9"]
+      fail-fast: false
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      runner: linux.12xlarge
+      repository: pytorch/captum
```
osalpekar, line 26: You can specify which docker image to run the job in using the `docker-image:` argument here. Feel free to use the CircleCI image that you were using before - as long as it's public, we should be able to use it here. Hopefully this makes the migration smoother

osalpekar (commented):
Thanks for kicking this off. Left a small suggestion to help fix the pip jobs. Can we also deprecate the conda jobs in this PR if they are now working on GHA?

osalpekar:
@vivekmig Can we merge this change, and fix the remaining broken job as a follow-up? We'd like to deprecate the old CircleCI jobs ASAP

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  tests:
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      runner: linux.12xlarge
+      docker-image: cimg/python:3.6-node
+      repository: pytorch/captum
+      script: |
+        sudo chmod -R 777 .
+        ./scripts/install_via_pip.sh -n -d
```
osalpekar, line 25: For any calls to `pip install` within this script or any other script called from these workflows, we probably should add the following option to make the logs more readable: `--progress-bar off`

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,42 @@
...
+
+        if [[ $CIRCLE_PROJECT_USERNAME == "pytorch" && -z $CI_PULL_REQUEST && -z $CIRCLE_PR_USERNAME ]]; then
```
osalpekar: `CIRCLE_PROJECT_USERNAME` does not have a set value. Any env vars without a set value with fail with `unbound environment variable` since we runs this shell script with `set -u`.

PR #1189, opened 2023-10-05
move pred to cpu, which is the same device as lm.classes()

theAdamColton:
without this `.cpu()` call there is an error when training the classifier using cuda, because the predictions will be on the gpu


craymichael:
Thanks for your contribution @theAdamColton!

PR #1211, opened 2023-11-27
Update Minimum PyTorch Version to 1.9

vivekmig:
Updates minimum version of PyTorch to 1.9 and updates GitHub actions tests accordingly to cover all versions after 1.9. This will also ensure torch.linalg.eig is available for influence implementations in all supported versions.

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -16,7 +16,7 @@ requirements:
     - python>=3.6
   run:
     - numpy
-    - pytorch>=1.6
+    - pytorch>=1.9
     - matplotlib-base
 
 test:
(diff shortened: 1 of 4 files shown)


NarineK:
Thank you for the change, @vivekmig ! Don't we have users using PT 1.8 or 1.7. Is this change necessary only for  `torch.linalg.eig`? Would it be possible to use `torch.eig` for lower versions of PT ? 
https://pytorch.org/docs/1.7.0/generated/torch.eig.html#torch.eig

vivekmig:
Closing since we were able to workaround the influence issues, and will keep the minimum version the same for now. Creating a new PR which adds tests for missing version between 1.9 and 2.1

PR #1227, opened 2023-12-28
mps fix

Jerry-Master:
By adding that single conversion from float64 to float32 the integrated gradients is fully compatible with MPS backend, which it wasn't previously. The change can be accepted as valid for any backend since float64 is a precision almost nobody using nowadays in deep learning. It seems that the default for torch.tensor is not updated with current trends.


vivekmig (approved):
Thanks so much for this fix @Jerry-Master !

PR #1328, opened 2024-08-15
Add support for tuples of layers (instead of strictly list) in LayerLRP

Quoding:
- Add checks for tuples instead of just lists of layers in LayerLRP
- Add new type for hinting / documentation -> ModuleOrModuleListOrModuleTuple


craymichael:
Hi, thanks for the contribution! Can you please make the following changes:
- Replace utilization and import of `ModuleOrModuleList` with `ModuleOrModuleListOrModuleTuple`
- Simplify type checks as `isinstance(self.layer, (list, tuple))`
- Rebase onto captum's master

PR #1541, opened 2025-04-05
tests: better isolate distributed group initialization

booxter:
- **tests: initalize gloo group only when test will run**
- **tests: use a temporary file:// as rendezvous**


booxter:
I don't have time to get it to completion after all. If anyone would like to take it over, feel free.

PR #1566, opened 2025-05-17
Update link and formatting

j-t-1:


j-t-1:
Can `import numpy as np` be removed as is not used in the examples?

jjuncho:
@j-t-1 can you elaborate by what you mean by this? which example are you referring to?

j-t-1:
https://github.com/pytorch/captum/blob/b9133d4bd55fede1959e5fd3473ad85d1ae98bb8/README.md?plain=1#L119-L122

I think it is unused in the example, unless it is used indirectly in some way...

jjuncho:
@j-t-1 it's used in some of the attribution classes ie. GradientShap so no changes would need to be made here

jjuncho (approved):
lgtm

Issue #1619, opened 2025-06-26
Adding a Fast Implementation to Calculate Concept Activation Vectors for High-dimensional Models to Captum


asver12:
## Motivation
The computation of Concept Activation Vectors (CAVs) is a fundamental component of concept-based explanation methods, most notably [Testing with Concept Activation Vectors (TCAV)](https://arxiv.org/abs/1711.11279), which is already implemented in Captum. The current standard for computing CAVs in Captum relies on training a Support Vector Machines (SVMs) for each concept. This can present a significant challenge, especially for modern, high-dimensional models. To address this, we propose adding an efficient drop-in with [FastCAV](https://arxiv.org/abs/2505.17883). We provide benchmarking and theoretical justification to support its effectiveness.

# Captum FastCAV API Design

## Background
FastCAV is a novel approach introduced at ICML 2025 based on insights into superposition, that accelerates the computation of CAVs. FastCAV defines the CAV as the vector from the global mean of all activations to the mean of the positive class activations. As such it acts as a drop-in replacement for the prevalent SVM for balanced classes. Additionally, we provide mathematical assumptions under which FastCAV is equivalent to an SVM.

## Requires:

- **Model**: The trained neural network model $f$ to be interpreted.
- **Layer**: A specific layer $l$ within the model from which to extract activations.
- **Concept Dataset $D_c$**: A set of input examples that visually represent the concept to be analyzed.
- **Control Dataset $D_r$**: A set of random or control examples used as a baseline.
## Pseudocode:
```python
def fit(self, x: Tensor, y: Tensor) -> None:
    self.mean = x.mean(dim=0)
    self.coef_ = (x[y == self.classes_[-1]] -self.mean).mean(dim=0).unsqueeze(0)
    self.intercept_ = (-self.coef_ @ self.mean).unsqueeze(1)
```    

### Performance benchmarking:
Given the example in the tutorial `tutorials/TCAV_Image.ipynb` we achieve an average speed-up to the `DefaultClassifier` of  ~8.26 (SVM: 1.09 $\pm$ 0.39, FastCAV: 0.132 $\pm$  0.025)
For bigger models this improves more drastically as shown in Table 1. of the paper.
### Visual Benchmarking
Given the example in the tutorial `tutorials/TCAV_Image.ipynb` we get the following results for `experimental_set_rand` with FastCAV and SVM:
FastCAV:
![Image](https://github.com/user-attachments/assets/c89606d3-1b42-4dab-9d93-46a033115c17)

[...]

PR #1622, opened 2025-06-27
Adding FastCAV

asver12:
Adding FastCAV to Captum. This includes documentation, type hints, and unittests.


asver12:
Closes #1619.

PR #1639, opened 2025-08-25
Remove file to make Captum more focused

j-t-1:
This list is not being maintained, and there are alternative ways of finding this information.


craymichael:
Thanks for the contribution @j-t-1!

PR #1644, opened 2025-09-09
Update FeatureAblation to handle precision loss when baseline is more granular than input when cross tensor attribution is enabled

sarahtranfb:
Summary:
Noticed when flipping the flag, this test case failed:

https://www.internalfb.com/code/fbsource/[faf71541b1ec0fae639f82d487b81fb18ea3e523]/fbcode/pytorch/captum/tests/attr/test_dataloader_attr.py?lines=138%2C134

The ablated tensor was `tensor([0])` instead of `tensor([[0.1])` since the baseline was a float-type and the input tensors were int tensors.

https://www.internalfb.com/code/fbsource/[f2fcc926a6f3669602bac4d28c2d92e4197c96b9]/fbcode/pytorch/captum/captum/attr/_core/feature_ablation.py?lines=707-709

`ablated_input` is just a copy of the `input_tensor`, so during assignment, the ablated feature tensor incorrectly gets cast to an int tensor for this case.

Differential Revision: D81980219


craymichael:
closed by https://github.com/meta-pytorch/captum/commit/c9fc87d1dddda7cd2f5dcdb62c859ca50c1c35d6

PR #1656, opened 2025-10-24
FIX: Automate download of titanic3.csv in Titanic tutorial

gogoi7:
Hi,

I found that the Titanic_Basic_Interpret tutorial was broken because it required a manual data download and changing the dataset path.

This PR fixes it by adding a few lines to the cell to automatically download the "titanic3.csv" file from the given link. I have also updated the markdown instructions right above the code cell to reflect this automated process.

This makes the tutorial runnable out-of-the-box. Let me know if any changes are needed!

Antareep


craymichael:
Thanks for your contribution @gogoi7!

